PR #18, opened 2016-12-22
docs(readme): typo fix

MarkPieszak:
mango-query sounds delicious though :+1:


pubkey:
It actually is called [mango-query](https://blog.couchdb.org/2016/08/03/feature-mango-query/).

MarkPieszak:
Yeah remembered just after, closed it out! Apologies.

PR #57, opened 2017-02-16
Make project compatible with "noImplicitAny" Typescript projects

RickCarlino:
# Why?

 * Using RxDB in Typescript projects with `noImplicitAny` set to `true` produces compile time warnings.

# What Changed?

 * For functions that do not specify a return type, I have added an `any` type.
 * For lifecycle hooks, I have made the return type `void`, as I could not find any tests or documentation that were using the return value. Please let me know if my assumption was incorrect.


pubkey:
@RickCarlino  Thx for your work.

PR #62, opened 2017-02-22
3.0.0

jayem-io:
I think there is an issue in KeyCompressor.js Function compressQuery

In here I think it should be making a call to schema.doKeyCompression

My schema specifies "disableKeyCompression": true


pubkey:
Hi @jayem-io 
You closed this again.
I couldn't figure out your changes, could you clone branch `3.0.0` and add your changes instead of doing a merge?

It's also ok if you just add a test-case which describes the bug, so I will fix it for you.

jayem-io:
I did the wrong thing here. I'm not a github expert. I'd like to contribute. I'll try it again later on today.

PR #70, opened 2017-03-04
Text updates to docs

CrashyBang:


CrashyBang:
Going through the docs now will fire everything off instead as one PR as opposed to individual small ones.

pubkey:
Ok. The PR looks good. It would be great to see a PR with even more enhancements.

PR #82, opened 2017-03-16
README typo 'you can so use' -> 'you can also use'

rdworth:


pubkey:
The word 'so' is meant as 'because of this'. So isn't `You can so use` correct?

rdworth:
In that case it should be 'You can therefore use' or 'So you can use'.

rdworth:
Closing in favor of #83

PR #122, opened 2017-03-25
Make it possible to findOne by id if you have a default PrimaryPath

cbranch101:
Right now doing `findOne('id')` returns all of the documents in the collection if you're using the default primaryPath

Making sure that the primPath isn't `_id` before deleting it from the selectors inside of `toJSON` in Query seems to fix it

### DIFF
--- a/src/RxQuery.js
+++ b/src/RxQuery.js
@@ -120,7 +120,7 @@ class RxQuery {
 
             // selector
             json.selector._id = json.selector[primPath];
-            delete json.selector[primPath];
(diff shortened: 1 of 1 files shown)


pubkey:
Thank you @cbranch101 for the PR. We also need a test for this.

pubkey:
Closed in favor of [55c0524bfe72aa65e0def9e78ae65837ff084f98](https://github.com/pubkey/rxdb/commit/55c0524bfe72aa65e0def9e78ae65837ff084f98)

PR #125, opened 2017-03-29
improve index.d.ts

Jimmysh:
export PouchDB for check adapter can use.

const db = new PouchDB('mydb', { adapter: adapter });
return db.info();

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -196,7 +196,7 @@ declare class RxDocument {
 }
 
 declare class PouchDB {
-    collection(name: string, options:{ adapter:string }): Promise<RxCollection>;
+    constructor(name: string, options:{ adapter:string });
     info();
 }
 
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -15,7 +15,7 @@ declare class RxSchema {
  * @link https://github.com/types/lib-json-schema/blob/master/v4/index.d.ts
  */
 type JsonSchemaTypes = "array" | "boolean" | "integer" | "number" | "null" | "object" | "string";
-export interface JsonSchema {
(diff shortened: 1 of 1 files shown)


pubkey:
Hi @Jimmysh 
I added the param-name to the create-function.
I don't get what the pouchdb-export should do?
There is no collection-method in pouchdb

Jimmysh:
`
function checkAdapterWorks(adapterName: string) {
  try {
    const db = new PouchDB('testDB', { adapter: adapterName });
    return !!db.info();
  } catch (e) {
  }
  return false;
}
const adapters: {
  name: string,
  entity: any
}[] = [
    { name: 'idb', entity: idb },
    { name: 'websql', entity: websql },
    { name: 'localstorage', entity: localstorage },
    { name: 'memory', entity: memory },
  ];

 const find = _.find(adapters, (adapter: any) => checkAdapterWorks(adapter.name));
    if (find) {
      this.useAdapter = find;
      console.log('current adapter', this.useAdapter);
    } else {
      console.log('don't support');
    }

`
sample.

pubkey:
You defined the function `PouchDB.collection` [here](https://github.com/pubkey/rxdb/pull/125/files#diff-b739f95ace6bbf2f0fd5e8cf1dab97d8R199).
But as far as I know, this function does not exists?

pubkey:
Merged, thank you @Jimmysh 
Maybe we can find a way to directly import the typings of pouchdb, like [this ones](http://definitelytyped.org/docs/pouchDB--pouch/globals.html#pouchdb)

pubkey:
The last commit was added here by mistake. Just ignore it.

PR #130, opened 2017-04-09
Refactor mquery to use ES6 module syntax and enable usage of the untranspiled source as a module

jacek-dargiel:
The included mquery library uses a non-standard export syntax. This makes webpack unable to transpile it in the angular2 build process. This PR updates the mquery exports.

With this fixed, angular2 will use the RxDB source files, transpile them in it's own build process, and will not require the babel-polyfill.

Additionally, I've added a `module` field to package.json, as [webpack 2.x doesn't use the provided `jsnext:main` field by default](https://webpack.js.org/configuration/resolve/#resolve-mainfields).

If this is merged, I'll update the included angular2 example.

BTW, I wasn't able to reliably run the tests - they seem to sometimes fail randomly for no apparent reason (I did manage to get 100% passes at least once ;) ).

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
   "license": "Apache-2.0",
   "main": "./dist/lib/index.js",
   "jsnext:main": "./src/index.js",
+  "module": "./src/index.js",
   "types": "./src/index.d.ts",
   "scripts": {
     "pretest": "npm run transpile",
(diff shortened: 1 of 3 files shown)


pubkey:
thank you for the great work :+1: 
I push it to npm i a few minutes.

The tests sometimes fail on slow computers. This is not optimal and I fixed the critical ones on branch [4.0.0](https://github.com/pubkey/rxdb/tree/4.0.0)

jacek-dargiel:
Thanks. I'll try to update the angular2 example today. `babel-polyfill` won't be required anymore. I'm also wondering if RxDB should list it as a peer or optional dependency. What do you think @pubkey?

PR #137, opened 2017-04-18
Fix findOne() with id

natew:


natew:
Attempted to add a test, but the test setup is a little confusing and tedious for me. Questions are:

1. How do I easily make a new test without having to wait for the entire test suite to re-compile each time?
2. How do you log out stuff in tests (console.log seems broken)?

Really hard to contribute without knowing these two things.

Any chance you'd consider moving to Jest?

pubkey:
The changed files do not work
```js
RxQuery.test.js .findOne() should only return one for primary key search:
     TypeError: Cannot read property '_id' of undefined
```

1. Currently everything gets recompiled for each test-run. I know this is annoying and I added a script so it only recompiles the changed files at the current dev-branch [here](https://github.com/pubkey/rxdb/blob/4.0.0/scripts/transpile.js). I'm sorry but it while take like 2 weeks until this gets merged into master.

2. Console.log should work on nodejs-env. For browser-tests it depends on which browser you run them. For chrome, console.log should work. Firefox and IE seem to not support it. Maybe we can change something at the karma-config to make it work.

Btw you can run node-tests only with `npm run test:node`.

There is currently no plan to move to jest. I never thought about it but I will keep it in mind for the next configuration-itteration.

pubkey:
Closed in favor of [55c0524bfe72aa65e0def9e78ae65837ff084f98](https://github.com/pubkey/rxdb/commit/55c0524bfe72aa65e0def9e78ae65837ff084f98)

Issue #144, opened 2017-04-26
definition of CollectionCreator#methods


rweng:
In the current version, the `CollectionCreator` interface requires `statics` and `methods` to be an object with numbers as keys: https://github.com/pubkey/rxdb/blob/83bee8b693d385a80e56b7942f1d275be3f4466d/src/index.d.ts#L78

I guess that is a bug, because in the docs (and mongoose as far as I remember), this is the method name.

pubkey:
Yes this is a bug. `key: string` is what we need.
Also the `function[]` at the migration-strategies is wrong, there we need the number-keys.

pubkey:
Fixed in [#147](https://github.com/pubkey/rxdb/pull/147)

PR #147, opened 2017-04-26
fix typings of CollectionCreator

rweng:
regarding #144


pubkey:
Thank you for the great work.

PR #153, opened 2017-05-04
added update functionality

lgandecki:
Very few changes to the actual code, since all the hard work is getting done in the modifyjs package.
I have a simple test case here just to check the integration, all the heavy lifting tests are done in modifyjs package [src/modify.test.js](https://github.com/lgandecki/modifyjs/blob/master/src/modify.test.js)

At this point all of the mongodb examples are working, with an exception of:

    $currentDate
    $pull (but $pullAll works)
    $push with $sort modifier

There might be more edgecases that don't work and didn't surface while going through the mongodb examples, but we/I can try to tackle them one by one. :) 

Please let me know what you think.


pubkey:
Hi @lgandecki thank you for the work. I will review it at the weekend.
Please do not add build-files (dist-folder) to the PR.

src/RxQuery.js
```diff
@@ -296,6 +296,23 @@ class RxQuery {
...
+        if (Array.isArray(docs)) {
+            await Promise.all(
+                docs.map(doc => doc.update(updateObj))
+            );
+        } else {
+            // via findOne()
+            await docs.update(updateObj);
+        }
+        return docs;
+    }
```
lgandecki: we could possibly extract this and remove to a common function, it's very similar (the only two differences are the "action" (.update vs .remove) and the fact that update takes an argument)

Up to you - some people like to apply a rule of three :)

drop-c:
Any news on this? Looks promising

pubkey:
TODO:

- Add to index.d.ts
- Check why `deepEqual` and `clone` are not used from rxdb

PR #169, opened 2017-05-14
ionic2 example

ntrp:


pubkey:
HI @ntrp 
Sorry for the late reply. Thank you for the great work.

I was a bit confused by the content of the README.md
But everything works as expected.

ntrp:
Oh right, I forgot to modify the default one.. I tried creating an Nativescript example but there are a lot of issues.. Are you sure it is compatible? The biggest problem is that puchdb depends on some packages that either need dom or nodejs packages. I don't think it's possible without modification.

pubkey:
@ntrp You could be right and maybe nativescript is not compatible.
I will also investigate a bit. We should remove nativescript from the docs if it RxDB does not work with it.

PR #173, opened 2017-05-18
Update README.md content: current version -> 4.0.2

lyquocnam:


pubkey:
Hi @lyquocnam 
The announcements are only shown for major-versions with breaking changes. Thats why the link is to the 4.0.0-changelog. I think the link-url should always describe the link-goal which is why it should stay 4.0.0 here.

PR #177, opened 2017-05-18
Update RxCollection.md

natew:


pubkey:
Thank you very much. I totally forgot the pouchSettings in the docs.
Merged.

natew:
By the way I did try changing a few settings and didn't see any change in
sync behavior. I passed them in for both database and collection.

Perhaps something on my end but still exploring how to get a fast initial
sync and in general how to control how much history people download. For
common use cases you just want the last revision. Any pointers appreciated!

PR #178, opened 2017-05-19
Added explicit any return for PouchDB info

JBlaak:
Allows usage of TypeScript compiler flag "noImplicitAny", this was the only place no explicit return type was provided for a function.


pubkey:
Thank you very much. Merged.

PR #183, opened 2017-05-21
two hotfixes

joehannes:
there are what seem to me 2 bugs in the v4.0.2 which render RxDB unusable for me - this fixes this (so far):

* first bug ... every top level property is 2 chars long after splitting it to "." ... so >=2 must be >2
* 2nd bug ... on keycompression you try to only recurse on nested objs, but typeof null === "object"


joehannes:
it occurs to me bug #2 is duplicated in the _decompression function, which I recognized now only.
can't seem to make a second pull request, don't know why

pubkey:
Hi @joehannes thank you very much for the PR.
Sorry for not writing back at gitter, I wasn't online the last 2 days.

Please do not commit auto-generated files (everything in the `dist`-folder).

It would also be cool to have 2 tests which ensure that the bugs will not occur anytime in the future. But it's also ok to commit without tests, so I will add them.

pubkey:
TODO
- [ ] remove dist-files from the PR
- [ ] add tests to describe the bug and ensure it will always work in the future

joehannes:
can you do the tests, currently no time here

pubkey:
Can you paste some example-code on how it didn't work for you without the hotfix?
That would save much time.

joehannes:
First bug is one that renders everything unusable to me instantly ... I can't have no primary key defined because of it, so I can't query later as no primary key is defined ...
Maybe I got a mistake in my schema due to some misunderstanding, here's an example which didn't work for me without the fix:

```json
{
  "title": "artist",
  "version": 0,
  "description": "describes a spotify api search artist result",
  "type": "object",
  "properties": {
    "id": {
      "type": "string",
      "primary": true
    },
    "external_urls": {
      "type": ["object", "null"],
      "disableKeyCompression": true,
      "properties": {
        "spotify": "string"
      }
    },
    "followers": {
      "type": ["object", "null"],
      "disableKeyCompression": true,
      "properties": {
        "href": {
          "type": ["string", "null"],
          "disableKeyCompression": true
        },
        "total": {
          "type": "number"
        }
      }
    },
    "genres": {
      "type": "array",
      "uniqueItems": true,
      "item": {
        "type": "string"
      }
    },
    "href": {
      "type": "string"
    },
    "images": {
      "type": "array",
      "uniqueItems": true,
      "item": {
        "type": ["object", "null"],
        "disableKeyCompression": true,
        "properties": {
          "height": {
            "type": "number"
          },
          "url": {
            "type": "string"
          },
          "width": {
            "type": "number"
          }
        }
      }
    },
    "name": {
      "type": "string"
    },
    "popularity": {
      "type": "number"
    },
    "type": {
      "type": "string"
    },
    "uri": {
      "type": "string"
    }
  }
}
```

joehannes:
second one simply happens on a save it seems, which automatically happens on insert/upsert I think.
At least I never save, but upsert only and as soon as I upsert in [this](http://www.github.com/joehannes-jobs/compucorp) project I get the error (without the hotfixes for compression, decompression) ...
This is because obj-key compression/decompression happens on reading/writing the localstorage and it doesn't check against null, although defined in the schema.
So it tries to check if it is an object and says yeah typeof null === 'object' .. .yeah, true ... then it says, but wait, is it a function, is it this and that, and eventually says, no, none of that, error ...
I also tried to "disableKeyCompression" in the schema, which seems not to be implemented yet ...

pubkey:
`disableKeyCompression` is implemented but must be used on the top-level of your schema, not in a single field.

joehannes:
ok, sorry for that, I'm really a noob to RxDB :)
however, does this help?
I think this still is very uncomfortable.
Say you don't think about your obj property being possibly null apart from adding `["obj", "null"]` to the type and then get a keycompression err ... it will cost you a sigh and some time to find the solution, don't you think?

pubkey:
I'm closing this for now.
I added a notice to the docs that disableKeyCompression can only be used at the top-level. [commit](https://github.com/pubkey/rxdb/commit/a6b35318308fc13dca5dcea2701cbc526b18bff2).
I also added the null-check [commit](https://github.com/pubkey/rxdb/commit/24a1e7d0816fd6bc2c9fdb0cb13c7973f2877278)

I don't understand what [this](https://github.com/pubkey/rxdb/pull/183/commits/afe7e9908fd01268a8a5bf9ca46154d395749b1a) does. Please commit a test-case to reproduce the bug which this fixes. Notice to not commit dist-files.

joehannes:
I've got it out in the wild at http://joehannes-jobs.github.io/compucorp ...
This uses my forked repo with that fix that you say you don't understand.

I don't remember too exactly what and why, but in short:
In the schema you can define props as primary: true which makes them primary key, right?
That said, since they are props they are not top-level, but your check at this position ensures that
only top level can be primary ... I don't get it, thus no data-prop can be primary since data-props aren't top level, but first level after that ... (defining top level as eg with disableKeyCompression)
thus you can't have primary keys, thus you can't do queries ...
Either I'm misunderstand or ...??

Anyway, you can easily test it ...
Clone http://github.com/joehannes-jobs/compucorp ... master ... and replace the forked RxDB with your original repo ... should through an error that primary keys must be toplevel

Cheers

joehannes:
As an example to what made me think this should work like that:
https://github.com/pubkey/rxdb (README)
expand the SCHEMA section and look at the first level name prop, its got a primary: true prop ...
using this schema that way should throw bespoke error ...
Though looking at the code the fix would be to remove the check for `if(primary && toplevel) throw error`

pubkey:
Looks like there is something wrong with the documentation.
`primary` must be defined on a top-level field, not on the top-level of the schema

pubkey:
The given schema works [here](https://www.webpackbin.com/bins/-Kl_UJwIRc7wysOY96PZ) as expected.

joehannes:
The error shows for me on trying to [upsert](https://www.webpackbin.com/bins/-KmSXHGNq6o0aNW23zqT) (please clear the local cache first) ...

pubkey:
@joehannes which browser do you have?
It works in mine (chrome)

PR #197, opened 2017-06-05
Split out watchForChanges to support external replications

natew:
This allows me to implement query based sync until its more officially supported.

### DIFF
--- a/src/RxCollection.js
+++ b/src/RxCollection.js
@@ -372,6 +372,54 @@ class RxCollection {
         return Promise.all(importFns);
     }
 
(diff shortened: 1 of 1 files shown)


pubkey:
I merged this but had to revert one thing. You removed the following line
`.filter(doc => !this._changeEventBuffer.buffer.map(cE => cE.data.v._rev).includes(doc._rev))`
but removing this breaks the tests. Did this happen on purpose, and if yes, what was not working with this filter?

natew:
Oh strange, did not notice I had removed that! No reason other than my error.

PR #201, opened 2017-06-12
fix react-native example up to date

iwater:


pubkey:
Thank you very much. Merged.

PR #203, opened 2017-06-17
Nice links

natew:


pubkey:
It looks like the webpack-bin got overwritten when I was testing around. I will add a better one in the next days

PR #205, opened 2017-06-17
Typescript improvements for CollectionCreator

YKalka:
I think it is useful to export the CollectionCreator interface as well in typescript.


pubkey:
Great commit.
I think we should rename `CollectionCreator` to `RxCollectionCreator` and use the same name internally as in the export.

YKalka:
Thanks for your feedback. I renamed the interface

Issue #235, opened 2017-08-15
Content-Security-Policy Error on validate schema


willyelm:
**EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive**

Since module `is-my-json-valid` is using function generators to validate the schema when the [validator function](https://github.com/pubkey/rxdb/blob/5f9fabd1a27df8db58fa7104b765e8aa8aadee52/src/RxSchema.js#L120) runs on a secured server using a CSP. it does throw an error every time.

I think for server side usage it would be fine to keep using `is-my-json-valid` but use another when running it from the browser.

pubkey:
Hi @willyelm .
Can you post some code to reproduce this problem? Maybe it also helps to describe your environment.
I never had issues with the validation, neither on browsers nor on node.

For future issues, please use the issues-template :)

willyelm:
Hi @pubkey,

Thank for the response you can reproduce by adding the following the header to emulate the server in webpack.

If you add 
headers: {
  'Content-Security-Policy': 'default-src \'self\' \'unsafe-inline\''
}
inside your [devServer config](https://github.com/pubkey/rxdb/blob/b5de998d6ab57564356334390996cbaf1dc7d88f/examples/angular2/webpack.config.dev.js#L29)
 
Then run the server any unsafe code like eval, new Function, Function.* will be rejected.

I've tried other schema validators and seems like `[z-schema](https://github.com/zaggino/z-schema)` works with no issue.

pubkey:
Hi @willyelm 
I realeased 5.3.0 a few minutes ago.
You can now create your own validation-plugin with `z-schema` instead of `is-my-json-valid`.
The easiest way would be to start with a copy of [this file](https://github.com/pubkey/rxdb/blob/master/src/modules/validate.js) and replace the validation-dependency.

pubkey:
Closing this for now. A PR with an alternative validation-module is welcomed.

phil-lgr:
@willyelm I fixed this FYI

PR #243, opened 2017-08-17
Fix CSP Issue by use z-schema as validator

willyelm:
## This PR contains:
A BUGFIX
JSON Schema Validator fails when running on a server with content security policy enabled, error sample:
**EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive**

related issue #235

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,6 @@
     "clone": "^2.1.1",
     "crypto-js": "^3.1.8",
     "deep-equal": "^1.0.1",
-    "is-my-json-valid": "^2.16.0",
     "modifyjs": "0.3.1",
     "object-path": "0.11.4",
     "pouchdb-core": "6.3.4",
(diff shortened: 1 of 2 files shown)


pubkey:
Hi @willyelm 
The CI is broken because of the import-changes.
Before I switched from ajv to is-my-json-valid, I made many tests for performance and build-reasons.
The [benchmark](https://rawgit.com/zaggino/z-schema/master/benchmark/results.html) of z-schema says it's a lot slower than is-my-json-valid. I don't know if we should just switch the schema-engine to solve an edge-case.

pubkey:
I'm closing this now. The ci is broken and it would be better to create a plugin for this. More info [here in the issue](https://github.com/pubkey/rxdb/issues/235#issuecomment-325036673)

PR #321, opened 2017-10-11
Improve typings for collections and queries

ihadeed:
## This PR contains:
IMPROVED typings

Let's start with a code example:
```ts
    // this.collection refers to RxCollection<Customer>
    const customers = await this.collection.find().exec();
    customers.forEach(customer => {
       // do something
    });
```

Without this PR, I would have the following issues/inconveniences:

#### ( 1 ) - Issues with `find` and `findOne` methods. 
**Problem**: IDE/Compiler will complain that `.forEach` does not exist on type `Customer | Customer[]`. The `|` operator in TypeScript works as expected when we're writing data that follows an interface/type. However, it doesn't work the same way when we're reading data. 

**Solution**: We can pass a type to `RxQuery` from the `RxCollection` class. This type will be used as a return type for the query when we run `.exec()`. `findOne` can pass `Customer` and `find` can pass `Customer[]`. This way the IDE/Compiler knows that the method will return either a single document or an array. 

#### ( 2 ) - `.exec` returns an `RxDocument`, but the typing says it returns `RxDocumentType`
**Problem**: I understand you are using `RxDocumentType` as a return type so we can access the properties of our model right away. This is not the best way to handle it though, because the real object we're getting back is a mix of `RxDocument` and `RxDocumentType`. 

** Solution**: I created a type called `IRxDocument` that accepts a type as a param. It returns back `RxDocument & T`. Now the IDE/Compiler knows that when we call `.exec` we will get both `RxDocument & T`, and when we call `toJSON` we get `T`. 

## TODO
- I did not follow the naming conventions you have in the typings. I will leave it up to you to decide what you wanna name `IRxDocument`. I just used the first thing that came to mind, to get things working.
- Need to export `IRxDocument` if you decide to merge my other PR #320

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -168,12 +168,12 @@ declare class RxCollection<RxDocumentType> {
     schema: RxSchema;
 
     $: Observable<RxChangeEvent>;
-    insert(json: any): Promise<RxDocument>;
-    newDocument(json: any): RxDocument;
-    upsert(json: any): Promise<RxDocument>;
-    atomicUpsert(json: any): Promise<RxDocument>;
-    find(queryObj?: any): RxQuery<RxDocumentType>;
-    findOne(queryObj?: any): RxQuery<RxDocumentType>;
+    insert(json: any): Promise<IRxDocument<RxDocumentType>>;
+    newDocument(json: any): IRxDocument<RxDocumentType>;
+    upsert(json: any): Promise<IRxDocument<RxDocumentType>>;
+    atomicUpsert(json: any): Promise<IRxDocument<RxDocumentType>>;
+    find(queryObj?: any): RxQuery<IRxDocument<RxDocumentType>[]>;
+    findOne(queryObj?: any): RxQuery<IRxDocument<RxDocumentType>>;
 
     dump(decrytped: boolean): Promise<any>;
     importDump(exportedJSON: any): Promise<Boolean>;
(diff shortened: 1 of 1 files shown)


pubkey:
Hi @ihadeed 
Thank you for the great work. I'm not a daily typescript-user so I did the best I could :)
#320 is merged, so you can export the IRxDocument.

I think `IRxDocument` should be names something else.
One reason is that the prefix is always `Rx` so we should stick to that convention.
Also I think it's confusing what the letter `I` is standing for. Is this something like a convention in other libs typings?
Maybe `RxCustomDocument` is better? I don't know.

Another thing is that on line 262 you set `populate(objPath: string): Promise<IRxDocument<T> | any>;`
but the populate-method mostly returns a document that are from another collection and so does have a different structure.

ihadeed:
Thanks for the feedback @pubkey !

I think `IRxDocument` can be renamed to `RxDocument`, and the `RxDocument` class can be `RxDocumentBase`. We never have to use the base class alone anyways, every document has to extend a schema.

pubkey:
Yes. Thats a good point. Lets use `RxDocument`

pubkey:
Info: I added some tests for the typings at [this commit](https://github.com/pubkey/rxdb/commit/416196f4826f0594cec2c5870180822142fb4962)

Gosev:
Hi, I'd like to use this in Angular but from what I see it isn't pushed to npm. Is a release coming soon or should I use the github master branch directly instead ?

pubkey:
The release will take something like 2 more weeks.
If you need this, include rxdb via github with the hash of the last [build-commit](https://github.com/pubkey/rxdb/search?q=build&type=Commits&utf8=%E2%9C%93)

Gosev:
thanks :)

PR #323, opened 2017-10-11
FEATURE: accept a function as a default value

### LINKED ISSUE #322: [Feature request] Accept functions as default values
## Case
New feature

## Issue
It would be useful to pass a function as the default value. Here are two use cases:
```ts
import { generate } from 'shortid';

const MySchema = {
  type: 'object',
  version: 0,
  properties: {
    id: {
      type: 'string',
      primary: true,
      // Results in a new unique ID generated every time a new document is created
      default: generate
    },
    creationDate: {
      type: 'number',
      // Results in a timestamp every time a doc is created
      default: Date.now
    }
  }
}
```

ihadeed:
## This PR contains:
 - A NEW FEATURE

Closes #322

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Typings
- [ ] Changelog

### DIFF
--- a/src/rx-schema.js
+++ b/src/rx-schema.js
@@ -92,7 +92,12 @@ export class RxSchema {
             this._defaultValues = {};
             Object.entries(this.normalized.properties)
                 .filter(entry => entry[1].default)
-                .forEach(entry => this._defaultValues[entry[0]] = entry[1].default);
+                .forEach(entry => {
+                    let defaultValue = entry[1].default;
+                    if (typeof defaultValue === 'function')
+                        defaultValue = defaultValue();
+                    this._defaultValues[entry[0]] = defaultValue;
+                });
         }
         return this._defaultValues;
     }


ihadeed:
FYI ... I wasn't able to test this by compiling the repo. I started getting 409 errors. I'm assuming the master branch isn't stable yet. However, I did test it in my app by modifying `dist/es/rx-schema.js` and it worked well!

pubkey:
I fixed the tests a few hours ago. I think they should work now with your PRs code.

When I created the `default`-attribute, I thinked a lot about using the same syntax as mongoose, which also allows functions for default. But I had chosen to not allow functions for the following reason:
Json-schemas that are used with rxdb should always be stringified and parsed again.
This allows us to store the schema of a collection next to the documents and read out the schema afterwards.
This is done for example in the data-migrator. When you create a collection with a higher version-number, the schema of the 'older' documents is read out and used to parse the key-compression and handle other things.

A solution to have a function generating the default-value is to use the [postCreate](https://pubkey.github.io/rxdb/middleware.html#postcreate)-hook

PR #329, opened 2017-10-16
Improve typings

### LINKED ISSUE #326: Query object typings
Hi again!

I just wrote some new typings for this library, but I wanted to ask you a few questions before submitting a PR. Here is the code so far:
```ts
export interface RxQueryOptions<T> {
  $eq?: T;
  $gt?: T;
  $gte?: T;
  $lt?: T;
  $lte?: T;
  $ne?: T;
  $in?: T[];
  $nin?: T[];
  $regex?: RegExp;
  $exists?: keyof T;
}

export type RxQueryObject<T> = keyof T & {
  [P in keyof T]?: T[P] | RxQueryOptions<T[P]>;
} & {
  $or: RxQueryObject<T>[];
  $nor: RxQueryObject<T>[];
  $and: RxQueryObject<T>[];
};
```

It's meant to be used in `.where()`:
```ts
declare class RxQuery<RxDocumentType>{
  ...
  where(queryObj: RxQueryObject<RxDocumentType>): RxQuery<RxDocumentType>;
  ...
}
```

So now the IDE's autocomplete will show us the available query options every time we use `.where()`. The method will accept a string that matches on of the schema properties, or an object containing operators (`$or, $nor, $and`) and/or properties of the schema. If the method is passed an object with a property of the schema, it will accept a value (same type defined in schema) or an operator.

My questions:
1. Is this something you are ok with?
2. does `$exists` accept an array or just a single string?
3. what does `elemMatch` do? is it a query operator?
4. what does `all` do?
5. am I missing any operators?

ihadeed:
- Improve typings for queries (closes #326)
- Improve typings for collections (instead of #321)
- Add `.editorconfig` to make sure IDEs follow the repo's format
- Add `.idea` in `.gitignore` to remove Jetbrains IDE temporary files

**Todo**:
- [ ] Tests
- [ ] Approval from @pubkey

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+
+[*.{js,json,ts}]
+charset = utf-8
+quote_type = single
+spaces_around_brackets = true
+
+[*.json]
+indent_size = 2
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ package-lock.json
 config.json
 .config.json
 .transpile_state.json
+.idea
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,6 +1,6 @@
 import { Observable } from 'rxjs';
 
-declare class RxSchema {
+export declare class RxSchema {
     jsonID: SchemaJSON;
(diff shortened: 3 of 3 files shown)


pubkey:
I merged this. Thank you for the great work @ihadeed 
I tested this locally and will add some tests soon.

PR #330, opened 2017-10-16
Fix typo in README

piperchester:
## Todos 
- [ ] Tests
- [ ] Documentation
- [ ] Typings
- [ ] Changelog

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@
 </p>
 
 <p align="center">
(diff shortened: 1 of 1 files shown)


pubkey:
[Wikipedia](https://en.wikipedia.org/wiki/JavaScript) says it is called `JavaScript` with a uppercase `S`. Same [here](http://web.archive.org/web/20070916144913/http://wp.netscape.com/newsref/pr/newsrelease67.html)

PR #355, opened 2017-11-08
Remove babel-polyfill from peerDependencies

ihadeed:
Since babel-polyfill isn't always needed, it shouldn't be listed in the peerDependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,6 @@
     "lint"
   ],
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


pubkey:
Merged. I improved the docs [here](https://github.com/pubkey/rxdb/commit/9a25623e334b2bce613ccec7f2f332bf559fccf0) incase someone does not know what polyfills are.

PR #356, opened 2017-11-10
Improve RxError typings

ihadeed:


ihadeed:
1. Do we know all the possible properties?
If yes, we can add them all to the interface. That shouldn't be an issue since they're all readonly anyways. It's better than having no typings at all.

2. Are they completely random? 
If yes, we can add whatever we have, and then provide a union type `RxErrorParameters | any` or `RxErrorParameters & any`. This will show the properties we provided in the IntelliSense, and still allow accessing different properties.

This isn't a big deal but it helps write code faster without double checking everything w/ docs. My use case is as follows:
```ts
try {
  // running methods that might throw errors
  // running RxDB queries/operations
} catch (e) {
  if (e.rxdb) {
     (e as RxError).errors; // access RxError properties
  } else {
      // handle regular Error class
  }
}
```

pubkey:
Ok you are right.
We know all possible properties. I will do a commit with the other ones.

PR #363, opened 2017-11-13
Update docs - clarify collection selection from database.

rybaczewa:
## This PR contains:
Improved docs

Further clarification on collection select in docs. Mentioned it on Gitter, fixed here: https://github.com/pubkey/rxdb/commit/01b8e9fd6d6a5e8c266ffd1154a59f57fd0e4db0

Code just underneath `Get a collection from the database` might suggest that first line is a way to get collection, not create it. It looked like two alternative methods for accessing collection. Might be to specific though, leaving decision to maintainer.


pubkey:
You are right. Thank you :+1:

PR #367, opened 2017-11-18
Fix: Typescript-Definition RxQuery::where argument

FlorianKoerner:
The Argument for RxQuery::where can be a RxQueryObject or key of RxDocumentType.


FlorianKoerner:
There are multiple issues with the new typescript definitions for me. I'm going to put some more time into this merge request.

PR #368, opened 2017-11-18
Fix TypeScript Definitions for RxQuery

FlorianKoerner:
RxQuery must differ between model and result to get typeof working. And `where` also allows a key from the model.

### DIFF
--- a/src/typings/rx-collection.d.ts
+++ b/src/typings/rx-collection.d.ts
@@ -62,7 +62,7 @@ export interface SyncOptions {
     },
     // for options see https://pouchdb.com/api.html#replication
     options?: PouchReplicationOptions,
-    query?: RxQuery<any>
+    query?: RxQuery<any, any>
 }
 
 export declare class RxCollection<RxDocumentType> {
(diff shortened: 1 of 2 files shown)


ihadeed:
Ahh good call. I just looked through the code and realized it doesn't make sense. `.find()` wouldn't be able to get the keys from the model since we're passing an array as a type.

ihadeed:
`SyncOptions` interface in `rx-collection.d.ts` need to be changed. 

The `query` property has the type `RxQuery<any>` assigned to it. Needs to be changed to `RxQuery<any, any>`.

Alternatively, we can remove `<any>` from there, and provide default types to `RxQuery` class declaration:
```ts
export declare class RxQuery<RxDocumentType = any, RxQueryResult = any> {
```

FlorianKoerner:
@ihadeed Thanks for the hint. Fixed.

pubkey:
Thx @FlorianKoerner for the PR. The ci is now working again, please merge the master into this PR.

I see one more problem with nested documents. For example the following does not work:

```typescript
type DocType = {
  age: number,
  nestedObject: {
    foo: string,
    bar: number
  }
};
const myCollection: RxCollection<DocType> = await myDb.collection({
  name: 'humans',
  schema: {},
  autoMigrate: false,
});

const query = myCollection.findOne().where('nestedObject.foo').eq('foobar');
```

FlorianKoerner:
@pubkey solved "nestedObject.foo" with `keyof RxDocumentType | string`. So you can get an autocompletion and go with other values also.

pubkey:
Thx @FlorianKoerner  for the great work. Merged.

PR #397, opened 2017-12-08
Change JSONSchema validator from is-my-json-valid to ajv

rybaczewa:
## This PR contains:
Potentially breaking change, few new tests, package change

At the moment, is-my-json-valid is not validating schemas with definitions properly. It causes invalid schemas to pass validation where it shouldn't. Ajv is faster and better maintained, much better coming into the future of the project.

Extra parsing function added for backward compatibility of errors.

### DIFF
--- a/docs-src/custom-build.md
+++ b/docs-src/custom-build.md
@@ -21,7 +21,7 @@ Some parts of RxDB are not in the core, but are required. This means they must a
 ### validate
 
 The validation-module does the schema-validation when you insert or update a `RxDocument`. To use RxDB you always have to add a validation-module.
-This one is using [is-my-json-valid](https://www.npmjs.com/package/is-my-json-valid) but you can also use your own validator instead. To import the default validation-module, do this:
+This one is using [ajv](https://www.npmjs.com/package/ajv) but you can also use your own validator instead. To import the default validation-module, do this:
 
 ```javascript
 // es6-import
--- a/package.json
+++ b/package.json
@@ -63,12 +63,12 @@
   },
   "dependencies": {
     "@types/core-js": "0.9.43",
+    "ajv": "^5.5.1",
     "babel-runtime": "^6.26.0",
     "clone": "^2.1.1",
     "crypto-js": "^3.1.8",
     "custom-idle-queue": "1.0.0",
     "deep-equal": "^1.0.1",
-    "is-my-json-valid": "2.16.1",
     "modifyjs": "0.3.1",
     "object-path": "0.11.4",
     "pouchdb-core": "6.3.4",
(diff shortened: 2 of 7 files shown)


pubkey:
Hi @rybaczewa thank you for the great PR.
The code looks good. But if have some problems:

1. The build-size is much bigger now:
(run `npm run disc` to verify)
```bash
before:

- disc: 408.9kB
- gzip: 156.8kB


after:

- disc: 489.6kB
- gzip: 174.3kB
```

2. It looks like this is a breaking change because `required: true` seems not to works anymore [see](https://github.com/pubkey/rxdb/pull/397/files#diff-89ecf35bad6d7b9f07c91ea99ab53013L259). Is this correct?


I think we should not swap out the default validator by now. It will be better to create a new plugin `ajv-validate` so users can cherry-pick it if they want. Then we can swap out the default validator afterwards if wanted.

rybaczewa:
If it comes to `required` keyword, that's another place where is-my-json-valid wasn't respecting JSONSchema specification. If specified, `required` keyword must be an array of unique string values. Can't be boolean.
http://json-schema.org/latest/json-schema-validation.html#rfc.section.6.5.3

Build size increase sounds about right looking at package sizes:
- [ajv](https://bundlephobia.com/result?p=ajv@5.5.1)
- [is-my-json-valid](https://bundlephobia.com/result?p=is-my-json-valid@2.16.1)

Generally speaking, the more I looked at old validator package, the less I liked it. Separate validator seems OK for now, but it might more beneficial for the future to migrate to package that does it's job. I am by no means pushing exactly ajv, just anything that works (and it's looking like the best choice for now).

pubkey:
You are right. I currently also regret the switch to is-my-json-valid and we should move to ajv in the future.
I will now create an ajv-validate-plugin from your code. And we I do the next major-release, I will move to ajv as breaking change.

pubkey:
Closed in favor of [20063b960ba87e9a7a561f45d56271aafef6a42e](https://github.com/pubkey/rxdb/commit/20063b960ba87e9a7a561f45d56271aafef6a42e)

pubkey:
Hi @rybaczewa 
I am working on RxDB v8.0.0 atm.
As described in the [before-next-major-todo-list](https://github.com/pubkey/rxdb/blob/master/orga/before-next-major.md), it was planned to switch to ajv by default.
I switched to ajv and checked the result, but now I am not sure if switching to ajv is a good thing because of the following reasons:

- Build size increased by 80kB, minified+gziped (run `npm run disc` to validate)
- Spawn Database Time increased by 9% (run `npm run test:performance` to validate)
- Insertion of documents increased by 6%

Yes, ajv is more compliant with the json-schema-standard, but I do currently not think that this is worth the disadvantages.

Any comment or argument would be welcomed.

Issue #157, opened 2017-05-08
Error: Cannot sort on field(s) "XXX" when using the default index


matsuo-ts:
## Case
Bug on `find()` and `findOne()`

## Issue
@pubkey  As I reported in #154 , after updating to `4.0.0` , error occurring in collection query.

```
Error: Cannot sort on field(s) "XXX" when using the default index
    at validateSort (bundle.js:165)
    at bundle.js:165
(anonymous) @ bundle.js:244
```
This error does not happen in version `3.0.2`.
Maybe this comes form [PouchDB](https://github.com/pouchdb/pouchdb/issues/6399)

## Info
- Environment: (Node.js/browser/electron/etc.)
  - Electron(1.6.7) 
  - React(15.4.2) 
  - Redux(3.6.0) 
  - React Router(3.0.2) 
  - Webpack(2.2.1) 
  - ES6

- Adapter: (IndexedDB/Localstorage/LevelDB/etc.)
WebSQL

## Code
```javascript
RxDB.create({
      name: DB_NAME,
      password: PASSWORD,
      adapter: 'websql'
    })
    .then( db => {
      return db.collection({
        name: COLLECTION_NAME,
        schema: COLLECTION_SCHEMA
      });
    })
    .then( collection => {
      collection
        .findOne( KEY1 )
        .where( KEY2 )
        .eq( VALUE )
        .exec()
        .then( result => {
          if ( result === null ) {
            // CODE FOR TRUE
          }
          else {
           // CODE FOR FALSE
         }
        })
        .catch( err => {
          console.error( 'Failed to open collection: ', err );
        });
    })
    .catch( err => {
      console.error( 'Failed to open Database: ', err );
    });
```

pubkey:
Is it possible to post your schema? This would help very much to reproduce.

matsuo-ts:
@pubkey Sorry for less information. Here is my schema for this collection.

```javascript
export const COLLECTION_SCHEMA = {
  'disableKeyCompression': true,
  'title': 'User Preferences',
  'version': 0,
  'description': 'DESCRIPTION',
  'type': 'object',
  'properties': {
    'user_id': {
      'type': 'string',
      'primary': true
    },
    'user_pwd': {
      'type': 'string',
      'encrypted': true
    },
    'last_login': {
      'type': 'number'
    },
    'status': { 
      'type': 'string'
    }
  },
  'required': [ 'user_pwd', 'last_login', 'status' ]
};
```
[...]

PR #438, opened 2017-12-22
Add more tests for issue #157.

rybaczewa:
## This PR contains:
Improved tests

Issue https://github.com/pubkey/rxdb/issues/157 still not resolved. Added a test that replicates that.
Using newest version (7.3.0)

Modified https://github.com/pubkey/rxdb/pull/434 with `disableKeyCompression`


pubkey:
@rybaczewa thank you for the test. It works now.
The problem was that only `string` and `integer`-fields where allowed as index. I now added `number`-fields.

PR #528, opened 2018-02-16
Test case and fix for default value bug

gvuyk:
## This PR contains:
 - IMPROVED TESTS
 - A BUGFIX

A default value of 0 is ignored
I'm assuming the expected behavior is for the default value to be stored regardless of its value, except when its null/undefined or limited by the "type"


pubkey:
Thank you @gvuyk for that great PR :+1: 
I added the test and a modified version of your fix.

Issue #558, opened 2018-03-09
preinstall script not working on cmd / powershell


adam-lebon:
## Case
Bug

## Issue
When you install rxdb from a windows system with the default shells (cmd or powershell) the preinstall script is the following in the package.json:
https://github.com/pubkey/rxdb/blob/1fe0f69e42fbfe12756bcd0d30828976bbd0e60e/package.json#L32
But in Windows system the ```-f``` argument does not exist on powershell. Worse, when you are using cmd the rm command simply doesn't exist (del is the equivalent)

## Info
  - Environment: (windows)
  - Adapter: (N.A.)
  - Stack: (N.A.)

## Step to reproduce

```shell
PS > yarn add rxdb
error C:\Users\****\Documents\Git\********\node_modules\rxdb: Command failed.
Exit code: 1
Command: rm -f ./package-lock.json
Arguments:
Directory: C:\Users\Adam\Documents\Git\NSRP-SERVER\node_modules\rxdb
Output:


*Blank output but the issue is that -f doesn't exist on powershell*


```
When you install the package they are no package lock so the rm command looks useless for fresh installation.

## """Solution"""
Use a cygwin / cmder / babun whatever, a shell emulator to get the rm command.

## Fix
Check the utility of the rm command and remove if possible

pubkey:
You are right. The preinstall-command was intended for dev-mode
 I forgot that it will also run when rxdb is installed as dependency.

pubkey:
Thx for the help. I released 7.4.1 with the fix.

PR #559, opened 2018-03-09
Removing the preinstall script to support windows shells

adam-lebon:
## This PR contains:
 - A BUGFIX

The preinstall script assume that all dev got the rm commands on their system
But on the cmd, rm does not exist
In powershell the rm command exist but not the -f arguments

https://github.com/pubkey/rxdb/issues/558


pubkey:
Thank you very much. I will do a new release tomorrow

PR #606, opened 2018-04-11
Fix RxChangeEvent typings

hubgit:
## This PR contains:

Improved typings.

This is a best guess at current typings for the [RxChangeEvent](https://github.com/pubkey/rxdb/blob/master/src/rx-change-event.js) class.

### DIFF
--- a/src/typings/rx-change-event.d.ts
+++ b/src/typings/rx-change-event.d.ts
@@ -1,6 +1,17 @@
+import { RxDocument } from 'rxdb/src/typings/rx-document';
+
+export interface RxChangeEventData {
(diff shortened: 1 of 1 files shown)


pubkey:
Hi @hubgit thx for the PR.
It looks like this makes a problem in the ng-example. `import { RxDocument } from 'rxdb/src/typings/rx-document';`
```bash
ERROR in [at-loader] ../../src/typings/rx-change-event.d.ts:1:28 
    TS2307: Cannot find module 'rxdb/src/typings/rx-document'.
Child html-webpack-plugin for "index.html":
```
You should import everything relative.

PR #618, opened 2018-04-23
Improve RxChangeEvent typing

hubgit:
This contains some improvements to the RxChangeEvent typing for things that were missed in https://github.com/pubkey/rxdb/pull/606

* Use a named type for the operation strings.
* Allow the type of the `v` object to be defined, optionally.
* `doc` is a string (the primary key), not a document.


pubkey:
CI says:
```bash
ERROR in [at-loader] ../../src/typings/rx-collection.d.ts:80:28 
    TS2314: Generic type 'RxChangeEvent<T>' requires 1 type argument(s).
```

Issue #630, opened 2018-05-06
Query cache is not being invalidated by replication


gvuyk:
## Case
Bug

## Issue
If a query is made, then new docs are added via replication, then the same query is made again, the results won't contain the new docs added by replication

## Info
  - Environment: Node.js

pubkey:
It looks like the completed-event is fired when the replication is not completed yet.
If i add `await AsyncTestUtil.wait(10);` before the second query, it works.

gvuyk:
yeah that could be it, measured the times and in absence of the first query, the second takes around 6 ms more to complete, thus allowing the replication to finish

gvuyk:
Strange thing is, I tried adding the extra wait time on a script outside the test environment and it didn't work, but destroying the queryCache did
```js
const RxDB = require("rxdb");
RxDB.plugin(require("pouchdb-adapter-memory"));

async function test()
{
	// Create a schema
	const schema = {
		version: 0,
		disableKeyCompression: true,
		type: "object",
		properties: {
			name: {type: "string"},
			number: {type: "number"}
		}
	};

	// Create dbA
	const dbA = await RxDB.create({ name: "dba", adapter: "memory" });

	// Create collection A
	const collectionA = await dbA.collection({ name: 'result', schema: schema });

	// Insert documents
	await collectionA.insert({ name: 'aaaa', "number": 1 });
	await collectionA.insert({ name: 'bbbb', "number": 2 });

	console.log("initial collection A: "+(await collectionA.find({}).exec()).length);

	//----------------------------------------------------------------------------

	// Create dbB
	const dbB = await RxDB.create({ name: "dbb", adapter: "memory" });

	// create collection B
	const collectionB = await dbB.collection({ name: 'result', schema: schema });

	// Pull from collection A
	const pullstate = collectionB.sync({
		remote: collectionA.pouch,
		direction: {pull: true, push: false},
		options: {live: false}
	});
	
	// Wait for replication to complete
	await new Promise((resolve, reject) => {
		pullstate.complete$.subscribe(completed => {
			if(completed){
				if(completed.ok == true){
					resolve();
				}else{
					reject(completed.errors);
				}
			}
		});
		pullstate.error$.subscribe(error => {reject(error);});
	});
	await new Promise(r => { setTimeout(r,100); });
	console.log("collection B after pull: "+(await collectionB.find({}).exec()).length);

	// Delete 1 doc from collection B
[...]

PR #641, opened 2018-05-13
added failing test with 2 replications

gvuyk:
## This PR contains:
Test case

## Describe the problem you have 
#630

### DIFF
--- a/test/unit/bug-report.test.js
+++ b/test/unit/bug-report.test.js
@@ -1,106 +1,77 @@
-/**
- * this is a template for a test.
- * If you found a bug, edit this test to reproduce it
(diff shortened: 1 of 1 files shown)


pubkey:
Hi @gvuyk Thank you for the test.
I think the problem is that you directly use the collections pouchdb-instance as `remote`. Please use the collection itself, than it works for me.

Instead of:
`remote: collectionA.pouch,`

Use:
`remote: collectionA,`

The reason is that rxdb does not per default listen to the pouchdb's changes because it has a performance-cost. But when the collection is used in a replication, rxdb starts listening.

gvuyk:
Nice! Can confirm the problem is gone
Thanks again

PR #646, opened 2018-05-17
Skip defining getter and setter when property not defined in schema

hubgit:
## This PR contains:

A BUGFIX?

I'm using RxDB in an app that defines only a minimal schema for each object, as there are lots of different types of object stored in the same collection. I've found that calling `get(x)` on a document, when the property at `x` is an object or array, calls `_defineGetterSetter` which tries to get the schema for the property, which is null when there's no schema defined for this path, causing an exception to be thrown.


pubkey:
Hi @hubgit 
thank you for the PR.
Can you add a test-case that fails without this change?
Just to ensure this still works after refactors. Thank you.

PR #667, opened 2018-06-01
Adds option for providing `_id` when inserting documents, adds tests.

will118:
## This PR contains:
 - an option to the `RxCollection` ctor allowing `_id` to be user defined

Is there a reason `_id` cannot be user defined?

I looked through the git history and seems that at one point this was an optional argument to `RxCollection.insert()`.

Is there a reason it was removed? A per-collection option suits my use case but passing an argument to `insert()` wouldn't be the end of the world.

I believe there is already a `positive` test for inserting a doc without `_id`, and a `negative` test for inserting a doc with `_id`.

I've added a test for each of those cases but with this option set.

If this is a change you're happy with I can make changes/add docs etc.

### DIFF
--- a/src/plugins/error-messages.js
+++ b/src/plugins/error-messages.js
@@ -68,6 +68,7 @@ const CODES = {
     COL16: 'given static method is not a function',
     COL17: 'RxCollection.ORM: statics-name not allowed',
     COL18: 'collection-method not allowed because fieldname is in the schema',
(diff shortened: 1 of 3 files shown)


pubkey:
This was intentional. If you want to set the value of the _id field, you can create a primary-key at your schema. Then you can set this value and it will be stored in the `_id`.

will118:
Ah I see. Thanks.

PR #669, opened 2018-06-01
Update rx-collection.html; Fixed typo

Endran:
There was a typo:
was _pouchdb instances or remo_ *v* _e servers which_
fix _pouchdb instances or remo_ *t* _e servers which_


pubkey:
Hi Endran. The docs-folder is generated from docs-src. Please only change files in docs-src.

PR #723, opened 2018-07-21
fix #717

### LINKED ISSUE #717: Field names can't be one character long.
## Case

Bug

## Issue

I'm trying to define a JSON schema with a type like:
```js
{ type: 'object',
  properties: {
    point: {
      type: 'object',
      properties: {
        x: { type: 'integer' },
        y: { type: 'integer' }
      }
    }
  }
}
```

It fails, though, because `x` and `y` don't match the regex `^[a-zA-Z][[a-zA-Z0-9_]*]?[a-zA-Z0-9]$`.

Can it be changed to something like `^[a-zA-Z](?:[[a-zA-Z0-9_]*]?[a-zA-Z0-9])?$` so only the first letter is mandatory?

errorx666:
## This PR contains:
A BUGFIX

https://github.com/pubkey/rxdb/issues/717

## Todos 
- [x] Tests


pubkey:
Hi @errorx666 
This PR looks good.
Can you merge the current master into this (and the other PRs).
There was a problem with the dependencies which is why the CI failed on your PRs. Merging would show that everything works fine.

PR #740, opened 2018-08-01
some advanced query examples

Celludriel:
- find by using sql equivalent '%like%' syntax
- find using a composite statement eg: $or
- do a case insensitive search


pubkey:
Awesome, merged.

Issue #754, opened 2018-08-14
inMemory collections don't sync up removals


rafamel:
## Case
Bug; inMemory collections don't sync up the removal of documents.

## Issue
When a doc is removed on an inMemory, it is still present in its original collection, and also vice versa.

## Info
  - Environment: Browser
  - Adapter: idb, memory
  - RxDB: 7.7.1

## Code

```js
const RxDB = require('rxdb');
RxDB.plugin(require('pouchdb-adapter-idb'));
RxDB.plugin(require('pouchdb-adapter-memory'));

async function test() {
  const db = await RxDB.create({
    name: 'testdb',
    adapter: 'idb',
    multiInstance: false
  });

  const schema = {
    version: 0,
    type: 'object',
    primaryPath: '_id',
    properties: {
      name: {
        type: 'string'
      }
    }
  };

  const collection = await db.collection({
    name: 'person',
    schema: schema,
    statics: {
      one() {
        return this.findOne();
      }
    }
  });

  // Insertion
  console.log('-'.repeat(20));
  const inMemCollection = await db.collections.person.inMemory();
  collection.insert({ name: 'hi' });
  // Wait
  await new Promise((resolve) => setTimeout(resolve, 1000));
  const inserted = await inMemCollection.findOne().exec();
  console.log(
    'Name in sync with inMemory?',
    inserted && inserted.name === 'hi'
  );

  // Removal from collection
  console.log('-'.repeat(20));
  console.log('Removing from collection');
  await collection
    .findOne()
    .exec()
    .then((x) => x.remove());
  // Wait
  await new Promise((resolve) => setTimeout(resolve, 2000));
  console.log(
    'Was removed from collection?',
    (await collection.findOne().exec()) === null
  );
  console.log(
    'Was removed from inMemory collection?',
    (await inMemCollection.findOne().exec()) === null
  );
  console.log('-'.repeat(20));
}
test();
```

rafamel:
This might be related to #761

pubkey:
Ok this was a hard one :)

First problem was that as described in #761 `.remove()` did delete all data of the document. I always thought remove would do the same as only setting `_deleted: true` which was wrong, I changed it and we now use `pouch.put()` and only set the _deleted-flag

Also there was a bug in the query-change-detection which is fixed now.

Because `pouch.bulkDocs` does not work for deleting documents silently, I added a workaround which could be removed in the future when the added pouch-db-integration-test fails.

rafamel:
[...]

PR #776, opened 2018-08-29
Fix query change detector for REMOVE events

hubgit:
## This PR contains:

 - A BUGFIX

We've found that sometimes a query subscriber isn't called when a document is deleted, only when RxDB.QueryChangeDetector is enabled. It seems that the `handleSingleChange` method isn't setting variables correctly when handling "REMOVE" events.

I've tried to update the code to match the comments that were already there, but I'm not sure how correct the comments were, or how best to test the asynchronous query subscribers.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Typings
- [ ] Changelog

### DIFF
--- a/src/query-change-detector.js
+++ b/src/query-change-detector.js
@@ -107,26 +107,26 @@ class QueryChangeDetector {
 
         if (changeEvent.data.op === 'REMOVE') {
             // R1 (never matched)
-            if (!doesMatchNow) {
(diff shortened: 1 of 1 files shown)


hubgit:
Closing as it looks like this was already described and fixed recently in https://github.com/pubkey/rxdb/issues/754

PR #777, opened 2018-08-29
Improve query change detector for REMOVE events

hubgit:
## This PR contains:

 - A BUGFIX

We found that sometimes a query subscriber isn't called when a document is deleted, only when RxDB.QueryChangeDetector is enabled. It seems that the `handleSingleChange` method isn't setting variables correctly when handling "REMOVE" events.

The main bug was fixed in https://github.com/pubkey/rxdb/issues/754 but I think it can be simpler: if the document previously matched the query but has now been removed, then it should always be removed from the results regardless of whether it now matches the query.

This allows the query change detector to handle deleted documents that no longer match the query as they don't have any of their previous properties.

I've tried to update the code to match the comments that were already there, but I'm not sure how correct the comments were, or how best to test the asynchronous query subscribers.

## Todos
- [ ] Tests
- [ ] Changelog

### DIFF
--- a/src/query-change-detector.js
+++ b/src/query-change-detector.js
@@ -109,41 +109,40 @@ class QueryChangeDetector {
 
         if (changeEvent.data.op === 'REMOVE') {
             // R1 (never matched)
-            if (!wasDocInResults && !doesMatchNow) {
+            if (!wasDocInResults) {
                 DEBUG && this._debugMessage('R1', docData);
                 return false;
             }
 
             // R2 sorted before got removed but results not filled
-            if (options.skip && doesMatchNow && sortBefore() && !isFilled) {
+            if (options.skip && sortBefore() && !isFilled) {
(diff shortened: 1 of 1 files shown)


pubkey:
Thank you for the investigation. I think we need more tests for the query-change-detection.

One problem I see with strategy R3:
- When we have a query with `skip: 3`
- One of the first document, which was not in the results, gets removed
- We now will returns the previous results again
- But correct would be to rexecute the query over the database to find the 3rd document which would be now in the result-set.

pubkey:
I'm closing this since it looks inactive. Please reopen if the work continues.

hubgit:
@pubkey You're right about the case where an item is deleted that was earlier in the results list than the current slice - I hadn't considered that, and it does mean the naive approach in this PR wouldn't be enough. As the initial bug is fixed, I'm happy to keep this closed.

PR #807, opened 2018-09-14
Fixes react-native example

rafamel:
- Changes name and versioning in `package.json`.
- Changes dependency versions.
- Removes unnecessary dependencies and files. Among them, it doesn't install `node` locally (travis is already running node@10).
- RxDB is imported from root, rather than installed locally (dependencies get messy and prone to errors).
- Uses yarn.lock - npm has trouble resolving the right dependencies in depth.
- Modifies travis.yml to run `npx yarn install` for the react-native example install.
- Modifies `README.md` to reflect install changes (yarn).


pubkey:
Awesome. Merged. Thank you so much.

Issue #763, opened 2018-08-21
replicationState.active$ emits false after pending changes have synced


rafamel:
## Case
Possible bug/FR.

## Issue
As per the docs replicationState.active$ "Emits true or false depending if the replication is running. For example if you sync with a remote server and the connection dies, this is false until the connection can be reestablished."

However, it doesn't hold a `true` value but rather turns `false` as soon as changes have been replicated.

In the same way, when the connection to the remote database is lost, `replicationState.error$` doesn't emit any error.

If this is intended behavior, I believe `$active` doesn't offer any further information to `$completed`, and there is no way for us to know when the connection is active or lost in a reliable manner. In that case, this would be a FR for a subscription that allowed us to - other than manually checking whether recent changes have been synchronized.

## Info
  - Environment: Browser
  - Adapter: idb, memory
  - RxDB: 7.7.1

pubkey:
From the code, the active$-state is just a mapping to the `active` and `paused` events of the pouchdb eventemitter.

```js
        // active
        this._subs.push(
            fromEvent(evEmitter, 'active')
            .subscribe(() => this._subjects.active.next(true))
        );
        this._subs.push(
            fromEvent(evEmitter, 'paused')
            .subscribe(() => this._subjects.active.next(false))
        );
```

Is it intentional by pouchdb to emit paused when the replication is finished?

rafamel:
After some testing with pouch:
* On every change, pouch will emit for:
** `active`, `change` (pull), and `paused` -in that order- when the connection is active.
** `active` and `paused` -in that order- when the connection is **not** active.
* Will emit an extra pause event when it disconnects.
* Will emit `active` and `pause` occasionally when connection is established and there are no changes.
* Will emit extra `pause` events occasionally even when the connection is established.

From all this, it wouldn't be adequate to derive `active$` from the active/paused sequence. I would propose that we remain faithful to pouchdb events with subscribers. Meaning, mapping $active to `active` events, and adding a additional `pause` and `denied` events [so we cover all event emitters as subscribers](https://pouchdb.com/api.html#sync).

[...]

PR #809, opened 2018-09-17
Replication alive$

rafamel:
Adds alive$

Relates to #763

### DIFF
--- a/docs-src/replication.md
+++ b/docs-src/replication.md
@@ -59,6 +59,13 @@ Emits `true` or `false` depending if the replication is transmitting data. A `fa
 replicationState.active$.subscribe(active => console.dir(active));
 ```
 
+### alive$
+Emits `true` or `false` depending if the replication is alive - data is transmitting properly between databases. A `false` value implies the connection has died -if you're replicating to a remote database, for example. It will only emit `false` if there are pending changes that couldn't be replicated -it won't emit immediately after the connection dies.
+
+```js
+replicationState.active$.subscribe(active => console.dir(active));
+```
+
 ### complete$
 Emits `true` or `false` depending if the replication is completed.
 If you do a `live: true`-sync (default) the replication never completes.
--- a/docs-src/replication.md
+++ b/docs-src/replication.md
@@ -63,7 +63,7 @@ replicationState.active$.subscribe(active => console.dir(active));
 Emits `true` or `false` depending if the replication is alive - data is transmitting properly between databases. A `false` value implies the connection has died -if you're replicating to a remote database, for example. It will only emit `false` if there are pending changes that couldn't be replicated -it won't emit immediately after the connection dies.
 
 ```js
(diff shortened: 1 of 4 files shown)


pubkey:
Looks good.
Can you remove the dist-file `dist/lib/plugins/replication.js` from the PR?

rafamel:
This is done. As a curiosity, why do we have dist files in the git repo? Wouldn't it make sense to include it in the `.gitignore`?

pubkey:
They are commited so people can try the latest version without having to do a build.
https://pubkey.github.io/rxdb/install.html#latest

docs-src/replication.md
```diff
@@ -59,6 +59,13 @@ Emits `true` or `false` depending if the replication is transmitting data. A `fa
 replicationState.active$.subscribe(active => console.dir(active));
 ```
 
+### alive$
+Emits `true` or `false` depending if the replication is alive - data is transmitting properly between databases. A `false` value implies the connection has died -if you're replicating to a remote database, for example. It will only emit `false` if there are pending changes that couldn't be replicated -it won't emit immediately after the connection dies.
+
+```js
+replicationState.active$.subscribe(active => console.dir(active));
```
motin: Typo: `active$` -> `alive$`

rafamel: Thanks @motin ! Solved

motin: @rafamel Great, but you missed the last of the three `active` instances

docs-src/replication.md
```diff
@@ -59,6 +59,13 @@ Emits `true` or `false` depending if the replication is transmitting data. A `fa
 replicationState.active$.subscribe(active => console.dir(active));
 ```
 
+### alive$
+Emits `true` or `false` depending if the replication is alive - data is transmitting properly between databases. A `false` value implies the connection has died -if you're replicating to a remote database, for example. It will only emit `false` if there are pending changes that couldn't be replicated -it won't emit immediately after the connection dies.
+
+```js
+replicationState.alive$.subscribe(alive => console.dir(active));
```
motin: Typo: `console.dir(active)` -> `console.dir(alive)`

docs-src/replication.md
```diff
@@ -59,6 +59,13 @@ Emits `true` or `false` depending if the replication is transmitting data. A `fa
 replicationState.active$.subscribe(active => console.dir(active));
 ```
 
+### alive$
+Emits `true` or `false` depending if the replication is alive - data is transmitting properly between databases. A `false` value implies the connection has died -if you're replicating to a remote database, for example. It will only emit `false` if there are pending changes that couldn't be replicated -it won't emit immediately after the connection dies.
```
motin: Nit: missing spaces after the dashes in "-if" and "-it"

rafamel: I'm a native Spanish speaker; in Spanish dashes go without inner spaces. Is that not the case for English?

PR #814, opened 2018-09-22
Tutorial "Using RxDB with Typescript" incorrect RxCollection type definition.

ashergarland:
## This PR contains:
 - Improved docs

- Tutorial "Using RxDB with Typescript" incorrect `RxCollection` type definition.

## Description
The documentation tutorial page "Using RxDB with Typescript" defines `HeroCollection` incorrectly as `RxCollection<HeroDocType, HeroDocMethods>` when it should be defined as `RxCollection<HeroDocType, HeroDocMethods, HeroCollectionMethods>` to provide typing for the static methods.

### DIFF
--- a/docs-src/tutorials/typescript.md
+++ b/docs-src/tutorials/typescript.md
@@ -59,7 +59,7 @@ type HeroCollectionMethods = {
 }
 
 // and then merge all our types
-type HeroCollection = RxCollection<HeroDocType, HeroDocMethods>;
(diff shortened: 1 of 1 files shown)


pubkey:
Hi @ashergarland thanks for the PR.
The code in the tutorial is unit-tests, so please apply the same changes [here](https://github.com/pubkey/rxdb/blob/master/test/tutorials/src/typescript.ts).
I have not found a plugin to automatically use the test-code inside of the docs.

PR #817, opened 2018-09-22
tests: test Node.js 8 and 10

DanielRuf:


pubkey:
Hi @DanielRuf 
Currently the node-version is pinned [here](https://github.com/pubkey/rxdb/blob/master/package.json#L147) so changing this in travis will not result in the tests running with a different version.

DanielRuf:
At least not 10.6. But why pin it in general? Latest is 10.11.0.

pubkey:
The problem is that there is no greenkeeper-bot for travis.
When a new node-version is released, atm the greenkeeper-bot will create a PR that shows if rxdb works with the latest version. Very often it does not because of some subdependencies, especially levedb. By running the PR throught the CI, we now that there is something wrong with latest node and can fix it.

By hard-setting the node-version inside of travis, we would have to change the version each time a new node is released.

DanielRuf:
If we could use the Travis cache key for storing `node_modules`, this would probably speed up (incremental) builds.

Issue #742, opened 2018-08-05
Adding an easily edited demo for rxdb using code sandbox


barakplasma:
## Case
I want to add a demo editor at CodeSandbox in order to allow one click preview / editing / deployment for the rxdb-react example. 

[![Edit/Demo rxdb-example-react](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/p9xnw75850?module=%2Fsrc%2FDatabase.jsx)

## Describe the problem you have without this
It is slightly more difficult for beginners to interact with the example when they have to 

1. Clone
1. Install
1. Navigate
1. Install again

Compared to simply clicking edit and seeing the live demo with everything working with batteries included.

## Todos 
- [x] Documentation
- [ ] Open PR after discussing this

## Why
To improve adoption of the project

pubkey:
HI @barakplasma 
Thank you for this. We had something like this before but I removed it for the following reason:
When rxdb is updated, we must either manually update the sandbox or set the version in the sandbox to `*` which will likely break things and end up with a broken tutorial.
This makes releasing much more time-prone and I decided that it's not worth it.

Any ideas on how to test this in the CI and auto-update would be great.

barakplasma:
CodeSandbox.io can [import the latest changes from github](https://codesandbox.io/docs/importing#import-from-github) which means it will not need a manual update. This won't work while the react example uses an import [from a few levels up](https://github.com/pubkey/rxdb/blob/ff1caa75369ff746155dc829bf18cc9f659e01a2/examples/react/src/Database.jsx#L1) so I'm not sure how to crack that egg.

I agree with the importance of CI testing the examples.

I could try to adapt the [E2E currently in the Angular](https://github.com/pubkey/rxdb/blob/master/examples/angular2/test/e2e.test.js) and Vue examples to run in the [./test](https://github.com/pubkey/rxdb/tree/master/test) folder based on the angular/vue/react examples. This would let us know if the latest version of rxdb has broken the examples. Alternatively, I could write a new E2E test using [puppeteer](https://github.com/GoogleChrome/puppeteer) to test the React version.

pubkey:
Thanks for that information, auto-update from github means we can make this possible.

So the steps are:
- [] e2e-tests for the react-example
- [] find a way to workarround the uplevel rxdb-import

pubkey:
I'm closing this. If you want to work on this feature, please reopen.

PR #818, opened 2018-09-22
[examples React]: Clean up

skvale:
Basically make it easier to read.

## This PR contains:
Improvements to the React example app

React example just needed some clean up


pubkey:
@skvale Thank you for the great work.
If you have some spare time, we also look for someone adding e2e-tests to the react-example which would then allow us to provide a CodeSandbox. [See here](https://github.com/pubkey/rxdb/issues/742)

PR #822, opened 2018-09-25
Returns promise from RxDB.removeDatabase

will118:
Should we not also `await` or chain L699 somehow?


pubkey:
Yes you are right, merged. Thank you very much.

Issue #681, opened 2018-06-11
Define `_id` as `primary` in the schema


will118:
## Issue
I wish to define `_id` as `primary` in the schema.

We have (legacy) reasons for wanting to use `_id` as the `primary`.

We could instead use `id` and let `RxDB` set `_id`. This passes our sync_gateway validator function but it doesn't work with our Typescript model definitions.

Is there a reason why the following is not possible?

## Code

```js
const RxDB = require("rxdb");
RxDB.plugin(require("pouchdb-adapter-memory"));

async function test() {
  const db = await RxDB.create({
    name: 'testdb',
    adapter: 'memory',
    multiInstance: false
  });

  const mySchema = {
    version: 0,
    disableKeyCompression: true,
    type: 'object',
    properties: {
      _id: {
        type: 'string',
        primary: true
      },
      name: {
        type: 'string'
      }
    }
  };

  const collection = await db.collection({
    name: 'person',
    schema: mySchema
  });

  const doc = {
    _id: "foobar",
    name: "abc"
  };

  await collection.insert(doc);
}
test();
```

will118:
I am happy to work on a PR if there is no reason for the current limitation

pubkey:
Hi @will118 
Sorry for the late answer.
I thought a lot about this, and yes, using `_id` as primary should be possible to make migrations from and to pouchdb easier. We should also make sure that `_id` is not allowed as non-primary field because this would cause many strange bugs.

will118:
@pubkey This is great news.

I got started writing some tests and changing code, but ran into an issue with field names with leading underscores.

I looked through the history and it goes all the way back to the initial commit (at least on GitHub): https://github.com/pubkey/rxdb/blob/db92c863ec3350b08653e22a89a55a1b9bb3c092/lib/RxSchema.js#L218-L219

I understand why there [cannot be a property with a trailing underscore](https://pubkey.github.io/rxdb/population.html), but I am unsure about why no leading underscores.

Is the issue that the getters defined on `RxDocument` could clash with internal properties?

pubkey:
@will118 we disallowed private (=leading underscore) fields, so when some internal methods are introduced, it would never happen that someone uses the method-name as fieldname and then have a problem using rxdb which could only be solved by migrating the data.
With your change, the check should still persist but not match `_id` as exception.

pubkey:
Any update on this?

pubkey:
[...]

PR #824, opened 2018-09-26
Allow `_id` as a primary key  

will118:
I have implemented https://github.com/pubkey/rxdb/issues/681

If there is an edge case or some behaviour not covered by my tests, let me know and I will add the test.

Apologies for the delay. I am committed to getting this merged so feedback is welcome.

#### RxCollection
- [x] Inserting a document with `_id` set on a schema with `_id: primary`
- [x] Inserting a document with `_id` set on a schema with `somethingElse: primary`
- [x] Inserting a document with `_id` set on a schema with no primary defined (implicit `_id`)

#### RxSchema
- [x] Creating a schema with `_id` as primary
- [x] Creating a schema with `_id` **not** as primary


pubkey:
Looks good.
One edge-case: What happens if an primary-field e.g. `passportId` is set, and I insert an object which contains `_id` and  `passportId`? I think we should throw an error if this happens.

will118:
@pubkey Great!

Does this logic not cover it:
https://github.com/pubkey/rxdb/pull/824/files#diff-ee35040c15aa81d65903334acff165c9R371

and this test:
https://github.com/pubkey/rxdb/pull/824/files#diff-162336d85de73bd515f48480759e07cdR375


I just tried writing a test for RxCollection where the schema has `passportId` (`primary: true`) and `_id` but of course this is rejected by `SchemaCheck`!

So I don't think it's possible to write any other tests for this case, let me know if you have any ideas.

pubkey:
Yes you are right. I merged this. Thank you for the great work.

PR #838, opened 2018-10-10
Make RxJsonSchema type optional

bkniffler:
It seems like pouchdb deprecates it, so it should be optional for smooth migrations.

## This PR contains:
 - IMPROVED typings

schema type is deprecated in pouchdb


pubkey:
@bkniffler what do you mean with 'deprecated by pouchdb'? Because pouchdb has no schemas.

pubkey:
Please reopen with more information. I do not understand why this change is needed.

PR #890, opened 2019-02-04
Fix react-native example project

garfieldnate:
The default plugin for validation, `is-my-json-valid`, depends on
`generate-functions`, which uses core Node libraries and is therefore not usable
outside of the Node environment. Attempting to build the project before led
to the following error:

    Unable to resolve module `util` from `<workspace/node_modules/generate_function/index.js`: Module `util` does not exist in the Haste module map

The solution to this is to use AJV for validation instead, which requires using custom 
build logic to determine which plugins are loaded.

Also added try/catch around await statements, since that was the only way I
could find this problem in the first place.

## This PR contains:
bug fix; I have not included a test because the project would not build at all before. 
Just building the project is the test :)

## Todos

Although I only aimed to fix the react-native project in this PR, other non-node 
environments will have the same problem. It would probably be better to provide
a built-in default plugin bundle for non-node environments.

This problem for non-node environments should probably be documented somewhere;
to me the preferable alternative would be to make AJV the default validation and tell
users that they can speed up performance on Node by using `is-my-json-valid` instead.
Right now, RxDB may be broken out-of-the-box for many users.

### DIFF
--- a/examples/react-native/App.js
+++ b/examples/react-native/App.js
@@ -13,7 +13,28 @@ import React from 'react';
 const { width, height } = Dimensions.get('window');
 
 import { default as randomToken } from 'random-token';
-import * as RxDB from '../../';
+
+// We have to do a custom build here because the default validate plugin,
+// is-my-json-valid, only works in the node environment.
+// See https://github.com/pubkey/rxdb/blob/c322b95422e37e9ce20513fdfca784077ad79020/docs-src/custom-build.md
+// for more plugins that you can add.
+import RxDB from 'rxdb/plugins/core';
+import RxDBReplicationModule from 'rxdb/plugins/replication';
+RxDB.plugin(RxDBReplicationModule);
+import RxDBAjvValidateModule from 'rxdb/plugins/ajv-validate';
+RxDB.plugin(RxDBAjvValidateModule);
+
+// required if the database has a password
+import RxDBEncryptionModule from 'rxdb/plugins/encryption';
+RxDB.plugin(RxDBEncryptionModule);
+
+// not required but important for development
+import RxDBSchemaCheckModule from 'rxdb/plugins/schema-check';
+RxDB.plugin(RxDBSchemaCheckModule);
+import RxDBErrorMessagesModule from 'rxdb/plugins/error-messages';
+RxDB.plugin(RxDBErrorMessagesModule);
+
+
 import schema from './src/Schema';
 
 RxDB.plugin(require('pouchdb-adapter-asyncstorage').default);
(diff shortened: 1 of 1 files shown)


garfieldnate:
🤦‍♂️ It works on my computer. Without the change, it does not work. I don't know why.

pubkey:
I'm not sure if this is an rxdb-problem.
The react-native example is in the ci and works, see [here](https://travis-ci.org/pubkey/rxdb/jobs/489481087)
I think we should start by adding something to the CI that shows what is broken.

Does your error appear on build or on runtime?

garfieldnate:
Okay, here's what I see:

* The tests pass. Building as in the instructions and running `npm test` works and passes just fine.
* The app will not run if you actually try to start it using the iOS simulator. Do `npm start`, press `i` to open the app in the iOS simulator. You will immediately get this error:

        Unable to resolve ../../" from ".//App.js`: The module `../../` could not be found"

* There's a discussion of the above issue [here](https://github.com/facebook/metro/issues/7). The solution is to put `"rxdb": "file:../../"` in the dependencies section of package.json. Then in `App.js` you change the import line to `import RxDB from "rxdb";`. I'm not sure how to do fix package.json correctly in this project as it seems that it is written dynamically when the parent project is built. If you could guide me here, I would like to update the PR to fix this.
* If you manually fix the above issue, then the next error is printed in the terminal:
 
        Failed building JavaScript bundle

and in the simulator you see the following:

    Unable to resolve module `util` from `...rxdb/examples/react-native/node_modules/generate-function/index.js`: Module `util` does not exist in the Haste module map or in these directories: `...rxdb/node_modules`

The problem being that `util` is only available in the node environment, and it is required by `is-my-json-valid`.

I believe that the problem is that jest runs tests on Node, completely hiding this problem. I feel like that is an annoying problem with the ecosystem, but I don't know any way to fix it with just jest. The only way I can think of to fix the erroneously-passing test suite is to add an actual end-to-end test using something like Detox or Cavy. The tests need to actually start up a simulator and load the app to show this error.

garfieldnate:
I was going to begin work on a Detox test for this, since that project has documentation on running with Expo and with travis-CI, but I couldn't even build T_T. I filed https://github.com/wix/Detox/issues/1145 and await a fix before I can use the tool.

pubkey:
I could reproduce this problem locally. I am now looking into how to lets this fail the CI so I can ensure the problem will never reapear

pubkey:
I could fix this. Please check out the new release.

garfieldnate:
Works great! Thanks!

Issue #1141, opened 2019-05-21
Allow multiple databases on the same port


josiahruddell:
## Case

Enhancement

## Issue

Cannot connect multiple databases to the same server/app running the same port through rxdb.

## Info
  - Environment: Node.js
  - Adapter: All
  - Stack: Typescript

## Code

```typescript
const db1 = await DB.create({
  name: 'db1',
})

// add some collections to db1

const db2 = await DB.create({
  name: 'db2',
})

// add some collections to db2
```

Create a server on the nodeJS side...

```typescript

db1.server({
  path: `/${db1}`,
  port: 3000,
})


db2.server({ // throws port in use
  path: `/${db2}`,
  port: 3000,
})

```

The feature request is to allow multiple calls to `db.server` with the same port and different paths.

josiahruddell:
I was looking through the code and would be happy to put up a PR but want to get the convo going here first.

pubkey:
Hi @josiahruddell great idea.
I think we should make it possible to pass any express-instance when calling `RxDatabase.server()`. by this people could also run the rxdb-server and the rest-endpoint on the same port.

To serve two rxdb-databases on the same port, we could then do

```js

const server1 = db1.server({
  path: `/${db1}`,
  port: 3000,
})


db2.server({
  path: `/${db2}`,
  express: server1.app
})
```

What do you think?

pubkey:
@josiahruddell have you started on this?
Coincidentally I also need this feature for my project. So I would implement it if you haven't.

josiahruddell:
Yeah I'll open a PR there are some TODOs still

pubkey:
Closed see https://github.com/pubkey/rxdb/pull/1156#issuecomment-517942797

PR #1156, opened 2019-05-28
Extend server plugin to reuse an existing app

josiahruddell:
## This PR contains:
 - #1141: Allow the `db.server` plugin to reuse an existing express app instead of creating a new one every call

 - cannot use the `db.server` on different databases using the same server.

## Todos
- [ ] Tests
- [ ] Documentation
- [x] Typings
- [ ] Changelog

### DIFF
--- a/src/plugins/server.js
+++ b/src/plugins/server.js
@@ -1,11 +1,9 @@
-import express from 'express';
+import expressApp from 'express';
 import ExpressPouchDB from 'express-pouchdb';
 import corsFn from 'cors';
 
 import PouchDB from '../pouch-db';
-import {
-    newRxError
-} from '../rx-error';
+import { newRxError } from '../rx-error';
 
 import Core from '../core';
 import ReplicationPlugin from './replication';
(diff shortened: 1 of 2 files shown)


src/plugins/server.js
```diff
@@ -46,46 +44,69 @@ function tunnelCollectionPath(db, path, app, colName) {
     app.use(path + '/' + colName, function(req, res, next) {
         if (req.baseUrl === path + '/' + colName) {
             const to = normalizeDbName(db) + '-rxdb-0-' + colName;
```
josiahruddell: I'd like to discuss the purpose of this line. It seems like this is here so that if someone uses the default `/db` path, we can still produce unique URLs in the formal `db/{dbname}-rxdb-0-${collectionname}`? If so, should we preserve this behavior? This was giving me problems syncing databases when not using the default path `/db`. Had to recreate the sync URL with the `rxdb-0` part.

josiahruddell: @pubkey ^

pubkey: I am not sure why we have this line. I think it is a workaround to some strange pouchdb-server behavior. If the tests still pass, we can remove it.

pubkey:
What is the state of this?

pubkey:
I'm closing this since there was no response. I still think this would be a good feature, reopen if you have the time.

PR #1157, opened 2019-05-28
ADD z-schema validator plugin

### LINKED ISSUE #235: Content-Security-Policy Error on validate schema

phil-lgr:
## This PR contains:
 - A NEW FEATURE

Both `is-my-json-valid` and `ajv-validate` use `eval()` to perform validation which might not be wanted when `'unsafe-eval'` is not allowed in Content Security Policies. This one is using [z-schema](https://github.com/zaggino/z-schema) as validator which doesn't use `eval`.

## Todos 
- [x] Tests
- [x] Documentation
- [x] Typings
- [x] Changelog

Related to #243, actually closes #235

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,8 @@
     "spark-md5": "^3.0.0",
     "unload": "2.1.0",
     "url": "^0.11.0",
-    "util": "^0.12.0"
+    "util": "^0.12.0",
+    "z-schema": "^4.0.2"
   },
   "devDependencies": {
     "@babel/cli": "7.4.4",
--- a/src/plugins/validate-z-schema.js
+++ b/src/plugins/validate-z-schema.js
@@ -30,7 +30,7 @@ function _getValidator(rxSchema, schemaPath = '') {
     const validatorsOfHash = validatorsCache[hash];
   
     if (!validatorsOfHash[schemaPath]) {
-        const schemaPart = schemaPath === '' ? this.jsonID : this.getSchemaByObjectPath(schemaPath);
+        const schemaPart = schemaPath === '' ? rxSchema.jsonID : rxSchema.getSchemaByObjectPath(schemaPath);
     
         if (!schemaPart) {
             throw newRxError('VD1', {
@@ -56,7 +56,7 @@ function _getValidator(rxSchema, schemaPath = '') {
  * @return {any} obj if validation successful
  */
 const validate = function (obj, schemaPath = '') {
-    const validator = this._getValidator(schemaPath);
+    const validator = _getValidator(this, schemaPath);
     const useValidator = validator(obj);
     /** @type {ZSchema.SchemaErrorDetail[]} */
     const errors = useValidator.getLastErrors();
@@ -78,7 +78,7 @@ const validate = function (obj, schemaPath = '') {
 
 const runAfterSchemaCreated = rxSchema => {
     // pre-generate the validator-z-schema from the schema
-    requestIdleCallbackIfAvailable(() => _getValidator(rxSchema));
+    requestIdleCallbackIfAvailable(() => _getValidator.bind(rxSchema, rxSchema));
 };
 
 export const rxdb = true;
--- a/src/plugins/validate-z-schema.js
+++ b/src/plugins/validate-z-schema.js
@@ -11,8 +11,6 @@ import {
     requestIdleCallbackIfAvailable
 } from '../util';
 
-const validator = new ZSchema();
-
 /**
  * cache the validators by the schema-hash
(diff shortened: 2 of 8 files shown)


phil-lgr:
@pubkey I took the same tests as the ajv-validator plugin, does that makes sense to you?

src/plugins/validate-z-schema.js
```diff
@@ -0,0 +1,104 @@
...
+const validatorsCache = {};
+
+
+/**
+ * returns the parsed validator from z-schema
+ * @param {string} [schemaPath=''] if given, the schema for the sub-path is used
+ * @
+ */
+function _getValidator() {
+    const schemaPath = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : '';
```
pubkey: Is there a reason why you do not use [javascript-defaults](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Functions/Default_parameters) like [here](https://github.com/pubkey/rxdb/blob/master/src/plugins/validate.js#L27)?

pubkey:
@phil-lgr thank you for that great PR.
I added one comment. I will review the rest tomorrow, looks good. :+1:

phil-lgr:
there is a couple of issues:

In my app I get:

```
"RxError: RxError: Sub-schema not found, does the schemaPath exists in your schema? Given parameters: { schemaPath:""} 
```

src/plugins/validate-z-schema.js
```diff
@@ -0,0 +1,103 @@
...
+ * @param {string} [schemaPath=''] if given, the schema for the sub-path is used
+ * @
+ */
+function _getValidator(rxSchema, schemaPath = '') {
+    const hash = rxSchema.hash;
+    if (!validatorsCache[hash]) validatorsCache[hash] = {};
+    const validatorsOfHash = validatorsCache[hash];
+  
+    if (!validatorsOfHash[schemaPath]) {
+        const schemaPart = schemaPath === '' ? rxSchema.jsonID : rxSchema.getSchemaByObjectPath(schemaPath);
```
phil-lgr: I don't understand where does the `getSchemaByObjectPath` comes from? @pubkey mmmm

phil-lgr:
Ok so I'm working on this again now

src/plugins/validate-z-schema.js
```diff
@@ -0,0 +1,103 @@
...
+            schemaPath,
+            obj,
+            schema: this.jsonID
+        });
+    }
+};
+
+const runAfterSchemaCreated = rxSchema => {
+    // pre-generate the validator-z-schema from the schema
+    requestIdleCallbackIfAvailable(() => _getValidator.bind(rxSchema, rxSchema));
```
phil-lgr: I'm not sure why but I had to .bind here... otherwise the test would work, but in the browser it wouldn't.. if you have a better way to handle this let me know

package.json
```diff
@@ -94,7 +94,8 @@
     "spark-md5": "^3.0.0",
     "unload": "2.1.0",
     "url": "^0.11.0",
-    "util": "^0.12.0"
+    "util": "^0.12.0",
+    "z-schema": "^4.0.2"
```
phil-lgr: I noticed that 'ajv' is not listed here, are users supposed to install themselves? if so I guess I should remove z-schema from here

pubkey: I think ajv is missing. Z-schema schould stay there. Im offline since monday btw, i can help then.

pubkey:
What is the state of this?

phil-lgr:
I've been using my fork for sometime, it works fine! 

Not sure what we need to add to the PR, if you can review the code style comments above that would be great

package.json
phil-lgr: what version of ajv should I add?

pubkey:
@phil-lgr thanks. I merged this for now.
I added an entry to the todo-list for the next major version. [See here](https://github.com/pubkey/rxdb/blob/master/orga/before-next-major.md#rewrite-jsonschema-validation)
I think we should rewrite the jsonschema-validation which I will do then.

PR #1225, opened 2019-06-15
Emit collection change after adding it to the db object

milanpro:
## This PR contains:
- A Bugfix

While creating a collection, a subscriber to the db is notified of that, but cannot access the collection via the db object, since it is added after the emitting process.


pubkey:
@milanpro thank you, merged.

PR #1268, opened 2019-06-30
Update to expo SDK 33 (doesn't compile)

garfieldnate:
Update to SDK 33 and set the react version number to that required by expo.
The instructions for upgrading are given here:
https://docs.expo.io/versions/latest/workflow/upgrading-expo-sdk-walkthrough/

Unfortunately, the app will not run because of a cryptic error that I have not been
able to solve:

    React Native version mismatch
    JavaScript version: 0.59.8
    Native version: 0.57.1

The error message also suggests clearing the build cache and rebuilding, but this has not helped me.
This issue is repeatable, and there may be some kind of incompatibility between expo 33 and rxdb.

## This PR contains:
Dependency Update

Expo should be updated to 33 now, since it was a rather large release and they recommend updating
before 34 to make life easier. It would be good to figure out the library incompatibility here for users
who use Expo.

## Todos
- [ ] Fix library incomptability


garfieldnate:
Unfortunately I'm pretty lost when it comes to this incompatibility error. For now I just wanted to open a PR so that we could see the error in Travis.

garfieldnate:
First one failed for the wrong reason so I'm pushing more fixes.

examples/react-native/package.json
```diff
@@ -9,26 +9,28 @@
...
-    "test:bundle": "rimraf -f ./dist && rimraf -f ./expo && EXPO_DEBUG=true expo export --public-url http://example.com/"
+    "test:bundle": "rimraf -f ./dist && rimraf -f ./expo && EXPO_DEBUG=true expo export --public-url http://example.com/",
+    "fix-jest-expo": "curl https://raw.githubusercontent.com/expo/expo/master/packages/jest-expo/src/preset/setup.js > node_modules/jest-expo/src/preset/setup.js",
+    "postinstall": "npm run fix-jest-expo"
   },
   "devDependencies": {
     "babel-preset-expo": "5.2.0",
     "expo-cli": "2.21.1",
-    "jest-expo": "32.0.1",
+    "jest-expo": "^33.0.0",
```
pubkey: Please pin the version (no `^`)

examples/react-native/package.json
```diff
@@ -9,26 +9,28 @@
...
     "react-native-gesture-handler": "1.3.0",
     "react-test-renderer": "16.8.6",
     "rimraf": "2.6.3",
     "schedule": "0.4.0"
   },
   "main": "node_modules/expo/AppEntry.js",
   "dependencies": {
-    "expo": "32.0.6",
+    "expo": "^33.0.0",
```
pubkey: Please pin the version (no ^)

pubkey:
@garfieldnate thank you so much, upgrading expo or anything in react native is pain most of the times :)
I added 2 comments, but the rest looks good for me.

garfieldnate:
No problem! Although I haven't actually resolved the issue that I found. The checks all passed, but in actuality the app will not run.

pubkey:
Hmm. So we should improve the CI I think

garfieldnate:
This is what I see with the iOS emulator:

Screen Shot 2019-07-02 at 6 11 22 PM

garfieldnate:
This last commit fixed the error. I don't know why it only showed up for me with rxdb; it seems like a much more general error than that! Anyway, it's ready to merge once travis gives the go-ahead.

pubkey:
Thank you, I merged this.

PR #1336, opened 2019-07-27
Fixes #1325

### LINKED ISSUE #1325: .populate() returns findOne() on empty string. This results in a random find.
First thanks for this great repo!

## Case
Bug

## Issue
Populate on a ref field that has an empty string returns a random document from that collection.

## Possible solution
Looking at the source, the populate() is the return of a findOne(). findOne() will return the first value when no arguments are supplied. I would suggest to check for falsy on the value of the path argument of the populate function and to return null if false.

## Info
  - Environment: Angular 8
  - Adapter: IndexedDB
  - Stack: 

## Code

Example:
```
transaction.populate('account').then(x => console.log(x.name));

// Schema to populate
{
title: 'sdfsdgsg',
  version: 0,
  description: 'cvbcvbvb',
  type: 'object',

  properties: {
    account: {
      ref: 'accounts',
      type: 'string',
    },
 }

// Schema account
{
title: 'fghfgh',
  version: 0,
  description: 'tryrtytry',
  type: 'object',

  properties: {
    name: {
      type: 'string'      
    }    
  }
}
 
```
Suggested fix:

```
populate(path) {


        const schemaObj = this.collection.schema.getSchemaByObjectPath(path);
        const value = this.get(path);

// fix or something like this
if (!value) { return Promise.resolve(null); }
//---      

        if (!schemaObj) {
            throw newRxError('DOC5', {
                path
            });
        }
        if (!schemaObj.ref) {
            throw newRxError('DOC6', {
                path,
                schemaObj
            });
        }

        const refCollection = this.collection.database.collections[schemaObj.ref];
        if (!refCollection) {
            throw newRxError('DOC7', {
                ref: schemaObj.ref,
                path,
                schemaObj
            });
        }

        if (schemaObj.type === 'array')
            return Promise.all(value.map(id => refCollection.findOne(id).exec()));
        else
            return refCollection.findOne(value).exec();
    },




```

PVermeer:
## This PR contains:
 - A BUGFIX

Issue #1325 

## Todos 
- [ ] Documentation
- [ ] Changelog

### DIFF
--- a/src/rx-document.js
+++ b/src/rx-document.js
@@ -168,6 +168,9 @@ export const basePrototype = {
     populate(path) {
         const schemaObj = this.collection.schema.getSchemaByObjectPath(path);
         const value = this.get(path);
+        if (!value) {
+            return Promise.resolve(null);
+        }
         if (!schemaObj) {
             throw newRxError('DOC5', {
                 path
--- a/test/helper/humans-collection.js
+++ b/test/helper/humans-collection.js
@@ -371,6 +371,7 @@ export async function createMigrationCollection(
 
 
 export async function createRelated(name = util.randomCouchString(10)) {
+    RxDB.PouchDB.plugin(require('pouchdb-adapter-memory'));
     const db = await RxDatabase.create({
         name,
         adapter: 'memory',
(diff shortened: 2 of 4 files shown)


PVermeer:
Typo

PR #1347, opened 2019-08-02
handle missing rev strings

garfieldnate:
Sometimes the `revString` parameter can be undefined when processing document changes.

## This PR contains:
 - A BUGFIX

I am debugging strange behavior on React Native. I am using pouchdb-adapter-asyncstorage and pouchdb-adapter-http to sync from CouchDB. During some of the initial document change reading, some change objects arrive with no revision field. I am not CouchDB expert so I can't tell why this is,
but for now I've patched over the problem by assuming a revision number of `0` for these incoming changes.

## Todos 
- [ ] Discuss: is this correct? Why has it never been seen before?
- [ ] Test

## Discussion

I do not know whether I am doing something weird that causes this or it's a bug. I will attach a backup of the (very small) CouchDB data that I'm working with in case it helps. Please feel free to ask lots of questions or ask for more information; I am all ears on this one.
 
[CouchDB2_backup.zip](https://github.com/pubkey/rxdb/files/3462605/CouchDB2_backup.zip)

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -389,6 +389,9 @@ export function flattenObject(ob) {
  * @return {number}
  */
 export function getHeightOfRevision(revString) {
+    if(!revString) {
+        return 0;
+    }
     const first = revString.split('-')[0];
     return parseInt(first);
 }


pubkey:
I do not think the proposed fix is a good solution. Returning `0` on `undefined` input is a strange behavior. It would be better to throw if the rev is not parseable and the fix the error deeper in the stacktrace.

Does your problem also appear with the memory-adapter?
Have you enabled pouchdb debug and can post the logs?

garfieldnate:
I meant to attach this picture yesterday. It is a screenshot from debugging the issue in VS Code. You can see that the change event data doesn't have a `_rev` attribute:

Screen Shot 2019-08-02 at 8 55 30 PM

The last log entry before this spot is:

    changeEvent: {"op":"INSERT","t":1564856738079,"db":"flashcards2","it":"ogqkcfbd08","isLocal":false,"col":"cards","v":{"headwordUserLang":{"english":"Nice to meet you."},"headwordForeignLang":"ยินดีที่ได้รู้จัก","category":"phrases"}}

This again shows a missing `_rev` attribute. I'll attach the full log here. The relevant error is shown on lines 119 and 120.
[rxdb_error.log](https://github.com/pubkey/rxdb/files/3464186/rxdb_error.log)

Also BTW, I just I noticed there's another issue/PR open in pouchdb regarding missing revisions: https://github.com/pouchdb/pouchdb/pull/7834. Glad to know I'm not the only one dealing with this :)

garfieldnate:
Also, unfortunately I was not able to test with the memory adapter. There seems to be some kind of incompatibility with RN because I get the following:

    The package at "node_modules/sublevel-pouchdb/node_modules/readable-stream/readable.js" 
    attempted to import the Node standard library module "stream". It failed because React Native does not include the Node standard library. Read more at 
    https://docs.expo.io/versions/latest/introduction/faq/#can-i-use-nodejs-packages-with-expo
    Failed building JavaScript bundle.

garfieldnate:
I was able to get the memory adapter to work by installing react-native-stream and then doing `mv node_modules/react-native-stream/ node_modules/stream`. The error did not occur when running in this configuration. I got lots of `document conflict` errors instead 🤦‍♂. PouchDB on ReactNative is such a mess.

You can close this if you think it's not worth pursuing; it may very well be a problem with pouchdb-adapter-asyncstorage. Although RxDB does use this module in its example RN project, so it might still be of interest. For my project, I have to patch this issue or else nothing works.

pubkey:
I'm closing this. For me it looks like this must be solved on the asyncstorage-adapter

PR #1396, opened 2019-08-18
Failing test case for importing dump with attachments

rixo:
## This PR contains:
 - IMPROVED TESTS

Actually, I'm not completely sure what I'm trying to do is supposed to be supported. That's why I'm opening a PR with a test case instead of an issue, so you can judge by yourself.

My problem is that importing a dump of a DB that contains attachments fails with an error.

Furthermore, the attachment's data are not present in the json data produced by the dump.

Is dumping of attachments meant to supported by RxDB?

### DIFF
--- a/test/unit/bug-report.test.js
+++ b/test/unit/bug-report.test.js
@@ -9,100 +9,41 @@
  * - 'npm run test:browsers' so it runs in the browser
  */
 import assert from 'assert';
-import AsyncTestUtil from 'async-test-util';
 
-import RxDB from '../../dist/lib/index';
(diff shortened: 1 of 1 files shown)


rixo:
Thanks again. That's good to me. I've already implemented a basic workaround that save & restore attachments for each doc in the dump. Maybe I'll dig deeper when I'll look into sync / replication.

@pubkey Is there any plan / desire to add support for attachments in dumps?

pubkey:
I added it to the backlog. I have no plans to implement it. If you have the time, a PR would be nice.

Issue #1450, opened 2019-09-03
Cors headers not properly set


dstudzinski:
## Case
Cors headers are not properly set

## Issue
Cors headers are not properly set which causes bugs like:

```
Access to fetch at 'URL1' from origin 'URL2' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: The value of the 'Access-Control-Allow-Origin' header in the response must not be the wildcard '*' when the request's credentials mode is 'include'.
```

But problem is not only with 'Access-Control-Allow-Origin'. Even when this header is passed properly you have:
```
Access to fetch at 'http://127.0.0.1:2000/rxdb/tasks/' from origin 'http://localhost:5000' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: The value of the 'Access-Control-Allow-Credentials' header in the response is '' which must be 'true' when the request's credentials mode is 'include'.
```

## Info
  - Environment: Node.js
  - Adapter: websql
  - Stack: ES6

## Code

```js
database.server({
  path: '/db',
  port,
  cors: true
});
```

PR #1451, opened 2019-09-03
BUGFIX for #1450

dstudzinski:
## This PR contains:
BUGFIX for #1450

Cors headers not properly set #1450


pubkey:
Thanks, merged

PR #1675, opened 2019-11-04
Fix serving migrated collections with server plugin

aredridel:
## This PR contains:

A fix for serving migrated schemas with the server plugin.

When using rxdb, I have a version 1 schema, and it doesn't sync new data from the server since it only finds the v0 database for the collection.

## Todos
- [ ] Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ### X.X.X (comming soon)
 
+Bugfixes:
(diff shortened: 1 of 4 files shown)


pubkey:
Looks good. I restarted the failed CI.
Please also add the change to the [changelog](https://github.com/pubkey/rxdb/blob/master/CHANGELOG.md)

aredridel:
Changelog entry made.

Issue #1655, opened 2019-10-30
Change the way indexes are specified at collection schema definition


marinrusu1997:
## Case
This is an improvement for the way indexed fields are specified at collection schema definition.

## Issue
The actual implementation of RxDB requires indexed fields be marked with a boolean property index. The way these fields are extracted when RxSchema is created implies creating a flatten object from json schema an then doing a lot of filtering and mapping over the keys of this object just to retrieve indexed fields (https://github.com/pubkey/rxdb/blob/master/src/rx-schema.ts#L240). The problem with actual implementation is that it induces huge performance penalties for big schemas with many leves of imbrication ( > 5). In my performance tests, I had a schema which had 9 leves of imbrication and 661 keys of flattened object, it took 52 ms to run _getIndexes_ function. Given the fact that collections are usually created at the application boot, this is a **huge penalty for  application boot time**.
A possible solution is to define indexed fields at the collection schema level, not at the fields level, like _compoundIndexes_. This way, _getIndexes_ could be run in a constant O(1) time. This will imply to add a new check in the _checkSchema_ function to make sure that specified fields actually exists in properties of the schema. 

## Info
  - Environment: Chrome 77.0.3865.120 64-bit on a Windows 10 Enterprise virtual machine running over an Intel Xeon E5-2620 v3 @2.40Ghz, 12GB of RAM. 
  - Adapter: PouchDB adapter using IndexedDB as its data store
  - Stack: Tests were made in a Web Worker

## Screenshots
Bellow is a screenshot from Chrome Profiler:
![image](https://user-images.githubusercontent.com/29524251/67851489-40b40200-fb13-11e9-821b-13138d0e5a55.png)
GC time took 9 ms.
![image](https://user-images.githubusercontent.com/29524251/67851686-9ee0e500-fb13-11e9-8f58-8985b464f2a2.png)

pubkey:
This is a great proposal.
In the last major release we also moved the `require: true` statements to the top-level of the schema.
Doing this also with the indexes makes sense.

marinrusu1997:
Should I implement the proposed improvements myself and do a PR, or you can implement them yourself? Also, when will the new major release come out?

marinrusu1997:
Perfect. I will create a new branch where these improvements will be implemented, and then the community could use it in beta-version.

marinrusu1997:
Could you please give me access rights so I can create remotely the branch which contains changes?

pubkey:
[...]

PR #1695, opened 2019-11-11
Indexes can be specified at top level of schema definition

marinrusu1997:
## This PR contains:
A breaking change: indexed fields have to be specified at the top-level of the collection schema definition, rather than at the field level.
- [x] Tests
- [x] Documentation
- [x] Typings
- [x] Changelog

The problem is described in this [issue](https://github.com/pubkey/rxdb/issues/1655).


.gitignore
```diff
@@ -17,4 +17,5 @@ config/pouch__all_dbs__/
 pouch__all_dbs__/
 pouch__all_dbs__/
 .DS_Store
-.DS_Store?
\ No newline at end of file
+.DS_Store?
+dist/
```
pubkey: RxDB can be installed via github instead of npm. This makes it easier to import builds which are not released as versions to npm.
Therefore please do not add the dist/ folder to the gitignore

docs-src/rx-schema.md
```diff
@@ -109,8 +109,8 @@ const mySchema = {
 
 
 ## Indexes
-RxDB supports secondary indexes which are defined at the schema-level of the collection.
-To add a simple index, add `index: true` inside field options.
+RxDB supports secondary indexes which are defined at the schema-level of the collection.<br>
```
pubkey: I do not think this `<br>` should be there

docs-src/rx-schema.md
```diff
@@ -109,8 +109,8 @@ const mySchema = {
...
+To add a simple indexes, add them in an array to a `indexes`-field at the top-level of the schema-definition.<br>
```
pubkey: the gramar here is wrong

docs-src/rx-schema.md
```diff
@@ -125,16 +125,27 @@ const schemaWithIndexes = {
...
+              type: 'object',
+              properties: {
+                    cvc: {
+                        type: 'number'
+                    }
+              }
+          }       
       }
   },
+  indexes: ['firstName', 'creditCards.[].cvc'], // <- this will create two indexes for these fields
```
pubkey: Are your sure that the `[]`-syntax works with pouchdb's indexes?

src/rx-schema.ts
```diff
@@ -232,41 +232,14 @@ export function getEncryptedPaths(jsonSchema: RxJsonSchema): { [k: string]: Json
  */
 export function hasCrypt(jsonSchema: RxJsonSchema): boolean {
     const paths = getEncryptedPaths(jsonSchema);
-    if (Object.keys(paths).length > 0) return true;
-    else return false;
+    return Object.keys(paths).length > 0;
```
pubkey: :) great catch.

pubkey:
@marinrusu1997 thank your for that awesome PR.
I added some comments.

Two more points:

I think we should sort the indexes at [RxSchema.normalize()](https://github.com/pubkey/rxdb/blob/069d7c35778fc03b56a53ff1e6cce5f58584048e/src/rx-schema.ts#L315). This would make sure that when the sort-order of the indexes is changed, the user would not have to increate the schema-version and run a migration.

Do you think it is a good idea to merge `indexes` and `compoundIndexes` into a single field in the schema? It would then look like
```ts
const schema = {
  indexes: [
    'aNormalIndex',
    [
      'compound',
      'index'
    ]
  ]
};

```

docs-src/rx-schema.md
marinrusu1997: Initially, [] brackets were replaced with **items** in the _getIndexes_ method. But then I found this example on stackoverflow https://stackoverflow.com/questions/52532232/how-to-get-specific-data-from-an-array-using-pouchdb-find, so `[]` remained as they are.

marinrusu1997:
1) Yes, you are right, we need to sort them. 
2) I thought about this. I left them separate, because it's more explicit to have indexes and compound indexes. But if we look closer, they are the same thing, no matter it is normal or compound, so we can merge them.

I can fix it and make a new commit on this PR.

marinrusu1997:
[RxSchema.normalize()](https://github.com/pubkey/rxdb/blob/069d7c35778fc03b56a53ff1e6cce5f58584048e/src/rx-schema.ts#L315) internally calls [sortObject](https://github.com/pubkey/rxdb/blob/069d7c35778fc03b56a53ff1e6cce5f58584048e/src/rx-schema.ts#L316), which already [sorts arrays](https://github.com/pubkey/rxdb/blob/069d7c35778fc03b56a53ff1e6cce5f58584048e/src/util.ts#L197) in the object, so I think there is no need an explicit sort of the indexes. An explicit sort may be needed if we need a different order, such as this one:
`[ 
       'aindex',
      ['aindex', 'bindex'],
      'bindex',
      ['bindex', 'cindex'],
      ['bindex', 'cindex', 'dindex']
  ]
`

pubkey:
Ok I think we have a bug in the old implementation.
Compound indexes are sort-sensitive and therefore should not be resorted by RxDB.

> One thing to note is that the order of these fields matters when creating your index
https://pouchdb.com/guides/mango-queries.html#more-than-one-field

I do not know If you want to also fix that with your PR.

marinrusu1997:
I can fix it.

pubkey:
Hi @marinrusu1997 
I reviewed everything and it looks good.
I have no idea why the CI fails at the moment. I have merge the latest master into `9.X.X` now, maybe that helps. When the CI is green, I can merge this PR.

Thanks again for your awesome work.

marinrusu1997:
I don't know why transpilling fails on CI. On my machine it works (transpilled after running npm clean). Something strange happens on CI. Can you please send me npm logs, I can't see them.

PR #1755, opened 2019-12-03
Fixes #1754

### LINKED ISSUE #1754: Typing error on database.dump()
## Case
Bug

## Issue
Typescript does not know database.dump() is a promise:

## Info
  - Environment: (Node.js/browser/electron/etc..)
  - Adapter: (IndexedDB)
  - Stack: (Typescript, Angular)

## Code

In rx-database.ts:
```
/**
* export to json
*/
dump(_decrypted: boolean = false, _collections?: string[]): string[] | null {
    throw pluginMissing('json-dump');
}
```

In json-dump.ts:
```
function dumpRxDatabase(
    this: RxDatabase,
    decrypted = false,
    collections?: string[]
): Promise<any> {
    const json: any = {
        name: this.name,
        instanceToken: this.token,
        encrypted: false,
        passwordHash: null,
        collections: []
    };

    if (this.password) {
        json.passwordHash = hash(this.password);
        if (decrypted) json.encrypted = false;
        else json.encrypted = true;
    }

    const useCollections = Object.keys(this.collections)
        .filter(colName => !collections || collections.includes(colName))
        .filter(colName => colName.charAt(0) !== '_')
        .map(colName => this.collections[colName]);

    return Promise.all(
        useCollections
            .map(col => col.dump(decrypted))
    ).then(cols => {
        json.collections = cols;
        return json;
    });
}
```
In the generated rx-database.d.ts:
```
    /**
     * export to json
     */
    dump(_decrypted?: boolean, _collections?: string[]): string[] | null;

```

PVermeer:
Issue #1754


PVermeer:
Did some extra work. Types for dumps where not correct. They are now depending on the encryption status via overloads. Also when something from the provided RxDocumentType can be used, it is used now. When encryption is set it will set all properties to any.

Also did a clean up and set checks for unused imports or vars on in tsconfig.json. There where a lot of unused imports.

pubkey:
@PVermeer thanks for your work, your changes look good.

But I think you should not change the linting in this commit. It makes it hard to understand what really has changed because most diffs are reformats. Also for me the linting belongs to a different PR with some discussion. I do not see an advantage on having a max line-length for normal code but not for imports. The max line length is to make it easy to compare files on a left-right-view which is hard when the import-line expands the screen.

PVermeer:
@pubkey I agree and understand.

I'm a bit new at making Pull Requests and I did not realise a PR is based on a branch, and not on a commit. I never had to work with them before.

My IDE (vscode) does the imports automatically (and checks if it's used or not) and before I new it lots of files where changed in this PR. I left it this way so I could see what you had to say about it :D.

I will change my settings for this repo and make a new PR with only the code changes for the fix. I will make another PR that shows al the unused variables / imports with the linter change.

pubkey:
Removing unused imports is not problem you can do that whenever you find them.
The problem comes with reformating existing imports.

PR #1762, opened 2019-12-07
Create a new example with Electron remote

SebastienWae:
## This PR contains:
- NEW EXAMPLE

None.
This is another example of implementing RxDB with Electron but using `remote` instead of `RxDB server plugin`.

### DIFF
--- a/examples/electron-remote/README.md
+++ b/examples/electron-remote/README.md
@@ -1,0 +1,11 @@
+# RxDB Electron Remote example
+
+This is an example usage of RxDB with Electron and `remote` instead of the `RxDB server plugin`. It implements a simple heroes-list which can be filled by the user.
+
+# Try it out
+1. clone the whole [RxDB-repo](https://github.com/pubkey/rxdb)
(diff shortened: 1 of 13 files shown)


examples/electron-remote/README.md
```diff
@@ -0,0 +1,11 @@
+# RxDB Electron Remote example
+
+This is an example usage of RxDB with Electron and `remote` instead of the `RxDB server plugin`. It implements a simple heroes-list which can be filled by the user.
```
pubkey: I think this should link to a description what 'remote' means with electron.
https://electronjs.org/docs/api/remote

pubkey:
Looks good, I added one comment and I restarted the CI

examples/electron-remote/README.md
SebastienWae: Done, I also added links to the other example in each of the Electron README files.

pubkey:
Thank you for that great PR. I merged it

PR #1783, opened 2019-12-13
Fixes #1754

### LINKED ISSUE #1754: Typing error on database.dump()

PVermeer:
#1755  This PR contains:
 - IMPROVED typings
 - A BUGFIX

#1754 

**Improved import export types:**
They are now depending on the encryption status via overloads. Logic:
When _decrypted parameter on x.dump() is set to false or or omitted, it will set all base properties from the RxDocumentType to any (we don't know if it's encrypted or not). When true it will be typed as the RxDocumentType (we definitely know values will not be encrypted).

I think this the best solution without restructuring the code thus making big changes.

**Added types on import-export tests and added negative tests to the database import tests:**
It is difficult in this code structure to let TypeScript infer most types. So I added some types to properties used in tests for type testing purposes.

**Added types to RxJsonSchema's and fixed some issues with them:**
I basically strong typed the test schema's so errors are visible now. Also fixed the some typ errors on them.


src/rx-database.ts
```diff
@@ -330,7 +333,7 @@ export class RxDatabaseBase<Collections = CollectionsOfDatabase> {
     /**
      * delete all data of the collection and its previous versions
      */
-    removeCollection(collectionName: string): Promise<string[]> {
+    removeCollection(collectionName: string): Promise<void> {
```
PVermeer: Changed because of type errors now that this.lockedRun returns the function type.

src/types/rx-collection.ts
```diff
@@ -19,7 +19,7 @@ export interface RxCollectionCreator {
     name: string;
     schema: RxJsonSchema;
     pouchSettings?: PouchSettings;
-    migrationStrategies?: NumberFunctionMap;
+    migrationStrategies?: KeyFunctionMap;
```
PVermeer: This is the type as defined in the class RxCollectionBase that is using this property.

pubkey:
I restarted the CI so it is green now.
I noticed that this PR increases the build-size by 400 byte. See [here](https://travis-ci.org/pubkey/rxdb/jobs/624557679?utm_medium=notification&utm_source=github_status) compared to [here](https://travis-ci.org/pubkey/rxdb/jobs/626207941?utm_medium=notification&utm_source=github_status)
Can you run a build and make a diff on the dist-folder to check what increased that much?

PVermeer:
Sorry don't have time anymore. But you're reply from text to byte is already 215 bytes and gZipped 170 bytes. I added some types + some text to go with it, so seems about right.

pubkey:
I fixed the build size. The problem was the unnecessary usage of async at one point. This made babel using the regenernator runtime.
```typescript
         return _this3._spawnPouchDB(collectionName, v);
       });
     }) // remove documents
-    .then(function (pouches) {
-      return Promise.all(pouches.map(function (pouch) {
-        return _this3.lockedRun(function () {
-          return pouch.destroy();
-        });
+    .then(
+    /*#__PURE__*/
+    function () {
+      var _ref = _asyncToGenerator(
+      /*#__PURE__*/
+      _regeneratorRuntime.mark(function _callee(pouches) {
+        return _regeneratorRuntime.wrap(function _callee$(_context) {
+          while (1) {
+            switch (_context.prev = _context.next) {
+              case 0:
+                _context.next = 2;
+                return Promise.all(pouches.map(function (pouch) {
+                  return _this3.lockedRun(function () {
+                    return pouch.destroy();
+                  });
+                }));
+
+              case 2:
+              case "end":
+                return _context.stop();
+            }
+          }
+        }, _callee);
       }));
-    });
+
+      return function (_x) {
+        return _ref.apply(this, arguments);
+      };
+    }());

```

pubkey:
I merged this now. Thank you @PVermeer

PR #1862, opened 2020-01-19
Added test for reproducing issue with Index inside array.

sofiageo:
## This PR contains:
A test for reproducing a potential bug

Index cannot be created for objects inside arrays


pubkey:
I do not think this works.
But your use case is fixed with https://github.com/pubkey/rxdb/pull/1695/files

PR #1915, opened 2020-02-17
Add collection methods to document type

evanc:
## This PR contains:
 - IMPROVED typings
 - Updated docs

I noticed that the `document.collection` property isn't typed with the collection's static methods. This updates the types so the collection static methods can be added when setting up the document's type.

I've also updated and rearranged the docs for types so it makes more sense- you can't send the collection methods in to the document type if you haven't declared them yet.

I set a default for the collection method types, so this is not a breaking change.

### DIFF
--- a/docs-src/tutorials/typescript.md
+++ b/docs-src/tutorials/typescript.md
@@ -45,13 +45,7 @@ type HeroDocMethods = {
 };
 ```
 
-We can merge these into our HeroDocument.
-
-```typescript
-type HeroDocument = RxDocument<HeroDocType, HeroDocMethods>;
-```
-
-Now we can define type for the collection which contains the documents.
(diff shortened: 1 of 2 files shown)


pubkey:
Hi @evanc sorry for the late answer.
I fixed the CI now, please merge the latest master into this so we can ensure everything works.
Also when the typescript-tutorial is changed, please also apply these changes to the unit-test [here](https://github.com/pubkey/rxdb/blob/master/test/tutorials/src/typescript.ts)

pubkey:
@evanc I had to revert this because it broke the CI when applied to the typescript test.
https://travis-ci.com/github/pubkey/rxdb/jobs/302223975
Please make a new PR where the change is also applied to the typescript test.

PR #1992, opened 2020-03-19
feat: expose pouchApp from spawnServer

Julusian:
## This PR contains:
 - A NEW FEATURE

I am wanting to use rxdb with pouchdb-auth, to provide user authentication, but I am unable to define an admin user at first startup.
This is because pouchdb-auth requires them to be defined as part of its config, which is only exposed on the pouchApp object this exposes. If an admin user is not defined, then every client will be an admin, so not having this control makes securing the master database on first startup very hard.

An example of how the result of this can be used for this is:
```
pouchApp.couchConfig.set('admins', 'admin1', 'password', cb);
```

There was some other fiddling I need to do with raw PouchDB databases to get the client side rxdb setup with pouchdb-auth, but that is fine. (I think just configuring pouchdb-auth with the special _users database)

## Todos 
- [ ] Tests
- [ ] Documentation
- [ ] Typings
- [ ] Changelog

### DIFF
--- a/src/plugins/server.ts
+++ b/src/plugins/server.ts
@@ -118,7 +118,8 @@ export function spawnServer(
         }));
     }
 
-    app.use(collectionsPath, ExpressPouchDB(pseudo));
+    const pouchApp = ExpressPouchDB(pseudo);
+    app.use(collectionsPath, pouchApp);
 
     let server = null;
     if (startServer) {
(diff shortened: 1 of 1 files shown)


pubkey:
@Julusian thanks for the PR.
Please fix the typings and the changelog, then I can merge this.

Julusian:
express-pouchdb doesn't have any typings that I can find, what do you want done here?
For the test is checking that the pouchApp property exists and looks roughly correct enough?

Issue #2048, opened 2020-04-07
GraphQL replication plugin fires exponentially increasing number of requests in case of server errors


gautambt:
## Issue
In scenarios where the server returns a GraphQL error, the number of requests being fired for retrying the request keeps increasing over time. This is happening because the retry is being done from both the runPush method as well as in the run method:

* https://github.com/pubkey/rxdb/blob/1c9832821c306bc66921f2573a57783ca6b9aa45/src/plugins/replication-graphql/index.ts#L168
* https://github.com/pubkey/rxdb/blob/1c9832821c306bc66921f2573a57783ca6b9aa45/src/plugins/replication-graphql/index.ts#L302

On an error, runPush method will schedule a retry and run method will also schedule a retry. When both the retries fire they will schedule two more retries each and so on.

The fix seems to be to remove the retries from runPush & runPull and do the retry once (in case of either method  failing) the _run method. I can raise a pull request if this looks like the correct fix.

## Info
  - Environment: Browser
  - Adapter: IndexDB
  - Stack: Typescript

## Code
The issue can be reproduced by applying the following patch to the heros and trying to insert a hero. In the network console you will see 3 request (1 Options request, 1 POST for setHuman and 1 POST of feedForRxDBReplication). After 10 seconds you will see 6 requests and after 20 seconds you will see 12 requests and so on.

```diff
diff --git a/examples/graphql/client/index.js b/examples/graphql/client/index.js
index c3d95640..624cf44b 100644
--- a/examples/graphql/client/index.js
+++ b/examples/graphql/client/index.js
@@ -157,7 +157,8 @@ async function run() {
          * we can set the liveIntervall to a high value
          */
         liveInterval: 1000 * 60 * 10, // 10 minutes
-        deletedFlag: 'deleted'
+        deletedFlag: 'deleted',
+        retryTime: 10000,
     });
     // show replication-errors in logs
     heroesList.innerHTML = 'Subscribe to errors..';
diff --git a/examples/graphql/server/index.js b/examples/graphql/server/index.js
index e852f1c6..3559eb2b 100644
--- a/examples/graphql/server/index.js
+++ b/examples/graphql/server/index.js
@@ -99,6 +99,7 @@ export async function run() {
             return limited;
         },
         setHuman: args => {
+            throw 'error';
             log('## setHuman()');
             log(args);
             const doc = args.human;
```

pubkey:
[...]

PR #2086, opened 2020-04-15
GraphQL Replication: Refactor retry logic

dome4:
## This PR contains:
 - A BUGFIX

If both push and pull fail `setTimeout(() => this.run(), this.retryTime);` is executed twice. Therefore run() is executed twice again and restarts itself 4x ... This can happen e.g. if there is no internet connection and results in an enormous stack of executing run(). In my opinion it should be sufficient to execute `setTimeout(() => this.run(), this.retryTime);` only once if there is an error in push OR pull.
```js
    async _run() {
        /* ... */

        // retry on error
        if (willRetry) {
            setTimeout(() => this.run(), this.retryTime);
        }
        return willRetry;
    }
```

### DIFF
--- a/src/plugins/replication-graphql/index.ts
+++ b/src/plugins/replication-graphql/index.ts
@@ -175,7 +175,6 @@ export class RxGraphQLReplicationState {
             const ok = await this.runPush();
             if (!ok) {
                 willRetry = true;
-                setTimeout(() => this.run(), this.retryTime);
             }
         }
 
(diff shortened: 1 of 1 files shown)


pubkey:
Can you check https://github.com/pubkey/rxdb/issues/2048 if this is equal?
The `run()` is written in a way that it never should stack up. Also there is a test which ensures that.

pubkey:
I'm closing this. I think it is fixed by https://github.com/pubkey/rxdb/issues/2048

PR #2152, opened 2020-05-05
FIX remove unused pouchdb-all-dbs dependency

### LINKED ISSUE #2098: Support yarnv2 (berry)
### Description
Currently, you can't install rxdb with yarn berry.

### Reproducible steps
1. In a directory, run `yarn init -2` 
2. Run `yarn add rxdb`

trik:
pouchdb-all-dbs package is poorly maintained and depends on an old version of es3ify that can't be installed using Yarn Berry

close #2098

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,7 +111,6 @@
     "jsonschema-key-compression": "1.1.1",
     "modifyjs": "0.3.1",
     "object-path": "0.11.4",
-    "pouchdb-all-dbs": "1.0.2",
     "pouchdb-core": "7.2.1",
     "pouchdb-find": "7.0.0",
     "pouchdb-replication": "7.2.1",
--- a/src/plugins/all-dbs.ts
+++ b/src/plugins/all-dbs.ts
@@ -1,0 +1,214 @@
+import * as PouchDB from 'pouchdb-core';
+import {defer, from, Observable, of as obsOf, Subject, throwError} from 'rxjs';
+import {catchError, concatMap, map, shareReplay, switchMap, withLatestFrom, take} from 'rxjs/operators';
+
+const PREFIX = 'db_';
+const ALL_DBS_NAME = 'pouch__all_dbs__';
+
+/**
+ * PouchDB interface extended by all dbs plugin.
+ */
+interface PouchDBWithAllDbs extends PouchDB.Static {
+    allDbs(callback?: (err: any, dbs: string[]) => any): Promise<string[]>|void;
+    resetAllDbs(callback?: (err: any) => any): Promise<void>|void;
+}
+
+/**
+ * Used to cache all available databases.
+ */
+interface DbsCache {
+    [key: string]: boolean;
+}
+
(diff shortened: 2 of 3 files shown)


pubkey:
Hi @trik 
thanks for the PR.
You have some conflicts in the dist files. I recommend to not commit the dist files but only the code you have changed.
Also the CI is broken because `pouchdb-all-dbs` is needed in the tests.

trik:
@pubkey can you confirm that is needed only for testing purposes? you could move it to dev dependencies but still we'll have the same problem with yarn berry in development.
probably it would be better to write a small testing util to keep a list of all the databases. pouchdb-all-dbs just listens for 'created' and 'destroyed' pouchdb events and stores the names of the databases in a separate db, but probably we can keep them in memory. if you think it could be a good solution, i can work on this

pubkey:
@trik  I cannot confirm, it is used in the [server plugin](https://github.com/pubkey/rxdb/search?q=pouchdb-all-dbs&unscoped_q=pouchdb-all-dbs), not only in the tests. Also it looks like pouchdb-server [is using it](https://github.com/pouchdb/pouchdb-server/search?q=pouchdb-all-dbs&unscoped_q=pouchdb-all-dbs)

Getting rid of that dependency would be great.

Julusian:
Perhaps the solution here is for rxdb to fork and 'maintain' their own version of this dependency (forced via resolutions). I have been having success with this minimal change: https://github.com/Julusian/pouchdb-all-dbs/commit/235fabd386f36135deef343009f9609836f71895

But it be that rxdb needs to mantain a fork of pouchdb-server too, as there are similar issues elsewhere there

trik:
@Julusian i got the point.. i was able to rewrite the alldbs plugin but this doesn't solve the original problem because we still have and indirect dependency to pouchdb-all-dbs through express-pouchdb. pouchdb-server has the same problems and lacks of maintenance, there's plenty of PRs waiting and last commit is from september 2019

pubkey:
I do not want to move the full depdency into RxDB. I have done this in the past with [mquery](https://github.com/pubkey/rxdb/tree/master/src/plugins/query-builder/mquery) and it was a big burden to maintain it. Also we would have a different version of pouchdb-all-dbs then pouchdb-server which might create strange side effects.

Because the rxdb server plugin is only used in nodejs, we could do a conditional require on the first server creation.

trik:
yep, this is clear, I was not proposing to move the dependency in RxDb, just pointing out that my work is totally useless. I’ve been using pouchdb for a very long time and this is not the first time I’m struggling for this kind of problems. it’s a great tool but too much of its ecosystem is poorly or not maintained at all (almost everything in pouch-community, worker-pouch, pouchdb-server, ...)

pubkey:
I am closing this because it will not get merged.
I think we should first add the yarn install to the CI, then work on a fix on this by doing a conditional require at the server plugin.

PR #2171, opened 2020-05-12
change bug-report.test.ts (possible revision tree/document collision …

renierG:
…bug)

## This PR contains:
a possible bug report

## Describe the problem you have with~~out~~ this PR
in-memory database causing documents to conflict and therefore the resulting revision tree property causing a schema validation error on unknown additional properties.
(full error on the bottom of the bug-report.test.ts)


pubkey:
Can you try without `class Item` and just use a plain object.
Everything else looks ok and should work.

renierG:
thats it! wow

PR #2213, opened 2020-05-23
bugfix in prepareQuery for pouchdb

kunal15595:
## This PR contains:
 - A BUGFIX

In `rx-storage-pouchdb.prepareQuery`, a `gt` operator in selector is added if not already present.
This creates a problem if the selector originally has `gte` operator as `pouch-find` internally merges startKey for both these operators, mostly overriding gte with gt.

This PR checks for any comparison operators in selector for that key, and adds that hack only if required, and thereby avoiding reproduced issue.

## Todos 
- [ ] Tests


pubkey:
@kunal15595 wow, great work. I moved the test and added the bugfix to the changelog. merged.

PR #2282, opened 2020-06-15
RxDBReplicationGraphQLPlugin import statement

gotjoshua:
didn't work for me unless i used the * as style import

## This PR contains:
 - IMPROVED DOCS


pubkey:
I fixed this, the plugin has no default import so you must directly import `RxDBReplicationGraphQLPlugin`

Issue #2285, opened 2020-06-16
Feature request: Optional ref properties


DDoerner:
## Case
Feature Request

## Issue
We have a server-side data model with an optional foreign key. This property can either be a string referencing another collection or NULL. It is not possible to define this schema for rxdb.

## Example
Let's say we have two classes, User and Car. Some users have cars, others don't. So the property user.car_id can either point to a car or be null.

A possible schema for User:
```js
  {
    version: 0,
    type: 'object',
    title: 'user',
    required: [ 'id', 'name' ],
    properties: {
      id: {
        type: 'string',
        primary: true
      },
      car_id: {
        type: 'string',
        ref: 'car'
      },
      name: {
        type: 'string'
      }
    }
  }
```

This schema will fail (e.g. during a migration) for any entry in the database with a car_id of NULL, since the schema does not allow for null values. It does not matter whether 'car_id' is a required property or not.

## Possible resolutions
We hope for either:
- ref-Properties allow for a type of ['string', 'null']
- ref-Properties allow for NULL-values if the property is not required

pubkey:
I think this is a bug. The 2nd prosal should be working.

DDoerner:
I've looked into it and the validate plugin throws the error. RxDB itself might allow ref properties with value NULL, but is-my-json-valid does not, since the type does not specify NULL as well. And I assume that the other validate plugins act the same (although I haven't tested it yet), since that is the correct behavior accrording to JSONSchema.

So the only solution I see is the 1st proposal. Or is there anything that can be said against that change?

pubkey:
Closing this, PR is merged. Thank you.

PR #2290, opened 2020-06-17
Nullable relation in RxSchema

DDoerner:
## This PR contains:
 - IMPROVED DOCS
 - A NEW FEATURE

Cannot define a relation property in an RxSchema with a nullable string, as a type of ['string', 'null'] is not accepted. But without a null-type, any validation of objects with a null-reference will fail because they don't fit the schema.

Discussed in issue #2285 .

## Todos
- [ ] Changelog


pubkey:
Great work, merged.

Issue #2367, opened 2020-07-20
Graphql replication async modifier function


zefman:
## Case
REQUEST FOR A NEW FEATURE

## Issue
I would like to be able to use an async function when modifying the document before or after the graphql replication happens. RXDB already provides a modifier function as part of the push and pull config, but I am not able to perform asynchronous code within it.

In my app I have two related models, a Submission and Submission Globals, these are both reflected as collections in RxDB. I want the push replication I have set up for Submissions to also create it's related Submission globals at the same time. I want my modifier function to look something like this:

```
push: {
        queryBuilder: pushQueryBuilder,
        modifier: async (doc: Submission) => {
         const submissionGlobals = await db.submissionGlobals.find()
                .where('submissionId')
                .eq(doc.id)
                .exec()
                .then((items) => items.map(item => item.toJSON()));

          return {
            ...doc,
           submissionGlobals
          };
        },
     },
```

I have created the modifications required but thought its best to discuss here first before making a PR. https://github.com/zefman/rxdb/tree/feat/graphql-replication-async-modifier

Is there any reason this is a bad idea? If you think this is a suitable request I will open a PR. Thanks

## Info
  - Environment: All
  - Adapter: All
  - Stack: All

pubkey:
I do not see any reason why this is a bad idea. The http-request in the end is async anyway so having an async modifier should work too.
Yes please make a PR.

pubkey:
Solved with #2369

PR #2369, opened 2020-07-21
feat: GraphQL replication async modifier function

zefman:
## This PR contains:
GraphQL replication async modifier function

Implementation for - #2367 

Existing tests seem to be adequate to check that this hasn't caused any regressions.

### DIFF
--- a/src/plugins/replication-graphql/helper.ts
+++ b/src/plugins/replication-graphql/helper.ts
@@ -8,7 +8,7 @@ import type {
 export const PLUGIN_IDENT = 'rxdbreplicationgraphql';
 
 // does nothing
-export const DEFAULT_MODIFIER = (d: any) => d;
+export const DEFAULT_MODIFIER = (d: any) => Promise.resolve(d);
 
 /**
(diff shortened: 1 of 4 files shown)


pubkey:
Code looks good. Can you add one test to ensure that it works with async modifiers?
Also we should change the docs and write there somewhere that the modifier can also be async.

src/plugins/replication-graphql/helper.ts
```diff
@@ -8,7 +8,7 @@ import type {
 export const PLUGIN_IDENT = 'rxdbreplicationgraphql';
 
 // does nothing
-export const DEFAULT_MODIFIER = (d: any) => d;
+export const DEFAULT_MODIFIER = (d: any) => Promise.resolve(d);
```
pubkey, line 11: Is there a reason that the default is async?
This will create new Promise-instances on run which costs performance. not much I know, but also we could just keep the previous function which just returns `d`

pubkey:
Nevermind, I added it.

PR #2388, opened 2020-07-29
Use terser for better JS language feature support

stevenroussey-privicy:
## This PR contains:
Switch from UglifyJS to Terser

This will needed if you you plan to include newer versions of Pouch (like 7.2.2). Also, it does a better job and code does not have to be transpiled to simple JS in order to be minimized.


pubkey:
@stevenroussey-privicy great thank you, this saved me some time. Merged.
PouchDB upgrade will follow soon.

PR #2456, opened 2020-08-22
Create compound index with primary key and sort by its items 

dome4:
## This PR contains:
A bugfix concerning compound indexes

I tried to create a _compound index_ (age + id) that contains the primary key (id) of the schema and sort docs by the attributes of the compound index (see test). This fails with the error message "_cannot sort on field(s) 'age' when using the default index_". Why is the default index ('_id)' used here, although an appropriate custom index has been defined?

The first problem was that in _RxStoragePouchDbClass_ - **prepareQuery()** the primary key is changed to '_id' and therefore the new custom index does not match the attribute anymore: 
https://github.com/pubkey/rxdb/blob/8b954b44ce67177003d2ceef2d79c54bc1d27c23/src/rx-storage-pouchdb.ts#L221
Therefore in my opinion the method **createIndex()** of _RxCollection_ has to be adapted and the change of the primary key has to be done as well.

Second, the order of the attributes in **sort()** is changed. This has proven to be a problem with the util **sortObject()**. When the method is called recursively, the param _noArraySort_ is currently not passed to further calls, which results in nested arrays being sorted. The problem occurred in the method **toString()** of _RxQuery_, so the custom index was not found. Additionally, the test `typeof obj === 'object' ` in the method is not sufficient to exclude an array. Therefore I have added `!Array.isArray(obj)` here.

Hopefully I explained the problem sufficiently :)

## Todos
- [ ] Changelog

### DIFF
--- a/src/rx-collection.ts
+++ b/src/rx-collection.ts
@@ -825,10 +825,12 @@ function _prepareCreateIndexes(
             .map(indexAr => {
                 const compressedIdx = indexAr
                     .map(key => {
+                        const primPath = rxCollection.schema.primaryPath;
+                        const useKey = key === primPath ? '_id' : key;
                         if (!rxCollection.schema.doKeyCompression()) {
-                            return key;
+                            return useKey;
                         } else {
-                            const indexKey = rxCollection._keyCompressor.transformKey(key);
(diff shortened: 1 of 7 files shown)


pubkey:
@dome4  great work, merged.

Issue #2471, opened 2020-08-28
localDocument.$.subscribe invoked twice for an atomic insert 


de-robat:
Hello, first of all kudos for this generally well thought of and nicely running software. This Issue is rather to be considered a question before opening a potential PR.

We recently run into a problem when upserting localDocuments. We have a line that reads as follows:

```
document.$.subscribe((doc: MyState) => {
        console.log(doc)
       ...
      })
```
And this one gets invoked twice for a single upsert. I could pinpoint the issue to this part of the code:

https://github.com/pubkey/rxdb/blob/master/src/plugins/local-documents.ts#L195
which reads like this:

```
_saveData(this: any, newData: any) {
        const oldData = this._dataSync$.getValue();
        newData = clone(newData);
        newData._id = LOCAL_PREFIX + this.id;

        const startTime = now();
        return this.parentPouch.put(newData)
            .then((res: any) => {
                const endTime = now();
                newData._rev = res.rev;
>>>         this._dataSync$.next(newData);

                const changeEvent = new RxChangeEvent(
                    'UPDATE',
                    this.id,
                    clone(this._data),
                    isRxDatabase(this.parent) ? this.parent.token : this.parent.database.token,
                    isRxCollection(this.parent) ? this.parent.name : null,
                    true,
                    startTime,
                    endTime,
                    oldData,
                    this
                );
>>>         this.$emit(changeEvent);
            });
    },
```

The two marked lines both invoke the observables listener. I do not understand if this is intended. To add some context, this basically leads for a react application of ours to invoke unnecessary (virtual) rendering cycles which we would like to avoid.

Thanks in advance!

pubkey:
@de-robat Good investigation. I think you have found a bug and we should change that. As you have pointed out, doing `this.$emit(..` will trigger a new invocation downwards.

de-robat:
Ty for the quick reply, i'll prepare a PR and verify the integrity by running the tests.

de-robat:
Ok, unfortunaltey it not just a matter of removing the line. The tests will fail at:

```
1) local-documents.test.js
       .upsertLocal()
         positive
           should update when exists:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:

'bar' !== 'bar2'

      + expected - actual

      -bar
      +bar2
```
So i have to postpone a PR as i have no clue about the architecture around it. :/

pubkey:
[...]

PR #2474, opened 2020-08-28
#2471 multiple invocations of subscription after upsertLocal()

christbald:
## This PR contains:
IMPROVED TESTS revealing a potential bug

Issue #2471: 
- when upsert a localDocument subscriptions invoked twice
- the implemented test additionally revealed that subscribing with
   .$.subscribe on a document it is directly invoked already, 
    not sure whether this is intended

## Todos
- [x] Tests
- [ ] Changelog
- [ ] Bugfix

### DIFF
--- a/test/unit/local-documents.test.ts
+++ b/test/unit/local-documents.test.ts
@@ -114,6 +114,21 @@ config.parallel('local-documents.test.js', () => {
                 assert.strictEqual(doc.get('foo'), 'bar2');
                 c.database.destroy();
             });
+            it('should invoke subscription onces', async () => {
+                let invocations = 0;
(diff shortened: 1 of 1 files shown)


pubkey:
Thanks for adding the test. Can you fix the linting so that the CI runs the tests and does not stop before?

christbald:
> Thanks for adding the test. Can you fix the linting so that the CI runs the tests and does not stop before?

Yes sorry, I did not recognise this by myself. Now it failed exactly at the newly implemented test.

pubkey:
Fixed at 2cf425fbb308fd80b92ce4ae2768bd14607947fd
Thank you for adding the test, this really helped.

PR #2524, opened 2020-09-17
Rework gql replication pull to write docs in single db operation

corinv:
## This PR contains:

An optimization

(as discussed on gitter)

GQL replication was acting slower than expected - on inspection saw there was a `bulkDocs` operation per doc, per batch but this can be reduced to a single `bulkDocs` operation per batch, resulting in faster replication.

Existing tests are passing. Let me know if the PR needs anything else.


pubkey:
Code looks good. I restarted the react-native CI test.

pubkey:
Thank you very much. Merged.

Issue #2084, opened 2020-04-15
sync.error$.subscribe returning string instead object


jonathborg:
I got string [object object] instead an error object on syncGraphQL.

pubkey:
Sorry but this description does not help. Please open a new issue and use the issue template.

PR #2554, opened 2020-10-01
GraphQL replication - error handling

dome4:
## This PR contains:
 - A BUGFIX

GraphQL doesn't work with common HTTP status codes like REST does. Instead it returns a 200 HTTP response with the following body in case of an e.g. unauthenticated error:
```js
 'errors': [
                    {
                        'extensions': {
                            'code': 'UNAUTHENTICATED'
                        },
                        'message': 'user not authenticated'
                    }
                ]
```
References:
- Apollo Docu: https://www.apollographql.com/docs/apollo-server/data/errors/
- Hasura (GraphQL API) handle errros: https://hasura.io/blog/handling-graphql-hasura-errors-with-react/

In this context I stumbled over the fact that an RxError is returned with the message `[Object object]` if the GraphQL plugin sync fails due to the following [line](https://github.com/pubkey/rxdb/blob/196bab59f2f30eb880b336bd66a0b9f737a6ce1d/src/plugins/replication-graphql/index.ts#L215) in `runPush()`. In `runPull()` the erros gets stringified ([see code](https://github.com/pubkey/rxdb/blob/196bab59f2f30eb880b336bd66a0b9f737a6ce1d/src/plugins/replication-graphql/index.ts#L323)) but in my opinion it would be more helpful to have access to the error object to be able to handle it properly.

How the `[Object object]` string results:
```js
var obj = {
  prop1: "value"
}
throw new Error(obj);
```

```
Error: [object Object] 
```

## Todos
- [ ] Changelog


dome4:
might be the same issue like in #2084

dome4:
travis will pass after #2556 was merged

pubkey:
@dome4 code looks good. I merged your other PRs. Can you merge master into this so the CI gets green?

PR #2561, opened 2020-10-05
FIX(angular-example) use correct dependency versions

thefliik:
## This PR contains:
 - A BUGFIX

Without this PR, the Angular example attempts to load the incorrect version of several dependencies and throws an error during compilation. All angular packages need to be for the same version of Angular. Currently, the Angular example mixes libraries intended for Angular 9 and Angular 10 (which is unsupported). I'll add inline comments to explain each of the changes.


examples/angular/package.json
```diff
@@ -5,6 +5,7 @@
     "preinstall": "npm run preinstall:rxdb && npm run preinstall:versions",
     "preinstall:rxdb": "(cd ../../ && npx yarn@1.13.0 pack ../../ --filename ./examples/angular/rxdb-local.tgz)",
     "preinstall:versions": "node ./scripts/ensure-equal-versions.js",
+    "postinstall": "ngcc",
```
thefliik, line 8: As [recommended in the Angular docs](https://angular.io/guide/ivy#ivy-and-universalapp-shell), I've added a `postinstall` step to run the Angular Compatibility Compiler to compile all Angular packages for this version of Angular. If this isn't done in a `postinstall` step, it will be automatically done when you attempt to serve the app. I _think_ if you don't do it in a `postinstall` step it needs to run `ngcc` each time you run `ng serve`, but I'm not positive.

https://angular.io/guide/glossary#ngcc

examples/angular/package.json
```diff
@@ -24,15 +25,15 @@
   "license": "MIT",
   "dependencies": {
     "@angular/animations": "9.1.12",
-    "@angular/cdk": "10.2.3",
+    "@angular/cdk": "9.2.4",
```
thefliik, line 28: All `@angular/` packages need to be for the same version Angular. Previously version 10 and version 9 packages were being mixed together, which is unsupported.

examples/angular/package.json
```diff
@@ -45,7 +46,7 @@
     "zone.js": "0.11.1"
   },
   "devDependencies": {
-    "@angular-devkit/build-angular": "0.1000.8",
+    "@angular-devkit/build-angular": "0.901.12",
```
thefliik, line 49: `@angular-devkit/` packages have unusual versioning policy. `0.10xx.xx` packages are for version 10 of the library, `0.9xx.xx` packages are for version 9 of this library.

examples/angular/tsconfig.app.json
```diff
@@ -4,15 +4,6 @@
...
-  ],
-  "include": [
-    "src/**/*.ts"
-  ],
-  "exclude": [
-    "src/test.ts",
-    "src/**/*.spec.ts"
-  ]
+  "files": ["src/main.ts", "src/polyfills.ts"],
+  "include": ["src/**/*.d.ts"]
```
thefliik, line 8: Angular CLI libraries should only feed `"src/main.ts"` and `"src/polyfills.ts"` to the tsconfig. Previously, a warning was being thrown because `"src/**/*.ts"` was slurping up all the `environment.ts` files (when only one should be used for any given build). I've still included any `.d.ts` files located in the source (though currently I don't think there are any).

thefliik:
I'm going to make another PR to update the Angular example to version 10.

FYI, I started down this road because the `pouchdb-adapter-memory` appears to be incompatibly with `Angular 10`. Attempting to verify that by trying out the Angular example has uncovered these other errors, so I haven't had a chance to explore my initial problem yet.

pubkey:
Yes, renovate bot did this and likely will do it again. But we have the angular example covered in the ci and also e2e tests. So no matter what updates, it should still work. Disabling the updates for angular does not seem correct for me since even non-equal versions will have a working app.

thefliik:
@pubkey 

> So no matter what updates, it should still work

It didn't didn't work in my testing so this appears to be provably incorrect. I'm not sure why your e2e tests didn't catch the problem, but if you try loading the version of this repo before this PR and manually running the Angular example, I found trying to open it in the browser produced an error message.

PR #2562, opened 2020-10-05
FEAT(angular-example) upgrade to Angular 10

thefliik:
## This PR contains:
Upgrades the Angular example to use Angular version 10 (the current latest version of Angular). This PR builds on #2561 to upgrade the Angular example repo to version 10 of the framework. If this PR is accepted, then #2561 can be rejected. PR #2561 contains some additional comments that are applicable to this PR.

No specific problem other than ensuring that rxdb works with the latest version of Angular

## More info
You'll notice that, when building this example app, Angular 10 will print a few new warnings to the console because some dependencies are in Commonjs/AMD format (which Angular doesn't like). I don't think these are new issues, but previously no warning was given for them.


examples/angular/tslint.json
```diff
@@ -36,7 +36,6 @@
     ],
     "no-internal-module": false,
     "no-trailing-whitespace": true,
-    "no-inferrable-types": true,
-    "no-unused-variable": true
-  }
+    "no-inferrable-types": true
```
thefliik, line 39: This file was automatically modified by an Angular-cli migration.

pubkey:
thank you very much, merged.

Issue #2606, opened 2020-10-15
Feature Request: replication-graphql, Allow specifying the data path


angelnar87:
The current code assumes that the data for graphql replication is always the first child of the response. It would be nice if this was actually configurable so that us as developers could specify a different root; in case that the actual data is 2 levels (or 3 or whatever) deep in the response.

Not an issue; just a would-be-nice-to-have 🙂

christofu:
Agreed. We're using postgraphile which kind of wants to return the nodes under the 'nodes' property on the response). To get around it for now we updated src/plugins/replication-graphql/index.ts to match our specific use case in runPull, adding in

```        
// this assumes that there will be always only one property in the response
// is this correct? it is not. Some graphql servers (at least postgraphile) will return a list of
// nodes as an array under the 'nodes' property which therefore needs to be flattened.
// This probably should check if there is any property that is an array and use that, instead
// of just checking for 'nodes'
const initialData = result.data[Object.keys(result.data)[0]];
const data = initalData.nodes !== undefined && Array.isArray(initialData.nodes) ? initialData.nodes: initialData;
```
but I agree that this would be nice as a configuration option!

pubkey:
Yes you are right.
We should make this an optional setting and use the first field if not set.

PR is welcomed.

pubkey:
I am closing this since there is no progress.
PR is still welcomed.

PR #2614, opened 2020-10-18
Allow custom data path for graphql pull replicatoin

angelnar87:
#2606
This PR allows a developer to specify a custom response data path for syncGraphQL

- [x] A new feature

TODO
- [ ] tests
- [ ] docs
- [ ] typings
- [ ] changelog

### DIFF
--- a/src/plugins/replication-graphql/index.ts
+++ b/src/plugins/replication-graphql/index.ts
@@ -14,7 +14,7 @@ import {
     filter
 } from 'rxjs/operators';
 import GraphQLClient from 'graphql-client';
-
+import objectPath from 'object-path';
 
 import {
     promiseWait,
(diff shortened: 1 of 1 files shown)


src/plugins/replication-graphql/index.ts
```diff
@@ -198,6 +199,17 @@ export class RxGraphQLReplicationState {
         return false;
     }
 
+    private getIn(data: any, path: string[] = []): any[] {
```
pubkey, line 202: We should use `object-path` instead. This is already used many times in the project.
https://github.com/pubkey/rxdb/search?q=object-path

pubkey:
Thanks for the PR. I added a comment.

src/plugins/replication-graphql/index.ts
angelnar87: Sounds good to me. I'll make the change

pubkey:
@angelnar87  any update on this?

PR #2627, opened 2020-10-21
FIX skipping docs in push replication

DDoerner:
When updating the rxdb package after the latest release and trying out the feature in our code base, we've noticed that a small bug has remained that . We are yet unsure why this bug wasn't caught by the tests that had been defined for the previous PR.

## Todos
- [ ] Tests

### DIFF
--- a/src/plugins/replication-graphql/index.ts
+++ b/src/plugins/replication-graphql/index.ts
@@ -293,7 +293,7 @@ export class RxGraphQLReplicationState {
             this.syncRevisions
         );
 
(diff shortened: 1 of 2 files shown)


pubkey:
Can you double check the test which was commited with the last PR?
There must be something tested wrong here and I also do not understand how it could pass :)

DDoerner:
I've found the issue. While a skipped document was not properly filtered, it was caught and emitted as an error in the replicationState. I've amended the original test to also listen to replication errors, which now only passes with the latest fix from this PR.

pubkey:
Thx, merged.

PR #2695, opened 2020-11-17
added test for bug with deleted docs showing up in find().sort() with…

bogenpirat:
… indexeddb

## This PR contains:
- Added a test to reproduce a bug described below
- Package dependency for `pouchdb-adapter-indexeddb` in package.json

When applying `sort()` to `find()` queries, documents that were previously deleted are returned.
Checking the browser's IndexedDB reveals that the documents in question are indeed saved with attribute `deleted: 1`, however the incorrectly returned documents have their `deleted` attribute set to `false`.

---
PS: I'm sure I've made some formal errors in this PR, please let me know what to fix.

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,6 +113,7 @@
     "modifyjs": "0.3.1",
     "object-path": "0.11.5",
     "pouchdb-adapter-http": "7.2.2",
+    "pouchdb-adapter-indexeddb": "^7.2.2",
     "pouchdb-all-dbs": "1.0.2",
     "pouchdb-core": "7.2.2",
     "pouchdb-find": "7.2.2",
--- a/test/unit/bug-report.test.ts
+++ b/test/unit/bug-report.test.ts
@@ -9,46 +9,36 @@
  * - 'npm run test:browser' so it runs in the browser
  */
 import assert from 'assert';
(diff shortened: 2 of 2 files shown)


pubkey:
Hi @bogenpirat thank you for the report.
It looks like this is very likely a pouchdb bug. Can you try to reproduce it with plain pouchdb? If it also is broken there, please create an issue at the pouchdb repo.

bogenpirat:
[this test code](https://gist.github.com/bogenpirat/924c7911c06f035a23f680e9deecb533) using only `pouchdb`, `pouchdb-find`, `pouchdb-adapter-indexeddb` and an indexeddb shim for node work fine and does not return deleted documents.

bfritscher:
I also got hit by this error switching to "pouchdb-adapter-idb" works. Is there an issue related to this pullrequest?

pubkey:
I think the indexeddb adapter is broken at many things.
I added a warning about that in the docs: https://github.com/pubkey/rxdb/commit/543427a999e9adaa2ec4c221d08289368cbd10a5
I am closing this. The adapter should be fixed in the pouchdb repository.
Please let me know when you have the same problem with another adapter, then I will repoen.

PR #2725, opened 2020-12-01
Replace randomInt with randomNumber from async-test-util in test files

QuinKinser:
## This PR contains:

This PR replaces use of `randomInt` with`randomNumber` (from pubkey's `async-test-util` library) in test files that can make use of `async-test-util`. Motivation comes from a TODO comment in `test/helper/schema-objects.ts`.

Original TODO comment:
```
// TODO replace these 2 with methods of async-test-util
import randomToken from 'random-token';
import randomInt from 'random-int';
```

There is one other file that uses `random-int`, but for now I'll hold off on replacing it as it can't make use of `async-test-util` (thank you @dome4).

### DIFF
--- a/test/helper/schema-objects.ts
+++ b/test/helper/schema-objects.ts
@@ -4,9 +4,8 @@
 
 import faker from 'faker';
 
-// TODO replace these 2 with methods of async-test-util
+// TODO replace this method with corresponding method from async-test-util
 import randomToken from 'random-token';
-import randomInt from 'random-int';
 
 import {
     randomNumber
@@ -29,7 +28,7 @@ export function human(
         passportId: passportId,
         firstName: faker.name.firstName(),
         lastName: faker.name.lastName(),
-        age: randomInt(10, 50)
+        age: randomNumber(10, 50)
     };
 }
 
@@ -48,7 +47,7 @@ export interface SimpleHumanV3DocumentType {
 export function simpleHumanV3(): SimpleHumanV3DocumentType {
     return {
         passportId: randomToken(12),
-        age: randomInt(10, 50)
+        age: randomNumber(10, 50)
     };
 }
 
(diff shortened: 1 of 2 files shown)


test/helper/schema-objects.ts
```diff
@@ -4,9 +4,8 @@
 
 import faker from 'faker';
 
-// TODO replace these 2 with methods of async-test-util
+// TODO replace this method with corresponding method from async-test-util
```
QuinKinser, line 7: Original comment updated. Will go back and remove `randomToken` in a separate PR

examples/vue/src/components/HeroInsert.vue
```diff
@@ -12,7 +12,7 @@
 
 <script lang="ts">
 import { Component, Prop, Vue } from 'vue-property-decorator';
-import randomInt from 'random-int';
+import { randomNumber } from 'async-test-util';
```
dome4, line 15: hey @QuinKinser this replacement wont work because you're using a dev dependency (async-test-util) in a normal vue component. For the files in the test folder it is possible to reference this dependency but not for production components like HeroInsert. It would be nice to remove random-int as an additional dependency but another solution is necessary to achieve this goal.

One approach (not the cleanest one bc you duplicate code) would be to use the underlying function for the HeroInsert component:

`Math.floor(Math.random() * (max - min + 1)) + min;`

(see [async-test-util](https://github.com/pubkey/async-test-util/blob/94de97e1e43fcdc17db72f51e096729cf1ed5461/src/random-number.js))

QuinKinser: Thank you very much, @dome4. I've scaled back the changes to just replace the function in files that can make use of `async-test-util`. I'll look at removing the dependency entirely in a future pull request.

pubkey:
Thanks, merged.

PR #2745, opened 2020-12-10
graphql - await initial replication resolves on failed replication

dome4:
## This PR contains:
-  IMPROVED TESTS
- A BUGFIX

I experienced that `awaitInitialReplication()` of the graphql replicator plugin resolves even if the replication never succeeds (see attached issue). Is this the expected behavior? 

### Path of issue

`syncGraphQL()` executes `replication.run()` ([see code](https://github.com/pubkey/rxdb/blob/88a99ed9eab7cdac4af6abf003aa307c20da9909/src/plugins/replication-graphql/index.ts#L518)) without param and thus the default `retryOnFail = true` is used. If this first pull cycle fails `run()` is executed with the default param again ([code](https://github.com/pubkey/rxdb/blob/88a99ed9eab7cdac4af6abf003aa307c20da9909/src/plugins/replication-graphql/index.ts#L193)). Works as expected.

The problem is that `syncGraphQL` executes `run(false)` on every following pull sync cycle every `liveInterval` ms ([code](https://github.com/pubkey/rxdb/blob/88a99ed9eab7cdac4af6abf003aa307c20da9909/src/plugins/replication-graphql/index.ts#L528)). If `run(false)`  is executed and fails `this._run(retryOnFail=false)` returns false, the following if check resolves and `initialReplicationComplete` emits true:

```javascript
 if (!willRetry && this._subjects.initialReplicationComplete['_value'] === false) {
     this._subjects.initialReplicationComplete.next(true);
 }
```

([code](https://github.com/pubkey/rxdb/blob/88a99ed9eab7cdac4af6abf003aa307c20da9909/src/plugins/replication-graphql/index.ts#L163))

### Fix
Simply adding `retryOnFail` to the if statement fixes the issue. Thus, only the initial replication cycle and every retry execution of run() are able to emit  `true` to `this.subjects.initialReplicationComplete`.

### Discuss
The current behavior of `awaitInitialReplication()` is that it neither resolves nor rejects if the replication stays erroneous. I'm thinking about if it would be useful to reject the promise and *cancel the whole replication* after an specific timeout that can be set by the user because this timeout hardly depends on the amount of data pulled initially. Could be implemented with `Promise.race()` and an timeout Promise in the `awaitInitialReplication()` method. Looking forward to discus :+1: 

## Todos
- [ ] Changelog

### DIFF
--- a/src/plugins/replication-graphql/index.ts
+++ b/src/plugins/replication-graphql/index.ts
@@ -160,7 +160,7 @@ export class RxGraphQLReplicationState {
             this._subjects.active.next(true);
             const willRetry = await this._run(retryOnFail);
             this._subjects.active.next(false);
-            if (!willRetry && this._subjects.initialReplicationComplete['_value'] === false) {
+            if (retryOnFail && !willRetry && this._subjects.initialReplicationComplete['_value'] === false) {
                 this._subjects.initialReplicationComplete.next(true);
             }
             this._runQueueCount--;
--- a/test/unit/replication-graphql.test.ts
+++ b/test/unit/replication-graphql.test.ts
@@ -1195,6 +1195,48 @@ describe('replication-graphql.test.js', () => {
                 server.close();
                 c.database.destroy();
             });
+            it('should fail because initial replication never resolves', async () => {
+                if (config.isFastMode()) {
+                    // this test takes too long, do not run in fast mode
+                    return;
+                }
+                const liveInterval = 4000;
+                const [c, server] = await Promise.all([
+                    humansCollection.createHumanWithTimestamp(0),
+                    SpawnServer.spawn()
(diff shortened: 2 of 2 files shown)


pubkey:
Yes. Adding a timeout as attribute to `awaitInitialReplication()` makes sense.
The current behavior is intentional. RxDB is for offline first apps, and it can happen that the user opens the app and stays offline for very long. Then suddendly the user is online and `awaitInitialReplication()` resolves after hours.

pubkey:
Ah wait. I misread one part.
So `awaitInitialReplication()` should never resolve when the replication is not completed or in a failed state.

pubkey:
I merged this one. Another PR for the timeout param is welcomed.

PR #2774, opened 2020-12-30
clarify the ref keyword reffer to the properties, not the scheme obje…

RimashMohomed:
…ct name 'refHuman'

## This PR contains:
 - IMPROVED DOCS

`ref`- keyword misguides the reader to name the scheme object with ref, which is not the case

## Todos 
- [ ] Documentation


pubkey:
Yes that makes sense. Thank you.

PR #2805, opened 2021-01-20
Couch repl

pavellzubkov:
## This PR contains:
Was upgraded docker couchdb  to v 3.1.1. Was added some couchdb integration tests with create , delete remote database, sync externally created couchdb docs.

has been problem with externally created docs when rxdb shema compression key flag was true

### DIFF
--- a/config/.mocharc.js
+++ b/config/.mocharc.js
@@ -1,4 +1,5 @@
 'use strict';
+require('isomorphic-fetch')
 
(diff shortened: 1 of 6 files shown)


pubkey:
Hi.
So now I understand the problem.
When you do direct writes to the remote database, you have to write compressed documents.
I added a warning about that in the docs.

PR #2850, opened 2021-02-05
Enhance local document types

openscript:
## This PR contains:
 - IMPROVED typings

The typing of `insertLocal` in `rx-database.d.ts` and `rx-collection.d.ts` was the following:
```
    insertLocal<LocalDocType = any>(id: string, data: any): Promise<...>;
```
I haven't discovered any reason, why data should not be typed with the `LocalDocType` as `upsertLocal()` does it too.

There was also some code redundancy between `rx-database.d.ts` and `rx-collection.d.ts`, which was resolved as well.

A typing test was added accordingly.


pubkey:
@openscript thank you. Merged.

PR #2875, opened 2021-02-13
WIP: add bulkUpdate method

phal0r:
## This PR contains:
- add bulkUpdate as POC to RxCollection

- add bulkUpdate as POC to RxCollection

## Todos 
- [ ] Tests
- [ ] Documentation
- [ ] Typings
- [ ] Changelog

### DIFF
--- a/src/rx-collection.ts
+++ b/src/rx-collection.ts
@@ -543,6 +543,67 @@ export class RxCollectionBase<
         };
     }
 
+    async bulkUpdate<T>(docs: T[]): Promise<{
+        success: RxDocument<RxDocumentType, OrmMethods>[],
+        error: any[]
+    }> {
+        const docsMap: Map<string, T> = new Map();
+        const updatedDocs = docs.map((doc: any) => {
+            docsMap.set(doc.id, doc)
+            return this._handleToPouch(doc)
+        });
+
+        const ids = docs.map((doc: any) => doc.id)
+        const rxDocumentMap = await this.findByIds(ids);
+
+        let startTime: number;
+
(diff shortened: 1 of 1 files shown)


src/rx-collection.ts
```diff
@@ -543,6 +543,67 @@ export class RxCollectionBase<
         };
     }
 
+    async bulkUpdate<T>(docs: T[]): Promise<{
```
phal0r, line 546: @pubkey Currently I select the documents to change outside of this code, use toJSON(true) to transform them to plain objects and apply the modifications. I am not sure, if this is the best way to achieve this. On the other hand, it would be hard to detect changes in general inside this method as we cannot patch documents with pouch (at least I think so).

src/rx-collection.ts
```diff
@@ -543,6 +543,67 @@ export class RxCollectionBase<
...
+        const endTime = now();
+
+        const okResults = results.filter((r: any) => r.ok);
+
+        okResults.forEach((r: any) => {
+            const rxDocument = rxDocumentMap.get(r.id) as RxDocument<RxDocumentType>;
+            const newData = docsMap.get(r.id)! as any
+            newData._rev = r.rev
+
+            const emitEvent = new RxChangeEvent(
```
phal0r, line 579: @pubkey I started to create this as a third-party plugin and I was not able to import ceateUpdateEvent helper properly. Only RxChangeEvent is exported properly.

pubkey:
What is the state of this PR?

pubkey:
I am closing this. The feature is still wanted but I have no time to work on that.
PR with tests is welcomed.

PR #2925, opened 2021-03-02
Fix GraphQL example install docs and bad cp/paste

AntonOfTheWoods:
## This PR contains:
 - better GraphQL example instructions
 - Fix a copy/paste error in the GraphQL example

The GraphQL example will not install with the current instructions, which not follow the pattern of the other examples (which work). The example page shows "Vanilla" when it is for the `graphql` example


pubkey:
Thanks, merged.

PR #2977, opened 2021-03-22
Add notice findByIds doesn't respect input order

AntonOfTheWoods:
## This PR contains:
Notice on `findByIds` behaviour in the documentation

It is unclear from the docs whether `findByIds` will populate the `Map` in the same order as that of the ids passed for filling it. It does not in all cases, though might be expected to by some users, so a warning is warranted.


pubkey:
Thanks, merged.

PR #2984, opened 2021-03-24
Typo in findbyIds$()

justMeTimothyG:
logically I think it should have been an 'an' instead of 'and'

## This PR contains:
 - Improved Docs (Typo)

Just a small typo.


pubkey:
Thanks, merged.

Issue #3022, opened 2021-04-02
Unable to store blob type attachments in Electron


Wuchv:
## Case
bug

## Issue
I've got an error when trying to store an attachment in electron renderer process, with pouchdb adapter.
```typescript
    file: Pick<File, 'name' | 'type'> & { blob: Blob }

    await putAttachment(
              {
                id: file.name,
                data: file.blob,
                type: file.type,
              },
              true
            );
```
Throws:
```typescript
Uncaught (in promise) TypeError [ERR_INVALID_ARG_TYPE]: The first argument must be of type string 
or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received an instance of Blob
```
Versions
```bash
    "pouchdb-adapter-leveldb": "^7.2.2",
    "rxdb": "^9.12.1",
    "electron": "^11.1.1",
```

## Info
  - Environment: electron
  - Adapter: pouchdb
  - Stack: Typescript,React

## Code
After I see the source code, problem in this method, `Buffer.from` can't accept a Blob parameter.
```js
  createBlobBuffer: function createBlobBuffer(data, type) {
    var blobBuffer;

    if (_util__WEBPACK_IMPORTED_MODULE_3__.isElectronRenderer) {
      // if we are inside of electron-renderer, always use the node-buffer
      return Buffer.from(data, {
        type: type
      });
    }
  /* ... */
```

And it is impossible to encrypt the blob type attachments, `JSON.stringify` also can't accept a Blob parameter.
```js
var _encryptValue = function _encryptValue(value) {
  return encrypt(JSON.stringify(value), this.password);
};
```

pubkey:
Can you extend [this test](https://github.com/pubkey/rxdb/blob/master/examples/electron/test/render.test.js) to reproduce the error?

Wuchv:
[electron test](https://github.com/Wuchv/rxdb/blob/master/examples/electron/test/render.test.js)
I can't download the file rxdb-local.tgz , may be because of the network reasons, so I didn't run the code, but I think this will reproduce the error.

kuka:
Can confirm that this is true. What was the reason to use `node-buffer` when `isElectron()`?

kuka:
I don’t think there’s any downside in removing L55-L60: https://github.com/pubkey/rxdb/blob/86967dc951c3e3e4e0a36b352dde5f272ad384b5/src/plugins/attachments.ts#L55. Correct me if I’m wrong.

[...]

PR #3098, opened 2021-04-28
adds a blob attachment test for idb, #3022;

kuka:
## This PR contains:
- A test for blob attachments with IDB adapter with password (https://github.com/pubkey/rxdb/issues/3022)

- Needs to make sure blob attachments don’t cause any problems

## Todos
- [ ] Make sure test makes sense


pubkey:
I am closing this because it is likely fixed in the newest version. If not, please reopen.

PR #3115, opened 2021-05-04
fixes electron spectron test spec, #3022;

kuka:
## This PR contains:
- A minor fix for Electron Spectron test that seemingly failed due to no delay

- Electron test must be fixed before other fixes can be merged

## Todos
None


pubkey:
Thank you, merged.

pubkey:
I activated the electron tests, but they still seem to randomly fail https://github.com/pubkey/rxdb/runs/2513495325

kuka:
Thanks! Yeah, seems like it’s stuck for some reason. Was it the same for you locally? Is there a way to get artifacts out of the Github actions instance here? Not visible publicly for me. @pubkey

PR #3163, opened 2021-05-18
add null to getLocal return type

chrisdrackett:
## This PR contains:

 - IMPROVED typings

getLocal is currently set to only return a document, but should also be able to return null


pubkey:
The ci fails because we have to also change the tests.

PR #3195, opened 2021-05-28
#2606: Allow 'dataPath' to be specified on the GraphQL replication pull object to specify where to fetch result documents from

joshmcarthur:
## This PR contains:

This PR contains a new feature to allow `dataPath` to be specified as a property of the `GraphQLSyncPullOptions` to allow nested or alternative GraphQL data responses to be used. 

Without this feature, it's not possible to support any GraphQL response other than one that returns the raw document data within the first key on the `data` property of the response JSON.

An example of such a response is a paginated collection type, where the pagination info is returned to the client:

```
type HumanCollection {
  collection: [Human!]
  pageInfo: PageInfo!
}

type PageInfo {
  totalCount: Int
}

type QueryType {
  humanCollection(
      limit: Int = 10
      offset: Int = 0,
      minUpdatedAt: Int!
    ): HumanCollection!
}
```
In this case, the data we want will be found at `data.humanCollection.collection`, not `data.humanCollection`.

I've been using this patch in a project for a few weeks, and it's working nicely. The original implementation is in https://github.com/pubkey/rxdb/pull/2614 - I changed where the `dataPath` property is looked up to be on the `pull` property (to avoid the double-namespacing), and added types, tests, etc.

## Todos 
- [x] Tests (only basic tests though - I couldn't think of a way of testing this without extending the existing test schema with new types and duplicating some resolver functions)
- [x] Documentation
- [x] Typings
- [x] Changelog


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 Features:
   - Added `RxDatabase.migrationStates()` which returns an observable to observe the state of all ongoing migrations.
+  - Added `dataPath` property to GraphQL replication pull options to allow the document JSON lookup path to configured instead of assuming the document data is always the first child of the response [#2606](https://github.com/pubkey/rxdb/issues/2606)
```
pubkey, line 7: You should add this to the top "comming soon" section, not an already released version.

docs-src/replication-graphql.md
```diff
@@ -161,6 +161,7 @@ const replicationState = myCollection.syncGraphQL({
     pull: {
         queryBuilder: pullQueryBuilder, // the queryBuilder from above
         modifier: doc => doc // (optional) modifies all pulled documents before they are handeled by RxDB. Returning null will skip the document.
+        dataPath: null // (optional) specifies the object path to access the document(s). Otherwise, the first result of the response data is used.
```
pubkey, line 164: If it is optional, the default value must be `undefined` not `null`

CHANGELOG.md
joshmcarthur: @pubkey this is in the "X.X.X (coming soon)" section at the moment? Do you mean you'd like it moved elsewhere? Should I add a new 'coming soon' section?

joshmcarthur: Ah yep, my `master` was out of date. Just merged origin/master and resolved the conflict to place this in the correct version.

pubkey:
We should add at least one test to prevent accidentially breaking this on refactors. Atm if the whole functionallity would be removed, the CI would still be green.

joshmcarthur:
Happy to - I'll need to pull some stuff around in the GraphQL test server as I'll need to reuse the human collection resolver function.

pubkey:
Yes this seems necessary, thank you very much.

joshmcarthur:
Just a quick note to say this is still on my todo list, I expect to get to it in 3 days.

joshmcarthur:
@pubkey I added a new query field to the GraphQL test schema that I'm using in a new test that checks the expected document is fetched. Let me know if you have any other feedback!

pubkey:
Thank you josh, I merged it.

PR #3215, opened 2021-06-06
Enable `noUncheckedIndexedAccess`

### LINKED ISSUE #3204: Enable `noUncheckedIndexedAccess`
## Case

Enable  [the `noUncheckedIndexedAccess` compiler option](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) to prevent `undefined` value when accessing properties via index.

## Issue

By enabling this flag, projects depending on this library could also apply the flag without [the `skipLibCheck` option](https://www.typescriptlang.org/tsconfig#skipLibCheck) which degrades type checking.

I could provide a pull request if you think it would be helpful. However, since this wonderful project is relatively complex, it would be hard for me to get it correct. I will still do my best.

Also, to enable the flag, the `jsonschema-key-compression` library with the https://github.com/pubkey/jsonschema-key-compression/pull/173 fix must be released first to pass the compiler check.

seanwu1105:
## This PR contains:

 - IMPROVED typings

Fix #3204.

### DIFF
--- a/src/change-event-buffer.ts
+++ b/src/change-event-buffer.ts
@@ -48,6 +48,9 @@ export class ChangeEventBuffer {
      */
     getArrayIndexByPointer(pointer: number): number | null {
         const oldestEvent = this.buffer[0];
+
+        if (!oldestEvent) throw new Error(`oldestEvent is undefined.`);
+
         const oldestCounter = this.eventCounterMap.get(
             oldestEvent
         ) as number;
@@ -117,7 +120,7 @@ export class ChangeEventBuffer {
         while (t > 0) {
             t--;
             const cE = this.buffer[t];
-            if (cE.documentData && cE.documentData._rev === revision) return true;
+            if (cE && cE.documentData && cE.documentData._rev === revision) return true;
         }
         return false;
     }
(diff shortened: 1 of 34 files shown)


pubkey:
@seanwu1105 the CI fails, can you check?
I think having the optional operator like `myobject.subobject?.foobar` is not allowed in the current compiling with babel.

seanwu1105:
@pubkey

I will make some adjustment to pass the tests and linter in CI.

I am not sure why the TypeScript compiler used in this project does not support optional chaining, but I will replace it will a nullable condition check.

The bang in `meta.collectionStates[collectionName]!.lastSequence` means [non-null assertion](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator). I use the assertion in some places I think we could safely ignore the compilation error. If you think we should not use the assertion at all, still, I can add non-nullable condition check to ensure it is type-safe and passes compilation.

seanwu1105:
@pubkey I have updated the PR. To enable the flag, one breaking change is necessary. See this commit https://github.com/pubkey/rxdb/pull/3215/commits/de41c635d8170662346629dc5949f68c0ad5d1bc for details.

seanwu1105:
@pubkey I have fixed the typing for unit tests.

pubkey:
https://github.com/pubkey/rxdb/issues/3204#issuecomment-858921925

PR #3227, opened 2021-06-15
Update documentation on importing

### LINKED ISSUE #3216: Request for further documentation of imports
## Case

I am in the unfortunate situation that I regularly need to wipe some collections and insert them again.

I have resolved to use imports for faster insert of documents in my database

E.g.
```javascript
const constructedDump = {
  docs: tableDocuments,
  encrypted: false,
  name: "myCollection",
  passwordHash: null,
  schemaHash: "1d...ac",
}
await database.myCollection.importDump(constructedDump)
```

## Issue

In this regard, I find the documentation lacking, and I have the following questions I would like to have answered:

- Does import yield events?
  - If it does: Can I disable it during import or bulk insert to gain performance? (maybe in the future?)
- Is there any speed to gain by constructing a dump for the whole database when I need to overwrite the whole thing anyway?
- If I import with `schemaHash` set to a previous version of the collection schema, will it then automatically migrate the imported data?

## Info
  - Environment: browser/Capacitor
  - Adapter: IndexedDB
  - Stack: Vue, Typescript

beruic:
## This PR contains:
 - IMPROVED DOCS

The intention of this PR is to fix #3216 , and as such it should not be merged until all the questions of the issue are resolved, and it is agreed that the added documentation covers the clarified information.

### DIFF
--- a/docs-src/rx-collection.md
+++ b/docs-src/rx-collection.md
@@ -249,6 +249,7 @@ To import the json dump into your collection, use this function.
 myCollection.importDump(json)
   .then(() => console.log('done'));
 ```
+Note that importing will fire events for each inserted document.
 
 ### sync()
 This method allows you to replicate data between other RxCollections, pouchdb instances or remote servers which support the couchdb-sync-protocol.


pubkey:
Thank you, merged.

PR #3231, opened 2021-06-17
Update middleware to be more flexible about input data

chrisdrackett:
## This PR contains:
 - IMPROVED typings
 - Suggested Features
 
I want to pass in an object that has luxon dates for my date fields and then convert them into strings within the `preInsert` hook. This is currently difficult because this hook expects the input to be the RxDocumentType.

### DIFF
--- a/src/rx-collection.ts
+++ b/src/rx-collection.ts
@@ -120,7 +120,8 @@ export class RxCollectionBase<
     InstanceCreationOptions,
     RxDocumentType = { [prop: string]: any },
     OrmMethods = {},
-    StaticMethods = { [key: string]: any }
+    StaticMethods = { [key: string]: any },
+    MiddlewareInputType = RxDocumentType
     > {
 
     constructor(
@@ -348,7 +349,7 @@ export class RxCollectionBase<
      * to not have duplicated code.
      */
     async insert(
-        json: RxDocumentType | RxDocument | any
+        json: MiddlewareInputType | RxDocument
     ): Promise<RxDocument<RxDocumentType, OrmMethods>> {
         // inserting a temporary-document
         let tempDoc: RxDocument | null = null;
@@ -386,7 +387,7 @@ export class RxCollectionBase<
     }
 
     async bulkInsert(
-        docsData: RxDocumentType[]
+        docsData: MiddlewareInputType[]
     ): Promise<{
         success: RxDocument<RxDocumentType, OrmMethods>[],
         error: RxStorageBulkWriteError<RxDocumentType>[]
@@ -508,7 +509,7 @@ export class RxCollectionBase<
     /**
      * same as insert but overwrites existing document with same primary
      */
-    upsert(json: Partial<RxDocumentType>): Promise<RxDocument<RxDocumentType, OrmMethods>> {
+    upsert(json: Partial<MiddlewareInputType>): Promise<RxDocument<RxDocumentType, OrmMethods>> {
         const useJson = fillObjectDataBeforeInsert(this as any, json);
         const primary = useJson[this.schema.primaryPath];
         if (!primary) {
@@ -535,7 +536,7 @@ export class RxCollectionBase<
     /**
      * upserts to a RxDocument, uses atomicUpdate if document already exists
      */
-    atomicUpsert(json: Partial<RxDocumentType>): Promise<RxDocument<RxDocumentType, OrmMethods>> {
+    atomicUpsert(json: Partial<MiddlewareInputType>): Promise<RxDocument<RxDocumentType, OrmMethods>> {
         const useJson = fillObjectDataBeforeInsert(this as any, json);
         const primary = (useJson as any)[this.schema.primaryPath];
         if (!primary) {
--- a/src/rx-collection.ts
+++ b/src/rx-collection.ts
@@ -348,7 +348,7 @@ export class RxCollectionBase<
      * to not have duplicated code.
      */
     async insert(
-        json: RxDocumentType | RxDocument
+        json: RxDocumentType | RxDocument | any
     ): Promise<RxDocument<RxDocumentType, OrmMethods>> {
         // inserting a temporary-document
         let tempDoc: RxDocument | null = null;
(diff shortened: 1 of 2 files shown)


chrisdrackett:
I updated the types on preInsert to attempt and type `this`. However in use the type does not come though. I'm not 100% sure why yet. Here is some example code:

```
collection.preInsert(function (document: RawDocumentType) {
      console.log('collection', this.name)
      mutateNewDocument(document)
    }, true)
```

I was hoping I could get thing so that `this` is typed as `RxCollection<RxDocumentType, OrmMethods>` but it is still coming through as any.

src/rx-collection.ts
```diff
@@ -349,7 +349,7 @@ export class RxCollectionBase<
     }
 
     insert(
-        json: RxDocumentType | RxDocument
+        json: RxDocumentType | RxDocument | any
```
pubkey, line 352: I really do not want to set the input of `insert()` to any. In my opinion this destroys the value of typescript here.

chrisdrackett: maybe there is a way to specify the shape of the input via a generic that defaults to `RxDocumentType | RxDocument`?

chrisdrackett:
@pubkey I updated this to support sending in a type used for input. I'm open to suggestions here as well!

src/rx-collection.ts
chrisdrackett: @pubkey I updated this type to come from either the user supplied input type, or, if that isn't supplied, what it is now.

chrisdrackett:
@pubkey I think this might be getting close. The biggest issue that I'm not sure how to handle is that because the middleware currently expects an object to be mutated in place you can't really confirm in typescript that the resulting document will pass validation.

If you have an input doc that has a `createdAt` with a type of `DateTime` and RxDB is expecting a `string` I'm not sure how to reconcile that in the current design.

one way around it would be to have the middleware functions expect a return of the type `RxDocumentType` so it can check and make sure what you are returning is the correct type, but this feels like a pretty big API change.

chrisdrackett:
I guess the suggestion would be to move the middleware to work like the following:

existing:
```
myCollection.preInsert(function(plainData){
    plainData.age = 50;
}, false);
```

proposed:
```
myCollection.preInsert(function(inputData){
    return { age: 50, …inputData }
}, false);
```

the type expected by the return could be `RxDocumentType` so we can support different types for `inputData` (of type `MiddlewareInputType`) and output (`RxDocumentType`).

src/types/rx-collection.d.ts
```diff
@@ -61,30 +63,33 @@ export type RxCollectionHookCallbackNonAsync<RxDocumentType, OrmMethods> = (
...
 
     // HOOKS
-    preInsert(fun: RxCollectionHookNoInstanceCallback<RxDocumentType, OrmMethods>, parallel: boolean): void;
-    preSave(fun: RxCollectionHookCallback<RxDocumentType, OrmMethods>, parallel: boolean): void;
-    preRemove(fun: RxCollectionHookCallback<RxDocumentType, OrmMethods>, parallel: boolean): void;
-    postInsert(fun: RxCollectionHookCallback<RxDocumentType, OrmMethods>, parallel: boolean): void;
-    postSave(fun: RxCollectionHookCallback<RxDocumentType, OrmMethods>, parallel: boolean): void;
-    postRemove(fun: RxCollectionHookCallback<RxDocumentType, OrmMethods>, parallel: boolean): void;
+    preInsert(fun: RxCollectionHookNoInstanceCallback<RxDocumentType, OrmMethods, MiddlewareInputType>, parallel: boolean): void;
+    preSave(fun: RxCollectionHookCallback<RxDocumentType, OrmMethods, MiddlewareInputType>, parallel: boolean): void;
```
pubkey, line 88: Are you sure we can assume the input type `MiddlewareInputType` for all hooks, not only the preInsert one?

chrisdrackett: good point, I don't have any expereince with hooks other than the insert or save ones. I think I was looking at the docs and saw `plainData` and figured this meant there was a way to provide data from outside RxDB, but thinking more seems like this probably isn't the case.

chrisdrackett: now I'm curious what `plainData` is for in the case of the `Remove` hooks

chrisdrackett: updated

chrisdrackett:
@pubkey I'm curious if I can get your thought on this comment: https://github.com/pubkey/rxdb/pull/3231#issuecomment-883507463

I know it's a big change, but I'm not sure how to do anything else in typescript without running into one issue or another. At the moment we're just ignoring typescript completely for all our middleware essentially, which I'd rather not do if at all possible.

pubkey:
I think the main problem is that the middleware hooks feature was copied from the mongoose project. And there it was implemented way before people used typescript. So it is not really possible to make these hooks compatible with typescript and fully typed.
Users could always add any pre/post hooks that expect and return a different datatype.
So maybe it would be easier to just add the `any` type to the hook inputs.

chrisdrackett:
@pubkey one thing I wonder is if the hooks could be updated to work both ways.

Modify in Place:

```
myCollection.preSave(function(plainData, rxDocument){
    // modify anyField before saving
    plainData.anyField = 'anyValue';
}, false);
```

Return Modified Doc:

```
myCollection.preSave(function(newData, rxDocument){
   updatedData = plainData.toJSON()
   updatedData.anyField = 'anyValue'

   return updatedData
}, false);
```

I *think* this way you could add types to the return type and make sure it matches the collection schema. I believe as a typescript user using the `Return Modified Doc` version you could type both the input (`newData`) and the return. I'm not quite sure how or if this would work under the hood, and its obviously more complex, but it would give a way to type these functions.

pubkey:
Yes you are right. I haven't had this in my mind, but the current implementation of the middleware hooks are based on side effects. We should return the new data instead of mutating the input. 
But this would be a breaking change.

pubkey:
Closing this. I think we should first do #3426 so it becomes easier to have the correct typings.
Also a PR with adding `any` would be ok for me.

PR #3245, opened 2021-06-24
add missing possible `null` for `getLocal` on database

chrisdrackett:
## This PR contains:
 - IMPROVED typings

getLocal looks like it always returns a doc, but can return `null` if they key does not exist.

### DIFF
--- a/test/typings.test.ts
+++ b/test/typings.test.ts
@@ -666,7 +666,12 @@ describe('typings.test.js', function () {
                 const myDb: RxDatabase = {} as any;
                 const typedLocalDoc = await myDb.getLocal<{foo: string;}>('foobar');
                 const typedLocalDocInsert = await myDb.insertLocal<{foo: string;}>('foobar', { bar: 'foo' });
-                const x: null = typedLocalDoc.foo;
+
+                if (!typedLocalDoc) {
+                    throw new Error('local doc missing');
+                }
+
+                const x: string = typedLocalDoc.foo;
                 const x2: string = typedLocalDocInsert.foo;
             });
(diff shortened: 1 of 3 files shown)


pubkey:
I think we need a null check for `typedLocalDoc` in the typings tests to make the CI green.

test/typings.test.ts
```diff
@@ -666,7 +666,7 @@ describe('typings.test.js', function () {
                 const myDb: RxDatabase = {} as any;
                 const typedLocalDoc = await myDb.getLocal<{foo: string;}>('foobar');
                 const typedLocalDocInsert = await myDb.insertLocal<{foo: string;}>('foobar', { bar: 'foo' });
-                const x: string = typedLocalDoc.foo;
+                const x: null = typedLocalDoc.foo;
```
pubkey, line 669: This one fails.
You have to add a null check to resolve the `LocalDoc | undefined` type.

```ts
if(!typedLocalDoc){
  throw new Error('local doc missing');
}
```

chrisdrackett: ok, not sure why I can't get it to fail locally 🤔

pubkey: Did you run 'npm run build'?

chrisdrackett: I'll give that a try next time!

pubkey:
I restarted the CI because it looked like a random crash.

chrisdrackett:
yeah, I didn't see a way to re-run it myself! thanks!

pubkey:
Thank you, merged.

PR #3260, opened 2021-06-29
export all types

chrisdrackett:
## This PR contains:

 - IMPROVED typings

This PR would remove the need for #3259 by just exporting all types. This does not use the `type` prefix, but I'm pretty sure that is ok as I don't think there is any JS in the types folder so this shouldn't increase the bundle size as far as I can tell.


src/core.ts
```diff
@@ -84,71 +84,4 @@ export type {
...
-    RxQueryOptions,
-    RxReplicationState,
-    RxTypeError,
-    ServerOptions,
-    SyncOptions,
-    SyncOptionsGraphQL,
-    MigrationStrategies,
-    WithAttachmentsData
-} from './types';
+export * from './types';
```
pubkey, line 87: `export` is different from `export type`. A non-type-only export will include many empty files in the resulting webpack build because babel cannot strip them. An `export type` removes the export on build time and reduces the build size.

chrisdrackett: good to know, I was just figuring if there wasn't any code in the files it wouldn't add to the bundle size, but didn't actually test

PR #3287, opened 2021-07-21
Svelte example project

bkeating:
This PR contains a base-bones Svelte project that demonstrates how to use RxDB by way of a simple note-taking application.


pubkey:
Thank you.
We need to add the example to the CI to ensure it works.
In the other examples we run some e2e tests via testcafe, maybe you can add this to the svelte example also. https://github.com/pubkey/rxdb/blob/master/examples/angular/test/e2e.test.js

bkeating:
Absolutely. Just thought I’d kick off a PR to pick up some momentum. Thank you for your quick reply and I’ll get working on the CI and e2e tests in-place.

DominikGuzei:
I would suggest refactoring this example app to https://kit.svelte.dev/ as this would reduce the setup boilerplate code tremendously and will probably become the recommended way to write real-world Svelte apps.

bkeating:
Thank you for looking into this. A SvelteKit example would be great as well, but I made this for raw svelte on purpose. I’m making a few electron apps that utilize rxdb and that are written in svelte so this was the byproduct that i promised myself I’d share out. (:

bkeating:
Had a little time to look at this today. I don't have a test completed yet, but it looks straightforward. Svelte is a compiler, so if all these tests are going off element ids/classes, we should have no problem as it's all there and accessible. Eager to see what the result is. Im hoping the reactivity of it all shines through.

bkeating:
ok, @pubkey I just committed a basic e2e test. Now im itching to go test more things. Super cool framework. 

Is there anything else I need to cover?

cgsmith (approved):

pubkey:
I tested this locally and merged it. I will add it to the CI. Thank you very much for your work.

PR #3376, opened 2021-09-08
remove batch from push replication example

chrisdrackett:
as this option isn't available it is confusing to have it in the documentation.

## This PR contains:

 - IMPROVED DOCS

### DIFF
--- a/docs-src/replication-graphql.md
+++ b/docs-src/replication-graphql.md
@@ -200,7 +200,6 @@ const replicationState = myCollection.syncGraphQL({
     url: 'http://example.com/graphql', // url to the GraphQL endpoint
     push: {
         queryBuilder: pushQueryBuilder, // the queryBuilder from above
-        batchSize: 5, // (optional) amount of documents that will be send in one batch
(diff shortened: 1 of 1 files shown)


pubkey:
I do not fully understand.
The batch size option of the push part is [used](https://github.com/pubkey/rxdb/blob/master/src/plugins/replication-graphql/index.ts#L298) and [typed here](https://github.com/pubkey/rxdb/blob/master/src/types/plugins/replication-graphql.d.ts#L23).
So setting this option should be possible.

chrisdrackett:
maybe I'm confused, as far as I can tell the push endpoints are setup to send a single document at a time. In this case what does `batch` do? or is there a way to setup an endpoint that sends multiple docs to the server at once?

pubkey:
Ah now I see.
So the batchSize is used for how many changed documents get pulled out of the database before sending them (one at a time) to the server.
So we should not remove the batchSize from the docs but clearify that it does not actually affect how docs are send to the server.

pubkey:
I changed the docs in 788d4b1d5b21820a6d6ffab24f5ded9f8f9b8434

PR #3385, opened 2021-09-13
added failing test for collection name-reuse

jfrancos:
## This PR contains:
A new, failing test

When I remove a collection and then recreate it (a new collection with the same name), operations (e.g. `insert`) on that collection result in unhandled rejection errors: `Uncaught (in promise) Error: primary missing !!`

Operations don't directly `throw`, so I added an `unhandledrejection` `eventListener` in order to fail the test.

Other (more difficult to directly "test") issues coincide with this error, and I'm hoping that fixing this issue will result in my other issues resolving.  For example, before the recreation of a collection, a `find` observation on the collection is appropriately run/updated when an initial-replication completes, but this doesn't happen after the collection is recreated. (I'm not asking for help with this other issue at this point -- just giving additional info)

### DIFF
--- a/test/unit/bug-report.test.ts
+++ b/test/unit/bug-report.test.ts
@@ -8,7 +8,7 @@
  * - 'npm run test:node' so it runs in nodejs
  * - 'npm run test:browser' so it runs in the browser
  */
-import assert from 'assert';
+import assert, { fail } from 'assert';
 import AsyncTestUtil from 'async-test-util';
 
 import {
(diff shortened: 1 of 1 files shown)


pubkey:
Hi @jfrancos 
your test only runs in the browser because it is using the window variable. This breaks the ci atm without fully running your test.
Can you ensure that is only runs in the browser and not in node. [I added some example code on how to do that here](https://github.com/pubkey/rxdb/commit/bcb38dd493e9aaf71ecc518745a0bb052cfcb77a)

pubkey:
Can you check if this works with plain pouchdb?
I am pretty sure it is a bug of the idb adapter.

PR #3392, opened 2021-09-16
fix spelling of recieved => received

chrisdrackett:
## This PR contains:

 - SOMETHING ELSE

Fixing misspelled `recieved` -> `received`

Because this touches code I think its a breaking change.

## Todos 
- [ ] Tests
- [ ] Documentation
- [ ] Typings
- [X] Changelog


pubkey:
Thank you. Yes this would be a breaking change.
We could add both getters `recieved$` and `received$` and add a deprecation warning to the wrong one. Then we can remove the deprecated getter in the next major version.

chrisdrackett:
@pubkey updated!

PR #3395, opened 2021-09-16
Replication doc key

### LINKED ISSUE #3394: Feature Request: Need a way to reset graphql replication
## Case
Feature Request

## Issue

I'd like to add a `reset` function on `RxGraphQLReplicationState`. We need this because we currently have a way for a user to delete data on our server (but not in their local app) Right now we have the following bug:

1. User chooses to delete server data (we turn off replication and delete data on server)
2. Later User chooses to store data on server again
3. !! Users data does not go to server because RxDB thinks all data has already been replicated.

This feature would have a way to remove the local documents that track replication so that replication will start again from scratch.

chrisdrackett:
fixes #3394

## This PR contains:

 - A NEW FEATURE

As part of updating to 10.0 we noticed that we're getting events for the replication status when watching collection document updates. This is new and was breaking some things in our app. For now we're filtering this out by constructing the ID ourselves, but I thought others might have this same issue and it would be nice for `RxGraphQLReplicationState` to provide these id's to make it easer.

The other thing this PR does is provide a `reset` function on `RxGraphQLReplicationState`. We need this because we currently have a way for a user to delete data on our server (but not in their local app) Right now we have the following bug:

1. User chooses to delete server data 
2. User chooses to store data on server again
3. User's data does not go to server because RxDB thinks all data has already been replicated.

I figured it might be useful for others to be able to do this same thing.

## Todos 
- [X] Tests
- [ ] Documentation
- [ ] Typings
- [ ] Changelog

### DIFF
--- a/docs-src/replication-graphql.md
+++ b/docs-src/replication-graphql.md
@@ -279,9 +279,13 @@ When the clients sends a document to the server which causes a conflict, this ha
 
 When you call `myCollection.syncGraphQL()` it returns a `RxGraphQLReplicationState` which can be used to subscribe to events, for debugging or other functions.
 
-#### .replicationDocId
+#### .pullDocId
 
-This is the id of the document used to track the replication state. This is a local document added to the collection. You can use this id to filter out this doc when subscribing to `update$` on the collection if needed.
+This is the id of the document used to track the pull replication state. This is the id of a local document added to the collection. You can use this id to filter out this doc when subscribing to `update$` on the collection if needed.
+
+#### .pushDocId
+
+This is the id of the document used to track the push replication state. This is the id of a local document added to the collection. You can use this id to filter out this doc when subscribing to `update$` on the collection if needed.
 
 #### .isStopped()
 
@@ -326,6 +330,10 @@ Cancels the replication. This is done automatically if the `RxCollection` or it'
 ```js
 await replicationState.cancel();
 ```
+
+#### .reset()
+
+Resets the replication state. This will make your replication start from scratch pulling and pushing all docs again.
 
 #### .recieved$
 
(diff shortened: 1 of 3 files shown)


pubkey:
Changes look good. We need at least one test for the reset function.

special-character:
@pubkey I am working with @chrisdrackett on re-replicating our database and ran into a case that keeps some documents from replicating (even after removing and re-creating the `localDocumentStore` for the collections). 

In [crawling-checkpoint.ts](https://github.com/pubkey/rxdb/blob/master/src/plugins/replication-graphql/crawling-checkpoint.ts#L169) our document we want to re-replicate is filtered out because the `_rev` ends with the replication endpoint. It looks like this is to keep documents from replicating too many times.

We have been thinking of some solutions but don't know the full uses of the `_rev` field. One idea we had was to do some kind of bulk write to change the `_rev` property to either not include anything after the `-`, or mutate it in some other way. Do you have any inputs/ideas that might help? For `_rev`, what is the format for (it seems like the schema version and something else, maybe sequence)?

Thanks for any input!

@chrisdrackett Feel free to add anything I missed

pubkey:
Yes RxDB uses the _rev field to track if the write came out of a graphql replication. Normally the rev field is used in couchdb do detect conflicts. But even if RxDB would use some different field, you still would have the same problem of resetting that state after you reset the replication.
When I implemented the GraphQL replication, I did not plan for that case.
Maybe we could change the behavior to not only include the endpoint hash but also another id which describes the replication instance. Then when you reset the replication, there would be a new replication instance and rxdb could know again what was already replicated and what was not.

pubkey:
Closing this. If you still working on that please ping me so I can reopen it.

jcs224:
I'm wondering if there's a way to do this in CouchDB replication + Dexie local driver? Even if this feature isn't added, is there guidance on which data to remove from Dexie to make it so I can replicate fresh from local device to CouchDB? Thanks!

Issue #3396, opened 2021-09-16
Some local documents don't seem to get deleted on remove


special-character:
## Case
`removeRxDatabase` seems to not clear some local documents. The steps I took to get here:

1) Have an RxDB database with replication turned on (and some document replicated at least once)
2) Call `removeRxDatabase` and start from scratch
3) Call `syncGraphql` to start syncing with a fresh database

In this case, the `pullReplicationQueryBuilder` gets the document that was replicated in step one, skipping any documents before it (since it uses that document's `updatedAt`).

## Issue
The local doc from the first sync should not show up in the replication for the second one.

Example: 
<img width="807" alt="CleanShot 2021-09-16 at 15 03 21@2x" src="https://user-images.githubusercontent.com/5407471/133691729-248d2c4a-a8df-480c-9ffa-3e665c843ce3.png">

The doc returned above was replicated the first time this database replicated before calling `removeRxDatabase`.

## Info
- RxDB Version: 10.0.3
- Environment: React native
- Adapter: getRxStoragePouch('react-native-sqlite')
- Stack: React, TS

I am looking into why this happens.

pubkey:
Likely equal to #3319

pubkey:
@special-character I think this is solved with the latest release.
Please test and maybe close this issue.

special-character:
Thanks, will do!

special-character:
This appears to be fixed now, thanks.

Issue #3319, opened 2021-08-08
database.remove() doesn't remove IndexedDB local documents


ofekd:
## Case
Bug

## Issue
GraphQL replication tried pulling new documents even after I ran `db.remove()`. Using the storage inspector in Firefox I could see the pull checkpoints were still there. I also suspect other local documents remain, but didn't have time to test.

In most browsers I can use `IndexedDB.databases()` to enumerate and delete them all manually. Does anyone know of a workaround for Firefox?

## Info
  - RxDB Version: 10.0.2
  - Environment: Browser
  - Adapter: IndexedDB
  - Stack: Typescript, Babel, Vue

pubkey:
Thank you for reporting. Fixed in the attached commit.

ofekd:
Thank you for fixing, much appreciated

ofekd:
I have installed 10.0.3 and the pull checkpoint still persists after `remove()`.

Is there a way to get the list of the pouch IndexedDB databases? I would like to erase everything, regardless of `remove()`.

pubkey:
I am not sure how to get all indexeddb databases.
Can you make a PR with a test to reproduce the problem?
In the new release we remove the whole local document store, so I do not see any way how the checkpoint can still persist.

ofekd:
I see in the storage explorer that RxDB created IndexedDB databases for _both_ version 0 and version 1 of my schema (currently set to version 1 in the schema file). Is that intentional? I think it has something to do with it because the graphql pull checkpoint is set on the version 0 database. This was a fresh DB so it's not something that carried from old data.

pubkey:
RxDB since version 10 creates a separate IndexedDB for local documents and 'normal' documents.
The local documents indexeddb always has version 0.

ofekd:
I've added a failing test

![2021-08-10-124330](https://user-images.githubusercontent.com/6827960/128845484-ca0b3943-052b-4a8a-b252-21a4b6766d28.png)

PR #3397, opened 2021-09-17
Remove local docs when remove is called on the database

special-character:
## This PR contains:
 - A BUGFIX

fixes: https://github.com/pubkey/rxdb/issues/3396

## Todos
- [x] Tests
- [x] Fixing the bug

### DIFF
--- a/src/rx-database.ts
+++ b/src/rx-database.ts
@@ -705,23 +705,31 @@ export async function removeRxDatabase(
         {}
     );
 
+
     const docs = await getAllDocuments(storageInstance.internalStore);
+    
     await Promise.all(
         docs
-            .map(colDoc => colDoc.collectionName)
-            .map(async (id: string) => {
+            .map(async (colDoc) => {
+                const id = colDoc.collectionName
                 const split = id.split('-');
-                const name = split[0];
+                const collectionName = split[0];
                 const version = parseInt(split[1], 10);
+
                 const instance = await storage.createStorageInstance<InternalStoreDocumentData>(
                     {
                         databaseName,
-                        collectionName: name,
+                        collectionName,
                         schema: getPseudoSchemaForVersion(version, 'collectionName'),
                         options: {}
                     }
                 );
-                return instance.remove();
+                const localInstance = await storage.createKeyObjectStorageInstance(databaseName, collectionName + '-local', {})
+
+                await instance.remove()
+                await localInstance.remove()
+
+                return 
             })
     );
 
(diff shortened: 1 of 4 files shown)


src/rx-database.ts
```diff
@@ -705,23 +705,31 @@ export async function removeRxDatabase(
...
+                            databaseName,
+                            collectionName,
+                            schema: getPseudoSchemaForVersion(version, 'collectionName'),
+                            options: {}
+                        }
+                    ),
+                    storage.createKeyObjectStorageInstance(databaseName, collectionName + '-local', {})
+                ]);
+
+                await Promise.all([instance.remove(), localInstance.remove()]);
```
special-character, line 731: The "fix" was to make sure the local instance was removed. If we take out `localInstance.remove()`, the test fails.

test/helper/humans-collection.ts
```diff
@@ -55,6 +55,40 @@ export async function create(
...
+    name: string = 'human',
+    multiInstance: boolean = true,
+    eventReduce: boolean = true
+): Promise<RxCollection<HumanDocumentType, {}, {}>> {
+    if (!name) {
+        name = 'human';
+    }
+    PouchDB.plugin(require('pouchdb-adapter-memory'));
+    const db = await createRxDatabase<{ human: RxCollection<schemaObjects.HumanDocumentType> }>({
+        name: 'humandatabase',
```
special-character, line 69: This is duplicated code and shouldn't really be this way. This was an easy way to quickly create the database without a different random name so that I could prove the local docs aren't removed.

pubkey:
Hi @special-character 
I fixed #3319 with [this commit](https://github.com/pubkey/rxdb/commit/544f466cf7b107a1f09fa49b0d41899dd5637cb1).
Can you check if that also solves your problem?

special-character:
Thanks @pubkey, it looks like it should but I can test it out. Is this deployed in `10.0.3`? I was certain when I upgraded I got the latest version I saw on npm so I am confused about how I didn't get this.

pubkey:
I started the [build job](https://github.com/pubkey/rxdb/actions/runs/1254723983) when this is finished the dist folder should be up to date.

special-character:
@pubkey I got a chance to pull the dist folder into my project and try it out. As far as I can tell, it appears to work

pubkey:
Thank you for testing. I will release a new version tomorrow.
By the way you can also use the latest non-released rxdb version by using the commit hash in your package.json. [see here](https://rxdb.info/install.html#latest)

special-character:
Going to close this because it was already done in the upstream repo.

Issue #3417, opened 2021-09-27
Migration from RxDB 9 does not work


swnf:
## Case
Bug, `.addCollections()` throws one of the following error messages (depends on platform):

```
TypeError: primaryKey is undefined
TypeError: Cannot read property 'key' of undefined
```

## Issue
I'm trying to migrate a collection created with RxDB 9. Its (old) schema does not contain the `primaryKey` field. RxDB 10 crashes when I call `addCollections()` with the new schema (which has a `primaryKey`). My migration strategy is not called.

## Info
  - RxDB Version: 10.0.3
  - Environment: Browser (Firefox), reproduced in NodeJS
  - Adapter: IndexedDB
  - Stack: Angular/Typescript

## Code (NodeJS)

Install dependencies:
```bash
npm install old-rxdb@npm:rxdb@9 rxdb rxjs pouchdb-adapter-leveldb leveldown
```

```js
import { createRxDatabase, addRxPlugin } from "old-rxdb";
import PouchdbAdapterLeveldb from "pouchdb-adapter-leveldb";

(async () => {
  addRxPlugin(PouchdbAdapterLeveldb);

  const mySchema = {
    version: 0,
    type: "object",
    properties: {},
  };

  const db = await createRxDatabase({
    name: "test",
    adapter: "leveldb",
  });

  const collections = await db.addCollections({
    mycollection: {
      schema: mySchema,
    },
  });

  process.exit(0);
})().catch((err) => {
  console.error(err);
  process.exit(1);
});
```
```js
import { createRxDatabase, addPouchPlugin, getRxStoragePouch } from "rxdb";
import PouchdbAdapterLeveldb from "pouchdb-adapter-leveldb";

(async () => {
  addPouchPlugin(PouchdbAdapterLeveldb);

  const mySchema = {
    version: 1,
    type: "object",
    primaryKey: "id",
    properties: { id: { type: "string" } },
    required: ["id"],
  };

  const db = await createRxDatabase({
    name: "test",
    storage: getRxStoragePouch("leveldb"),
  });

  const collections = await db.addCollections({
    mycollection: {
      schema: mySchema,
      migrationStrategies: {
        1: (oldDoc) => {
          const { _id, ...data } = oldDoc;
          return { ...data, id: _id };
        },
      },
    },
  });

  process.exit(0);
})().catch((err) => {
  console.error(err);
  process.exit(1);
});
```

pubkey:
[...]

PR #3418, opened 2021-09-28
Add unit test for #3417

swnf:
## This PR contains:
Unit test

#3417


swnf:
Updated the migration strategy to do nothing because id is not actually stored in the document.

PR #3423, opened 2021-09-29
fix: add missing property from rxdb

miroswd:

I copied and pasted the example, but was returned this error: "getRxStoragePouch is not defined".

## Todos 
- [ ] Tests
- [ ] Documentation
- [ ] Typings
- [ ] Changelog


pubkey:
Equal to #3424
But I already merged #3424 even given that you where first, sorry for that.
I now merged both of them.

PR #3431, opened 2021-10-04
Fix "downsides" typo

FND:
I noticed the `<title>` of https://rxdb.info/downsides-of-offline-first.html erroneously spells "downsites".

    $ git grep -l ownsites | xargs sed -i s/ownsites/ownsides/

However, I'm not sure whether this is the correct way to fix all these files; please double-check.


pubkey:
Thank you, merged.

PR #3446, opened 2021-10-11
use deleted flag for pushing in graphql replication

### LINKED ISSUE #3436: deletedFlag not getting set on push during graphQL replication
## Case
Bug

## Info
  - RxDB Version: 10

## Code

using a config like in the docs:

```
const replicationState = myCollection.syncGraphQL({
    url: 'http://example.com/graphql', // url to the GraphQL endpoint
    push: {
        queryBuilder: pushQueryBuilder, // the queryBuilder from above
        batchSize: 5, // (optional) amount of documents that will pulled out of the storage at once. This does not affect how many documents are send to the server in a single request.
        modifier: d => d // (optional) modifies all pushed documents before they are send to the GraphQL endpoint. Returning null will skip the document.
    },
    deletedFlag: 'deleted', // the flag which indicates if a pulled document is deleted
    live: true // if this is true, rxdb will watch for ongoing changes and sync them
});
```

We're finding that we get an error like the following after updating to RxDB 10:

```
[{"extensions": {"code": "BAD_USER_INPUT", "exception": [Object]}, "locations": [[Object]], "message": "Variable \"$calendar\" got invalid value { color: \"#D796F8\", createdAt: 1629398262001, id: \"4600CBB8-C57C-4A00-81F2-CBA8D89AA8F0\", title: \"My Calendar\", type: \"NORMAL\", updatedAt: 1633630617642 }; Field \"isDeleted\" of required type \"Boolean!\" was not provided."}]
```

I'm looking at https://github.com/pubkey/rxdb/blob/master/src/plugins/replication-graphql/index.ts#L345 and wondering if this should be the deleted flag? seems like the deleted flag used to be used: https://github.com/pubkey/rxdb/blob/c4be6acf422db4d33d2aa394779f0368312fd81e/src/plugins/replication-graphql/index.ts#L294

but I haven't had time to dig in and see what is actually going on yet.

chrisdrackett:
## This PR contains:
 - A BUGFIX

Fixes #3436

I don't feel like this is a full fix yet as I'm not sure how if a doc is deleted in pouch that gets set via the deleted flag. I think this has something to do with checking for if this is a delete operation, but I'm not sure why that was changed via https://github.com/pubkey/rxdb/commit/480c56bfa4d18d304525e1cd2b2e20b34a5364ce#diff-9d7498b685187264496834aaae5e2d04b1273c86b921c6b1bc1a2bf5c797f6d8 and I don't want to rebreak things without understanding the fix. 

## Todos 
- [X] Tests
- [ ] Documentation
- [ ] Typings
- [X] Changelog

### DIFF
--- a/src/plugins/replication-graphql/index.ts
+++ b/src/plugins/replication-graphql/index.ts
@@ -345,7 +345,7 @@ export class RxGraphQLReplicationState<RxDocType> {
 
                 // TODO _deleted should be required on type RxDocumentData
                 // so we do not need this check here
-                if (!changeWithDoc.doc.hasOwnProperty(this.deletedFlag)) {
+                if (!changeWithDoc.doc.hasOwnProperty('_deleted')) {
                     changeWithDoc.doc[this.deletedFlag] = false;
                 }
 
--- a/src/plugins/replication-graphql/index.ts
+++ b/src/plugins/replication-graphql/index.ts
@@ -345,8 +345,8 @@ export class RxGraphQLReplicationState<RxDocType> {
 
                 // TODO _deleted should be required on type RxDocumentData
                 // so we do not need this check here
-                if (!changeWithDoc.doc.hasOwnProperty('_deleted')) {
-                    changeWithDoc.doc._deleted = false;
+                if (!changeWithDoc.doc.hasOwnProperty(this.deletedFlag)) {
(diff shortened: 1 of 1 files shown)


chrisdrackett:
I think this is actually fine now that I look at it more. I'm pretty sure we were just using it wrong. I was looking for our flag in our query builder when I should have been looking for `_deleted`.

PR #3454, opened 2021-10-18
Fix usage of Buffer for browsers

swnf:
## This PR contains:
Improved tests + bugfix

`Buffer` is not available in Browsers but `util.ts` tries to access it. The browser tests cannot detect this because they use browserify which polyfills `Buffer`. I've adapted the karma config to detect this issue and patched `util.ts`.

## Todos
- [X] Tests


pubkey:
Thank you. Merged.

Issue #3495, opened 2021-11-10
Migrate away from `graphql-client` and `is-my-json-valid`


jesec:
## Case

Something else

## Issue

npm emits:

<details>
  <summary>expand</summary>
  
  ```
# npm audit report

jsonpointer  <5.0.0
Severity: moderate
Prototype Pollution in node-jsonpointer - https://github.com/advisories/GHSA-282f-qqgm-c34q
fix available via `npm audit fix --force`
Will install rxdb@3.0.2, which is a breaking change
node_modules/jsonpointer
  is-my-json-valid  >=2.0.0
  Depends on vulnerable versions of jsonpointer
  node_modules/is-my-json-valid
    rxdb  >=3.0.3
    Depends on vulnerable versions of graphql-client
    Depends on vulnerable versions of is-my-json-valid
    Depends on vulnerable versions of pouchdb-adapter-http
    Depends on vulnerable versions of pouchdb-core
    Depends on vulnerable versions of pouchdb-find
    node_modules/rxdb

node-fetch  <2.6.1
The `size` option isn't honored after following a redirect in node-fetch - https://github.com/advisories/GHSA-w7rc-rwvf-8q5r
fix available via `npm audit fix --force`
Will install rxdb@3.0.2, which is a breaking change
node_modules/node-fetch
node_modules/pouchdb-fetch/node_modules/node-fetch
  isomorphic-fetch  2.0.0 - 2.2.1
  Depends on vulnerable versions of node-fetch
  node_modules/isomorphic-fetch
    graphql-client  >=1.0.2
    Depends on vulnerable versions of isomorphic-fetch
    node_modules/graphql-client
      rxdb  >=3.0.3
      Depends on vulnerable versions of graphql-client
      Depends on vulnerable versions of is-my-json-valid
      Depends on vulnerable versions of pouchdb-adapter-http
      Depends on vulnerable versions of pouchdb-core
      Depends on vulnerable versions of pouchdb-find
      node_modules/rxdb
  pouchdb-fetch  >=7.1.0
  Depends on vulnerable versions of node-fetch
  node_modules/pouchdb-fetch
    pouchdb-abstract-mapreduce  >=7.1.0
    Depends on vulnerable versions of pouchdb-fetch
    node_modules/pouchdb-abstract-mapreduce
    pouchdb-adapter-http  >=7.1.0
    Depends on vulnerable versions of pouchdb-fetch
    node_modules/pouchdb-adapter-http
[...]

PR #3497, opened 2021-11-10
Vendor "graphql-client" and bump "is-my-json-valid" to 2.20.6

jesec:
## This PR contains:

A BUGFIX

#3495

npm emits a bunch of security warnings.

## Notes

I also migrated "graphql-client" to TypeScript. Though, there is still one warning left due to "pouchdb-fetch" (pouchdb/pouchdb#8408).


package.json
```diff
@@ -105,9 +105,9 @@
     "event-reduce-js": "1.4.1",
     "express": "4.17.1",
     "get-graphql-from-jsonschema": "8.0.16",
-    "graphql-client": "2.0.1",
     "is-electron": "2.2.0",
-    "is-my-json-valid": "2.20.5",
+    "is-my-json-valid": "^2.20.6",
+    "isomorphic-fetch": "^3.0.0",
```
pubkey, line 110: Please do not use version ranges.

pubkey:
I am not sure if this is a correct use of the license, when we just copy the whole code over to the RxDB repo.
RxDB uses apache2, while `graphql-client` has set `ISC` on the package.json, but the repo has no license file.

jesec:
> I am not sure if this is a correct use of the license, when we just copy the whole code over to the RxDB repo. RxDB uses apache2, while `graphql-client` has set `ISC` on the package.json, but the repo has no license file.

That project made it clear that it is under ISC with the license field in package.json.

Otherwise, if a repo indicates no license, that means no license. You can't use it in any fashion (not even as a dependency) when that's the case.

package.json
jesec: It is not a good idea for a library to pin dependencies. It prevents dependents from bumping packages, should the need arise, or if there is a security issue. In other cases, it might lead to out of sync dependency trees. Last, if you are busy for a period of time and can't publish new versions, the pinned dependencies will remain a PITA for all dependents. 

Anyways, I will leave that decision to you. I pushed the commit you requested.

pubkey:
I am really against copying this code into the RxDB repo.
ISC might be compatible, but it adds more complexity for users when they want to evaluate if the license is ok for them.
Also I would have to maintain this code for a long time which I do not want.
Isn't there another library that we can use?

jesec:
> I am really against copying this code into the RxDB repo.
> ISC might be compatible, but it adds more complexity for users when they want to evaluate if the license is ok for them.
> Also I would have to maintain this code for a long time which I do not want.
> Isn't there another library that we can use?

There might be. But that’s not going to be a drop in replacement. 

It is really not that complicated. It is the same thing when you choose to depend on the package. ISC is in no way viral.

package.json
pubkey: Yes all this is true in theory. But in practice, stuff breaks all the time, it did so in the past and it will so in the future.
When a new version is released, the renovate bot will make a PR to this repo with the updated version.

pubkey:
I really do not want to copy over code. I did this in the past. Years on from now, you will long be gone but I will still be sitting there maintaining the code.
Also most non-side-project projects I work with are full of people who are afraid of using any open source tool. I will not make this more complicated by mixing up licenses.

pubkey:
Closed since no answer.
`is-my-json-valid` was updated in ded4fdbfaf5b5deb282392799a1d9cb85259445a

pubkey:
Hi @jesec 
I have to admit, you where right, I was wrong.
The graphql-client will be replaced: https://github.com/pubkey/rxdb/pull/4101

PR #3519, opened 2021-11-23
Fix an error on server spawn when full leveldown-module is used

morngaraf:
## This PR contains:
 - A BUGFIX

When full leveldown-module is passed to `getRxStoragePouch`, [spawning a server](https://rxdb.info/rx-database.html#server) throws the following error:
```
Error: Invalid Adapter: function LevelDOWN (location) {
  if (!(this instanceof LevelDOWN)) {
    return new LevelDOWN(location)
  }

  if (typeof location !== 'string') {
    throw new Error('constructor requires a location string argument')
  }

  AbstractLevelDOWN.call(this, {
    bufferKeys: true,
    snapshots: true,
    permanence: true,
    seek: true,
    clear: true,
    createIfMissing: true,
    errorIfExists: true,
    additionalMethods: {
      approximateSize: true,
      compactRange: true
    }
  })

  this.location = location
  this.context = binding.db_init()
}
    at PouchAlt.PouchDB (/home/runner/work/rxdb/rxdb/node_modules/pouchdb-core/lib/index.js:1292:11)
    at new PouchAlt (/home/runner/work/rxdb/rxdb/node_modules/pouchdb-core/lib/index.js:1412:13)
    at /home/runner/work/rxdb/rxdb/node_modules/express-pouchdb/lib/create-or-delete-dbs.js:30:10
```

## Todos
- [ ] Documentation?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 Bugfixes:
   - LokiJS: Do not call `saveDatabase()` when no persistence adapter is given.
   - Query returns outdated result in second subscription [#3498](https://github.com/pubkey/rxdb/issues/3498) Thanks [@swnf](https://github.com/swnf)
+  - Spawning a server when full leveldown-module is used must not throw an error.
+
 ### 10.5.3 (19 November 2021)
 
 Bugfixes:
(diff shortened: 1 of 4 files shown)


pubkey:
Thank you for your work. Merged!

PR #3546, opened 2021-12-05
Unhandled Rejection (RxError (DOC19)): RxError (DOC19): Value of primary key(s) cannot be changed

nothingkid:
## This PR contains:
Bug demonstration

"Unhandled Rejection (RxError (DOC19)): RxError (DOC19): Value of primary key(s) cannot be changed"
Looks like you sort primaryKey values at some point of time, but sort mutates array


pubkey:
Thank you for the test. I was able to reproduce and fix the problem.

PR #3562, opened 2021-12-08
Failing test for: primary _id field causes "fieldnames do not match th…

SuperKirik:
…e regex" error

## This PR contains:
Bug demonstration

"fieldnames do not match the regex" error for _id field which is set as primary in schema


pubkey:
Thank you for the test.
I could reproduce and fix the problem.

chiru22:
HI, I am using "rxdb": "10.5.4" in my angular(v9) project. am facing the same problem as mentioned. can I know what is the solution?

Issue #3569, opened 2021-12-10
query will never executed in some cases (freezes forever)


4321ip:
occurs only if few tab is opened
here is simple example
![image](https://user-images.githubusercontent.com/11890535/145666865-831f907a-3dec-47cc-bb3f-be0c6d81b599.png)

if one page - its ok(see in log 'after exec')
on the second page 'after exec' will never printed (from onBtnClick)

_minimal example:_
```
git clone https://github.com/4321ip/test-rxdb-loki.git
cd test-rxdb-loki
npm install
npm run dev
```
1. open two pages http://localhost:8080/#/
2. on the second page - click clearDb button
3. see log

with multiInstance=false the problem is gone, but rxdb is useful due to Multi-Tab support(espesially with lokiJs)

pubkey:
@4321ip I am aware of the problem, I also have that in production.
It is fixed in the newest master branch.
I will release the beta of version 11 in a few days so that you can try.

pubkey:
I released the beta. Please try it out. https://github.com/pubkey/rxdb/blob/master/orga/releases/11.0.0.md

4321ip:
the issue is still there

pubkey:
Can you create a PR with a failing test case?

4321ip:
ок

4321ip:
I don't know how to repeat the bug in unit tests.
The error is present only when multiple browser tabs are open

4321ip:
@pubkey 
I don't know how to repeat the bug in unit tests.
The error is present only when multiple browser tabs are open

pubkey:
Hmm not sure. Maybe you can reproduce it in the e2e tests of the graphql example.
I can only fix things that I can reproduce.

4321ip:
@pubkey 
https://github.com/pubkey/rxdb/pull/3581

PR #3581, opened 2021-12-20
query will never executed in some cases

4321ip:
https://github.com/pubkey/rxdb/issues/3569
way to reproduce:
1. clone the whole [RxDB-repo](https://github.com/pubkey/rxdb)
2. go into project `cd rxdb`
3. run `npm install`  (you may need to run `npm install --legacy-peer-deps`)
4. go to this folder `cd examples/graphql`
5. run `npm install`
6. run `npm start`
7. open two tabs ~http://localhost:8888/~ [http://localhost:8888/?storage=lokijs&sync=false](http://localhost:8888/?storage=lokijs&sync=false)
8. on last tab - click "insert btn"
9. wait for 3 second


pubkey:
@4321ip thank you for providing more information.
So the `alert('FREEZE')` will fully block the JavaScript process. When the process is blocked, it cannot respond to a request from the other tab. If I understand this right, there is not much that can be done here.

pubkey:
Ah sorry. Now I understand. Event when the alert box is closed and the process is free again, the query does not finish.
I will investigate..

pubkey:
Ok so I checked again and there are multiple things that are unclear.
- Why do you call `.remove()` on the collection if it exists? This will delete all data and when it runs while the collection is removed, it leads to undefined behavior.
- Why is `multiInstance: true` set? This is meant for testing only. Can you reproduce the problem without creating a new database instance on each insert?

4321ip:
@pubkey 
with multiInstance=false the problem is gone, but rxdb is useful due to Multi-Tab support(espesially with lokiJs)
i call collection.remove() in my project when user logout to delete some sensitive data.
common workflow:
```
async logout() {
await collection1.remove()
await collection2.remove()
.....
await createCollection()
await createCollection()
}
```

pubkey:
@4321ip 
I added a fix now that ensures that the remote operation re-runs when the leader dies.
See 06d0b79777d0950e22756c56e98c44a6364b3c77
This is released in the latest beta version. Please check if this fixes your problem.

pubkey:
@4321ip 
I think I found it now. Released `11.0.0-beta.13`
It was an error in the broadcast-channel module where events got missed because of inacurate javascript timing.
[See](https://github.com/pubkey/broadcast-channel/commit/91a3e648ad0ffb58acd6267f8ee063d6d5e115ad#diff-845a6b22541d2f98bc0055d168291df1175041abe8117b54677b29fe9db8c0afR247-R263)

4321ip:
@pubkey 
Sorry to tell you, but still doesn't work

pubkey:
We need a test to reproduce this. 
The current code calls collection.remove() and recreates the collection. I am not sure if this is supposed to work.

4321ip:
> Multi-Tab

4321ip:
> We need a test to reproduce this. The current code calls collection.remove() and recreates the collection. I am not sure if this is supposed to work.
@pubkey 
its works! But only without multiTab support

PR #3583, opened 2021-12-23
Fixed vanilla example

dharmaturtle:
Vanilla example doesn't work

I think this is correct... at least the console doesn't throw errors and the demo seems to work.


pubkey:
Thank you, merged.

PR #3597, opened 2021-12-28
feat: strict mango dsl

raveclassic:
Hey @pubkey, I noticed a `TODO` comment regarding implementing the types for `MangoQuerySelector` and decided to implement them. Here's the PR, looking forward to some feedback.

## This PR contains:
IMPROVED typings;
A BREAKING CHANGE - the types are now stricter which may result in a breaking change for codebases that were previously using the types incorrectly.

Currently, `MangoQuerySelector` is completely untyped.

## Todos 
- [x] Typings
- [x] Changelog

## Changelog
`MangoQuerySelector` is now converted into a recursive union type consisting of 3 main subtypes - `MangoPropertyOperator`, `MangoCombinationOperator` and `MangoConditionOperator`. 

`MangoPropertyOperator` is suited for defining property based shortcuts: `id: 'foo'`, `id: { $eq: 'foo' }` etc.
`MangoCombinationOperator` is suited for defining combinations such as `$and`, `$or` etc.
`MangoConditionOperator` is suited for defining conditions such as `$eq`, `$in` etc.

Only `MangoPropertyOperator` and `MangoCombinationOperator` are allowed on the query's top level because `MangoConditionOperator` requires a target property.

### DIFF
--- a/src/rx-query.ts
+++ b/src/rx-query.ts
@@ -79,7 +79,7 @@ export class RxQueryBase<
 
     constructor(
         public op: RxQueryOP,
-        public mangoQuery: Readonly<MangoQuery>,
+        public mangoQuery: Readonly<MangoQuery<RxDocumentType>>,
         public collection: RxCollection<RxDocumentType>
     ) {
         if (!mangoQuery) {
@@ -390,7 +390,7 @@ export class RxQueryBase<
     }
 }
 
-export function _getDefaultQuery(): MangoQuery {
+export function _getDefaultQuery<RxDocType>(): MangoQuery<RxDocType> {
     return {
         selector: {}
     };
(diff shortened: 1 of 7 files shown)


raveclassic:
Meh, looks like there's some really crazy stuff in the `/src/plugins` directory which I've completely missed. There are some arbitrary and relaxed index signatures required for the plugins to work correctly which defeats the very purpose of having strict type checking for mango queries... :/ Plus, there's some strange behavior  - TS 4.5.2 just hangs sometimes trying to check the queries.
I'll leave this PR open in case I find time to investigate this, but feel free to close it.

pubkey:
I think we can type every internal usage of mango queries inside of `src/plugins/query-builder/` as `any`.
The query builder does not use 'correct' mango queries internally, so it might be better to not use the same types on them.

pubkey:
I am closing this.
PR still wanted. Ping me if I should reopen.

PR #3627, opened 2022-01-16
[GraphQL Replication] Only create a new push sequence if necessary

hdwatts:
## This PR contains:
 - A bugfix

I was working on a project using GraphQL replication without subscriptions and I left a browser tab open for an ungodly amount of time. To my surprise, my local storage had gotten bloated to the point where I couldn't access it anymore. While investigating I noticed that thousands of `push-checkpoints` were being created.

image

Indeed, I was able to confirm that a new push-checkpoint was made on every sync operation. @pubkey confirmed in Gitter [that this is indeed a bug](https://gitter.im/pubkey/rxdb?at=61e37eaf526fb77b317feec2).

My solution is to surround the `setLastPushSequence` with the same check that is in the `catch` block above. This works well in my testing. I'll note that this **doesn't technically solve the memory leak** and is only kicking it down the road to when thousands of pushes have been required. I would think there should be a cleanup operation on successful pulls that remove old and irrelevant push checkpoints - but this is early in my tooling with `rxdb`  and I'm very unfamiliar with both this codebase as well the intricacies of pouchdb.

In my testing, this is still working great and the records are no longer being created unnecessarily.

@pubkey wouldn't mind your assistance on if this is the right solution and what the best practice for the test should be

## Todos
- [x] Testing
- [ ] Changelog

### DIFF
--- a/test/unit/replication.test.ts
+++ b/test/unit/replication.test.ts
@@ -549,7 +549,7 @@ describe('replication.test.js', () => {
                 localCollection,
                 REPLICATION_IDENTIFIER_TEST
             );
-            assert.ok(originalSequence === newSequence);
+            assert.strictEqual(originalSequence, newSequence);
             localCollection.database.destroy();
             remoteCollection.database.destroy();
         });
--- a/test/unit/replication.test.ts
+++ b/test/unit/replication.test.ts
@@ -534,17 +534,17 @@ describe('replication.test.js', () => {
             });
 
             await replicationState.awaitInitialReplication();
-
+            await replicationState.run();
+            
             const originalSequence = await getLastPushSequence(
                 localCollection,
                 REPLICATION_IDENTIFIER_TEST
             );
             // call .run() often
-            await Promise.all(
-                new Array(3)
-                    .fill(0)
-                    .map(() => replicationState.run())
-            );
+            for (let i = 0; i < 3; i++) {
+                await replicationState.run()
+            }
+
             const newSequence = await getLastPushSequence(
                 localCollection,
                 REPLICATION_IDENTIFIER_TEST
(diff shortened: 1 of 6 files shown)


pubkey:
Change looks good.
We need at least one unit test.
Maybe you could subscribe to `collection.$` and count the writes to local documents. Then in the test you can make sure to not do more writes to local documents then needed.

hdwatts:
Thanks @pubkey - I've added tests that check to make sure the `lastPushSequence` isn't getting updated in both the replication primitives and graphql test suites. I've also implemented a potential fix in replication primitives.

Separately - thank youfor the compacting docs! This is exactly what I was looking for. As stated in my PR I'm still dipping my toes into PouchDB in general, so this is very helpful insight.

Let me know how else I can assist and I hope you have a nice Sunday

pubkey:
Code looks good.
Somehow the CI is failing now at another test, I am an not sure why.

hdwatts:
I'm investigating - it does appear to be failing for a legitimate reason caused by my code change.

hdwatts:
@pubkey I've adjusted the logic and now all tests are passing locally

test/unit/replication.test.ts
```diff
@@ -513,6 +513,47 @@ describe('replication.test.js', () => {
...
+            });
+
+            await replicationState.awaitInitialReplication();
+
+            const originalSequence = await getLastPushSequence(
+                localCollection,
+                REPLICATION_IDENTIFIER_TEST
+            );
+            // call .run() often
+            await Promise.all(
```
pubkey, line 543: Multiple calls to `run()` are automatically de-duplicated so that run() does never run in parallel.
Instead of using Promise.all() you should await each call before starting the next one.

```ts
for(var i = 0; i < 3; i++){
    await run();
}
```

hdwatts: Thanks - made the change and will push it once the current CI is finished

pubkey:
Great work, I merged it. Thank you.

Issue #3622, opened 2022-01-13
Add error handlers for replication push and pull


nirvdrum:
I apologize if this is provided in some way, but I haven't figured it out yet and I've read through _replication-graphql.test.ts_ looking for ways to handle replication push and pull errors. If there's agreement on the issue, I can try to pull together a PR.

I'm using GraphQL replication and I have a non-primary key column that should be unique in my database. In #728, adding support for local uniqueness checks was decided against due to potential conflicts during replication. Indeed, having that would only reduce the frequency with which I run into conflicts, although could make for a nicer user experience locally. In any event, sometimes I run into issues where the GraphQL push replication fails. When that happens, I have a local document already created but it does not have a persisted version in the central database.

I'd like a way to provide an error handle for the `push` property in the replicator setup. Having such a handler would allow me to gracefully handle the conflict. In this particular case, I can assign a new value to the column and attempt the insertion a second time. Unfortunately, the value is not something I can generate on the server. Alternatively to fixing the document, I could use that handler to delete the local document.

The `RxGraphQLReplicationState` class does have a way to observe errors via its `error$` field, but that is invoked on both push and pull. Moreover, it only has a reference to the GraphQL error, which doesn't have enough context to recover from. Nor is there a way to trigger a new replication attempt with the updated values.

I think what I'd like to see is an `onError` handler in `GraphQLSyncPushOptions`. For symmetry it might be worth adding it to `GraphQLSyncPullOptions`, although the only pull errors I've encountered have been when I've messed up a query and that's not recoverable. Although, I'm not entirely sure how that error handler could retry a push since there isn't a handle to the GraphQL client. Maybe the handler could just update the local document and let the normal retry logic replicating it during its normal retry cycle. That aside, does this sound like a reasonable feature to add to RxDB?

pubkey:
There is still the task open to reuse the replication-primitives plugin for the GraphQL replication. I do not really want to add more features into the GraphQL replication, until the code is unified.

Can we extend the `error$` observable to support your use case?

nirvdrum:
Do you have a link describing the refactoring you're talking about? I didn't see anything in the issue tracker.

I'll try extending the `error$` observable. I think that'll work and shouldn't adversely affect anyone. The error type is `any` at the moment, so I don't think anyone should be relying on any particular structure of that object.

nirvdrum:
This was addressed by #3630 and released with RxDB 11.5.0.

PR #3630, opened 2022-01-17
Record details about replication action and associated document, if any, upon replication failure.

nirvdrum:
## This PR contains:

 - A NEW FEATURE

This PR is a proposed solution to #3622. Errors encountered during GraphQL replication are wrapped in a new `ReplicationError` object that records whether it was a pull or push replication that failed, and if a push failure, the associated document data in the failed request.

For the time being, I've only applied these changes to GraphQL replication, although I think it could be handled for all forms of replication. There were a variety of design decisions made in getting this far and I'm happy to adjust as necessary. I just wanted to get something functional that could then be evaluated. Some areas that would be helpful to get feedback on are:

* Should the replication error payload be a discriminated union or just fields in the class? I went with the discriminated union because it made the constructor cleaner (less optional data), but I could also see that the replication type (pull or push) should be a field directly in the error.
* Should the replication error wrap all errors upon replication or just logical errors? E.g., GraphQL can have an HTTP 200 response with server-side logic errors encoded in the GraphQL response and I think those should always be wrapped/recorded. But, should we also wrap network errors, authentication errors, HTTP errors and so on? This PR does do that for consistency and because I think it could make logging and error handling easier in a client, but it makes the code a little uglier.
* If all errors are wrapped, I think we can safely indicate that the object supplied to `error$` in this case is a `ReplicationError`. I have kept it as `any` for the time being to avoid any future issues with the type definition changing. When typed as `any`, a client should not be relying on anything about the object structure.
* Should the replication error include all of the push document data or just the document ID?

To give an example of how I'm making use of this new feature, here's code from my client handling a uniqueness constraint violation from the GraphQL server:

```ts
replicationState.error$.subscribe((error: ReplicationError<Todo>) => {
  if (error.payload.type === 'push') {
    const { documentData } = error.payload;
    if (error.innerErrors && error.innerErrors[0].message === 'Uniqueness violation. duplicate key value violates unique constraint "todos_profile_id_rank_key"') {
      this.db.todos.findOne(documentData.id).exec().then((doc) => {
        doc?.update({
          $set: {
            rank: LexoRank.parse(documentData.rank).genNext().toString()
          }
        });
      })
    }
  }
```

## Todos 

- [ ] Documentation
- [ ] Changelog
- [ ] Evaluate design decisions
- [ ] Ensure testing is sufficient

### DIFF
--- a/docs-src/replication-graphql.md
+++ b/docs-src/replication-graphql.md
@@ -349,6 +349,52 @@ replicationState.error$.subscribe(error => {
 });
 ```
 
+GraphQL errors are wrapped in a `RxReplicationError`, which has a `payload` property with information to help you handle the underlying error.
+The payload has a `type` field with a value of `"pull"` or `"push"`, corresponding to an error in either a GraphQL pull or push replication operation, respectively.
+If the error occurs doing a _push_, the `payload` also contains a `documentData` property, which corresponds to the document data supplied to the push query builder.
+**Notice:** You may see errors in this observable that are not `RxReplicationError`.
+Replication may fail for reasons unrelated to the GraphQL service.
+E.g., your PouchDB or LokiJS database may have issues in which case a general error will be generated and passed on.
+
+As an example, you can try to recover from errors like so:
+
+```js
+replicationState.error$.subscribe((error) => {
+  if (error.payload) {
+      if (error.payload.type === 'pull') {
+          console.log('error pulling from GraphQL server', error.innerErrors);
+      } else if (error.type === 'push') {
+          if (error.innerErrors && error.innerErrors.length > 0) {
+              const graphQLError = error.innerErrors[0];
+              
+              // In this hypothetical case, there's a remote database uniqueness constraint being violated due to two
+              // clients pushing an object with the same property value. With the document data, you can decide how best
+              // to resolve the issue. In this case, the client that pushed last "loses" and we delete the object since
+              // the one it conflicts with will be pulled down during the next pull replication event.
+              // The `graphQLError` structure is dictated by your remote GraphQL service. The field names are likely
+              // to be different.
+              if (graphQLError.code === 'constraint-violation' && graphQLError.constraintName === "unique_profile_name") {
+                  this.db.profiles
+                      .findOne(documentData.id)
+                      .exec()
+                      .then((doc) => {
+                          doc?.remove();
+                      });
+              }
+          } else {
+              console.log('error pushing document to GraphQL server', documentData);
+          }
+      } else {
+          console.log('Unknown replication action', error.payload.type);
+      }
+  } else {
+      // General error occurred. E.g., issue communicating with local database.
+      console.log('something was wrong');
+      console.dir(error);
+  }
+});
+```
+
 #### .canceled$
 
 An `Observable` that emits `true` when the replication is canceled, `false` if not.
--- a/src/plugins/replication-graphql/index.ts
+++ b/src/plugins/replication-graphql/index.ts
@@ -58,9 +58,9 @@ import { getDocumentDataOfRxChangeEvent } from '../../rx-change-event';
 import {
     _handleToStorageInstance
 } from '../../rx-collection-helper';
+import {RxReplicationError} from '../replication';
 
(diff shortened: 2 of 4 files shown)


nirvdrum:
I'm not positive I'm testing this properly either, so please that give a good look. It also took me a very long time to get the tests in a good state. Is there a way to easily run just a single test or a way to debug one? I'm not familiar at all with the test runner being used. Consequently, my solution was to delete a bunch of tests temporarily and then run all remaining tests everytime I wanted to make a change. That still took several minutes, making this a very slow development effort.

pubkey:
Yes the test runner is pretty outdated. I also just commend out tests [here](https://github.com/pubkey/rxdb/blob/master/test/unit.test.ts)
You can run `npm run test:fast:lokijs` or `npm run test:fast:pouchdb` to run all tests in parallel which speeds up things.

Public facing types are prefixed with `Rx` so it should be `RxReplicationError`. The rest of the code looks good to me.
Documentation must still be updated.

nirvdrum:
Thanks for the review. It'll probably be a few days before I can get everything updated.

nirvdrum:
@pubkey I've rebased the branch against _master_ and updated the GraphQL Replication docs. I have two outstanding questions:

  1.  Currently, only interactions with the GraphQL server are wrapped in `RxReplicationError`. I have seen non-`RxReplicationError` errors show up in my error handler because of issues related to the underlying database (e.g., schema changed without migration). I can wrap all of `runPull` and `runPush` and wrap all exceptions, but I don't know if that's desirable.
  2. Do you want this applied to custom and CouchDB replication as well? If so, I can give it a shot and add tests for it. I just don't have any real world application using either, so it's difficult for me to really exercise the functionality. I suspect CouchDB replication has fewer sources of issues than GraphQL replication does since it's synchronizing at the DB level.

I know there's a desire to base GraphQL replication on custom replication. I can help with that in a future PR. When they're merged, I'll have a much easier time testing the error handling since I'll be using it in my own application. But, if you want the error wrapping added for custom replication now, that's sensible, too.

When you get a chance, please take a look and let me know what you'd like to see done before approving the PR. Thanks!

docs-src/replication-graphql.md
```diff
@@ -349,6 +349,52 @@ replicationState.error$.subscribe(error => {
 });
 ```
 
+GraphQL errors are wrapped in a `RxReplicationError`, which has a `payload` property with information to help you handle the underlying error.
+The payload has a `type` field with a value of `"pull"` or `"push"`, corresponding to an error in either a GraphQL pull or push replication operation, respectively.
+If the error occurs doing a _push_, the `payload` also contains a `documentData` property, which corresponds to the document data supplied to the push query builder.
+NB: You may see errors in this observable that are not `RxReplicationError`.
```
pubkey, line 355: Is "NB:" for "nota bene"?
We shouldn't use these abbreviation because most readers of the docs a from non-latin based mother languages.
In other parts I did it like [here](https://github.com/pubkey/rxdb/blob/2725b4f3f2ad7087521a442c83762be4b5d46f65/docs-src/rx-schema.md?plain=1#L166)

pubkey:
We do not need this for the couchdb replication for now.

I have no opinion about having non-RxReplicationError in the same stream. We can leave that for now. Maybe we should add an 'internal' error type in the future. Or some other way to determine if an error is expected or a problem in the code.

If you want to, you can add the improved error handling also to the replications plugin. But I am also ok with not having that for now.
Migrating the GrahphQL replication to use the replication-primitives will likely cause a breaking change which requires a new major version. This is why I did not do that for now because I want to first collect more breaking changes.

Code looks good for me, I added [one comment](https://github.com/pubkey/rxdb/pull/3630#pullrequestreview-866958382) to the docs.

docs-src/replication-graphql.md
nirvdrum: Thanks for taking a look. I hadn't considered the language barrier. Changing it is easy enough. Is it okay to use "e.g."? That's another Latin abbreviation.

nirvdrum:
Thanks for the review. I've updated the documentation issue you highlighted.

PR #3655, opened 2022-01-28
fix code example for replication primitives

mmouterde:
## This PR contains:
improved docs:  code example fix.

minor fix : missing awaits and wrong variables.


mmouterde:
forget about this PR I will split in two PR

PR #3656, opened 2022-01-28
Update SUMMARY.md

mmouterde:
fix menu to sync functions (lowercase to fit generated ids).


pubkey:
Thank you.
Merged.

mmouterde:
You're welcome, you are so reactive, thanks !

Note that `$` char also breaks generated links : `findByIds$()` does not work too.
https://stackoverflow.com/a/70034418 could be a workaround ?

pubkey:
The long term plan is to move away from gitbook because it is unmaintained. Also see here https://github.com/pubkey/rxdb/blob/master/package.json#L83

If you find any workaround, I will merge it. But make sure that the plain markdown file is still valid markdown.

Issue #3659, opened 2022-01-31
[Bug] collection.findByIds$ behaves weirdly


Hideman85:
Tested today on master and it does the same weird behavior than on my app.
- The subscription is triggered so much time
- Sometime the map does not have the objects even if the store have it for sure + funny thing executing findByIds(...) will get the objects
- The subscription is trigerred even if the update do not concerns the list of ids looking for

<details>
<summary>I made a repro out of bug-report test unit</summary>

```ts
/**
 * this is a template for a test.
 * If you found a bug, edit this test to reproduce it
 * and than make a pull-request with that failing test.
 * The maintainer will later move your test to the correct possition in the test-suite.
 *
 * To run this test do:
 * - 'npm run test:node' so it runs in nodejs
 * - 'npm run test:browser' so it runs in the browser
 */
import assert from 'assert';
import AsyncTestUtil from 'async-test-util';
import config from './config';

import {
    createRxDatabase,
    randomCouchString
} from '../../plugins/core';

import {
    getRxStoragePouch,
} from '../../plugins/pouchdb';
import {
    RxReplicationStateBase,
} from '../../plugins/replication';


describe('bug-report.test.js', () => {
    it('should fail because it reproduces the bug', async () => {

        /**
         * If your test should only run in nodejs or only run in the browser,
         * you should comment in the return operator and addapt the if statement.
         */
        if (
            config.platform.isNode() // runs only in node
            // config.platform.isNode() // runs only in the browser
        ) {
            return;
        }

        // create a schema
        const mySchema = {
            version: 0,
            primaryKey: 'passportId',
            type: 'object',
            properties: {
                passportId: {
                    type: 'string'
                },
                firstName: {
                    type: 'string'
                },
                lastName: {
                    type: 'string'
                },
                age: {
                    type: 'integer',
                    minimum: 0,
                    maximum: 150
                }
            }
        };

        // generate a random database-name
        const name = randomCouchString(10);

        // create a database
        const db = await createRxDatabase({
            name,
            storage: getRxStoragePouch('memory'),
            eventReduce: true,
            ignoreDuplicate: true
        });
        // create a collection
        const collections = await db.addCollections({
            mycollection: {
                schema: mySchema
            }
        });


        //  Simple helper to create data
        const createObject = id => ({
            passportId: id,
            firstName: id,
            lastName: id,
            age: 56
        })



        //  Record subscription
        const updates = []
        const errors = []

        const sub = collections.mycollection.findByIds$([
[...]

PR #3661, opened 2022-01-31
Bug report for collection.findByIds$(...)

Hideman85:
Bug report submission for https://github.com/pubkey/rxdb/issues/3659


Hideman85:
I think missed something I did commit a new file `test/unit/bug-findbyids.test.ts` to not override the sample file but it seems to be ignored by testing suite (except coding style of course :D)

Hideman85:
Now the test is failing according to what I observed

test/unit/bug-findbyids.test.ts
```diff
@@ -0,0 +1,180 @@
...
+
+describe('bug-report.test.js', () => {
+    it('should fail because it reproduces the bug', async () => {
+
+        /**
+         * If your test should only run in nodejs or only run in the browser,
+         * you should comment in the return operator and addapt the if statement.
+         */
+        if (
+            config.platform.isNode() // runs only in node
```
pubkey, line 36: Does it also fail in node?

Hideman85: I don't know I was not able to run node locally but yes this line is useless, I should remove it

Hideman85: The result on node is exactly the issue I'm facing too.
The observer get called for each single update and the map do not have the docs.
```
Got update Map(4) {
  'a' => undefined,
  'b' => undefined,
  'c' => undefined,
  'd' => undefined
}
```
Then this is a bug with `findByIds$`

PR #3671, opened 2022-02-03
Bug report lokijs collection.find()

Hideman85:


pubkey:
Now I understand the problem.
Sorry this was missing in the docs.
For the replication primitives, the `_deleted` field must be provided in the push and pull handlers.
Only the graphql replication does a mapping.

Hideman85:
Yes that is basically the only issue, I was very confused, batch get was working, get by id too but list always empty :D
With _deleted set everything works like a charm :) and can definitely be an alternative to PouchDB for me right now.
Later we will see if the data start to be too big and we will see when it happen :)

pubkey:
I added this to the docs and also added a dev-mode check to ensure the `_deleted` field is not missing:
https://github.com/pubkey/rxdb/commit/3b292973e2eb525c18551d210d2a35166c0c4eb2

Issue #3749, opened 2022-03-27
Migrations do not work with LokiJS adapter


11bit:
Hello! 

Looks like data migrations (`rxdb/plugins/migration`) don't work when I am using LokiJS adapter: migration function does run, but all the data just disappears (surprisingly it somewhat returns back after a few refreshes of the page). 

I found that unit tests for migrations work only for pouchdb and I don't know how to make them work with loki or other adapters. So I made a [repository](https://github.com/11bit/repro-rxdb-lokijs-migrations-bug) with the bug reproduction. It is nextjs project with two pages: 

1. Db is initialized here https://github.com/11bit/repro-rxdb-lokijs-migrations-bug/blob/main/db/db.ts
2. On the index page I am wiping all the data from indexeddb and then creating a simple document with schema version 0 https://github.com/11bit/repro-rxdb-lokijs-migrations-bug/blob/main/pages/index.tsx#L11-L17

3. On the '/migrate' page I am trying to migrate the created document to schema version 1 https://github.com/11bit/repro-rxdb-lokijs-migrations-bug/blob/main/pages/migrate.tsx#L11-L19

After running '/migrate' page I don't see any documents at all. But if I refresh the page a couple of times I can see an old document without any migrations applied. However, If I switch to pouchdb adapter, everything works well. 

Is it something I am doing wrong with migrations or it's a bug?

pubkey:
To fix the tests for migration,
each of the test storages must have a method `getPersistendStorage()` similate to [getStorage](https://github.com/pubkey/rxdb/blob/master/test/unit/config.ts#L70).
Then we could simulate a correct migration in the unit tests and also fix the lokijs migration.

alex-hladun:
Can confirm we are seeing the same behaviour, can only be solved by db reset.

11bit:
@pubkey Can we copy that setup from https://github.com/pubkey/rxdb/blob/master/test/performance.test.ts#L64-L68

pubkey:
yes coping from there might work

patrickcze:
@pubkey is this something that would be a relatively easy fix

pubkey:
This might be fixed in the latest beta release. Please try.

11bit:
Works perfectly. Thank you very much!

@pubkey One more question. If I open two browser tabs simultaneously and then run migration in one of them then: 
1. With lokijs it crashes with `SNH` error
2. With pouch it seems to be working for tab with migration but the second tab looses connection

[...]

PR #3752, opened 2022-03-28
Turn on migration tests with lokijs adapter

11bit:
Turn on migration tests with lokijs adapter for  #3749. 

Should I add it to separate `getPersistedStorage` method and use it only in `migrations.test.ts` or `getStorage` is ok?


pubkey:
getStorage is used to have a fast storage, mostly based on memory. It might or might not be persistent.
getPersistedStorage must be persistend so that when a storage instance gets closed and opened again, the data must still be there.

11bit:
So, I can do the following:
1. Copy `getPersistendStorage` from `performance.test.ts` config to `config.ts`, so that config will have both getStorage and getPersistendStorage (there is no persisted storage for Dexie though)
2. Replace all occurencies of `getStorage` with `getPersistendStorage` in `data-migration.test.ts`

Please let me know if I understood you correctly and those changes are useful.

pubkey:
Yes I think that goes into the right direction.
We can have a persistend getStorage for dexie, but only in the browser, not in node.js

11bit:
@pubkey I've copied config from `performance.test.ts`. Please have a look

11bit:
- Rebased on the latest master (lokijs migration issue is gone)
- Added test for running migrations with multiple instances (loki fails)

pubkey:
Atm this fails because the attachments test should not run with the lokijs storage.

```Error: Attachments are not implemented in the lokijs RxStorage. Make a pull request.```

Do you still have plans on working on this?

11bit:
Yes, I'll rebase and see what's changed today

11bit:
@pubkey I rebased the branch and skipped attachments tests for lokijs

11bit:
Migration seems to be working ok. The exception happens after the migration when an actual query on a second database happens: 
`const doc = await col2.findOne().exec();`
So I guess it's something related to leader election mechanics?

pubkey:
Yes it looks like we need a check somewhere to detect if the instance is closed already and we must cancel the migration.

pubkey:
Are you still working on this?

11bit:
Unfortunately didn't have time for it yet

pubkey:
Closing this. Ping me when I should reopen this.

PR #3766, opened 2022-04-16
test-bug: Retrieving docs using .find() are automatically sorted by the primary key in alphabetical order (lokijs)

AyluinReymaer:
## This PR contains:
A failing test showcasing a bug.

When using lokijs with RxDB, when trying to retrieve docs that were previously inserted in a collection, they are sorted alphabetically on the primary key by default.

### DIFF
--- a/config/karma.conf.js
+++ b/config/karma.conf.js
@@ -1,3 +1,4 @@
+/* eslint-disable linebreak-style */
 
 
 // while the karma tests run, we need some things which we start here
@@ -13,8 +14,7 @@ const configuration = {
     basePath: '',
     frameworks: [
         'mocha',
-        'browserify',
-        'detectBrowsers'
+        'browserify'
     ],
     browserify: {
         debug: true,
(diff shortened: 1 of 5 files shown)


AyluinReymaer:
I've had to change the karma.config.js file to remove the detect browsers plugin and also install a plugin that allows tests to be done in Edge Chromium.

The current config in karma fails when testing with Edge Chromium.

pubkey:
Is this a RxDB or a lokijs problem?
Can you check if the sort order is correct when using plain lokijs?

pubkey:
@AyluinReymaer ?

AyluinReymaer:
Hello @pubkey ,

Sorry for the late reply, I didn't notice that there was a reply and only noticed your latest reply.

I've tested with plain lokijs and confirmed that it works.

I've pushed a new commit with the changes and no longer receive any failures.

AyluinReymaer:
I've just noticed that I didn't setup an index in the test as per:
http://techfort.github.io/LokiJS/tutorial-Indexing%20and%20Query%20performance.html

I'll test again with the index this time and check if it works

AyluinReymaer:
Ok, tested again with the appropriate index field.
It also works with the index added.

pubkey:
What exactly do you mean by 'it works'.
Does it only work with plain Loki or also with RxDB?

AyluinReymaer:
When using RxDB using the lokijs adapter, if I attempt to create a collection of heroes in which the primary key is the 'name' field, RxDB will by default sort the documents in alphabetical order according to the primary key.

So, if I add documents with heroes named Batman, Antman, Deadpool and Captain America in this order, the documents will be sorted incorrectly as follows:
Antman, Batman, Captain America, Deadpool.

However, if I use plain Lokijs and add the same documents in the same order as in RxDB, the documents will be sorted as correctly as follows:
Batman, Deadpool, Antman, Captain America.

So, using RxDB, the collection is ordered by the primary key alphabetically, while using plain Lokijs, the collection is in the order in which the documents were added.

pubkey:
RxDB will always add the primary key to the sorting to ensure we have a deterministic sorting.
This is needed for the event-reduce algorithm.

Insert order does not affect query results in RxDB.

AyluinReymaer:
I understand now.
If it is necessary for the algorithm, then there is nothing that can be done as it shouldn't be considered a bug.

I tested with RxDB by adding an ID field to the schema and setting it as a pk and it properly sorts as expected.

I recommend however, updating the documentation to let others know of this behavior.

In my build for example, I use UUIDs as the pk for my project. Due to this, the entries are being incorrectly sorted which is not good for my project.

I've pushed another commit with a change in the documentation. This should be good enough to let others know regarding this behavior so please let me know if you agree with the suggested documentation improvement.

pubkey:
I am sorry. I know you had spend a lot of time on this and there was something missing in the documentation.
I added a notice block here https://github.com/pubkey/rxdb/commit/824300bbde745ee50d40a38d4e24c3d4c63981c4

PR #3772, opened 2022-04-22
fix behaviour to fit with documentation in back channel use.

mmouterde:
## This PR contains:
 - A BUGFIX

Actual:
- no push on document updates if live=false
- no way to disable polling: liveInterval=0 is rejected by ensureNotFalsy() line 581 when live=true

Expected:
- do not poll if liveInterval=0 and live=true
- push triggered on document updates

### DIFF
--- a/src/plugins/replication/index.ts
+++ b/src/plugins/replication/index.ts
@@ -575,7 +575,7 @@ export function replicateRxCollection<RxDocType>(
          * if it is a live replication.
          */
         if (replicationState.live) {
-            if (pull) {
+            if (pull && replicationState.liveInterval > 0) {
                 (async () => {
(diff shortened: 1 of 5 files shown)


pubkey:
Code looks good.
We need at least 1 unit test to ensure the behavior is correct.

pubkey:
I mean we need a test that reproduces the replication behavior that did not work without the fix.

mmouterde:
Sure! I'm working such tests.
Not so easy without mock or `assert.CallTracker` to test function calls.
Do you have any example or way you usually test this kind of test ?

pubkey:
Looks good to me, I merged it. Thank you @mmouterde

PR #3774, opened 2022-04-26
add new option `runInitialReplication`.

mmouterde:
## This PR contains:
- A NEW FEATURE PROPOSAL

`replicateRxCollection()` provides a way to configure the way to replicate a given collection.
However, by default, this function also run an initial replication as *side effect*.

IMHO, it's a matter of separation of concerns: 
- The moment to configure collections and replications is not always the same we get ready to do the first replication.
- I would like to trigger replications collection by collection and keep control on the order, the timing... and separately from the replication definition.

I suggest to add an option to prevent the default first replication and separate the replication configuration and triggering.

This PR adds new option `runInitialReplication`.
It could be set to `false` to avoid  initialReplication on replicateRxCollection().
default to true to avoid breaking change

Please consider this PR as a draft and support for discussions.
I know this issue is an opinionated one, feel free to challenge it ;)

- TODO `liveInterval>0` case (poll from the first `run()` call instead of starting on `replicateRxCollection()`)

### DIFF
--- a/docs-src/replication.md
+++ b/docs-src/replication.md
@@ -54,6 +54,7 @@ The deleted field must always be exactly `_deleted`. If your remote endpoint use
 ## The replication cycle
 
 The replication works in cycles. A cycle is triggered when:
+  - When calling `replicateRxCollection()` by default
   - Automatically on writes to non-[local](./rx-local-document.md) documents.
   - When `liveInterval` is reached from the time of last `run()` cycle.
   - The `run()` method is called manually.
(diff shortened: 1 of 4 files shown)


pubkey:
There was the `autostart` parameter in the past: https://github.com/pubkey/rxdb/blob/450574602e85b3607b7069f1988b30b30fe54266/src/plugins/replication-graphql/index.js#L385
I am not sure how this got removed.

mmouterde:
Probably when you factorized replications to use primitives 08f78fa5f4c1ce53ca9a0449a391d1c5209fa9ae

I will rename my option to `autoStart`

I'm working on the graphQL test impacts.

mmouterde:
replaced by https://github.com/pubkey/rxdb/pull/3775

PR #3775, opened 2022-04-26
 add new option autoStart.

mmouterde:
## This PR contains:

- A NEW FEATURE PROPOSAL

`replicateRxCollection()` provides a way to configure the way to replicate a given collection.
However, by default, this function also run an initial replication as side effect.

IMHO, it's a matter of separation of concerns:

The moment to configure collections and replications is not always the same we get ready to do the first replication.
I would like to trigger replications collection by collection and keep control on the order, the timing... and separately from the replication definition.

I suggest to add an option to prevent the default first replication and separate the replication configuration and triggering.

This PR adds new option `runInitialReplication`.
It could be set to false to avoid  `initialReplication` on `replicateRxCollection()`.
default to `true` to avoid breaking change

Please consider this PR as a draft and support for discussions.
I know this issue is an opinionated one, feel free to challenge it ;)

### DIFF
--- a/test/unit/replication.test.ts
+++ b/test/unit/replication.test.ts
@@ -547,24 +547,28 @@ describe('replication.test.js', () => {
         });
         it('should push data even if liveInterval is set to 0', async () => {
             const {localCollection, remoteCollection} = await getTestCollections({local: 0, remote: 0});
-            const replicationState = replicateRxCollection({
+            let callProof = null;
+            replicateRxCollection({
                 collection: localCollection,
                 replicationIdentifier: REPLICATION_IDENTIFIER_TEST,
                 live: true,
                 liveInterval: 0,
+                autoStart: false,
                 push: {
                     handler() {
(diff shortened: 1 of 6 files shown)


test/unit/replication.test.ts
```diff
@@ -547,32 +547,68 @@ describe('replication.test.js', () => {
...
+                    collection: {
+                        database: {},
+                        onDestroy: {then(){}}
+                    } as RxCollection,
+                    replicationIdentifier: REPLICATION_IDENTIFIER_TEST,
+                    live: false,
+                    autoStart: false,
+                    waitForLeadership: false
+                });
+                // by definition awaitInitialReplication would be infinite
```
pubkey, line 605: You should await a short time here to ensure that is is not timing dependend.

``` await wait(100)```

pubkey:
Merged, thank you very much.

PR #3777, opened 2022-04-27
use findOne().$ to subscribe a document, but the document's reference is not equal to the document's reference which created by newDocument

qinyang912:
## This PR contains:
bug report

When I use findOne().$ before newDocument, the reference of document which get from findOne().$ is not equal to the reference of document which created by newDocument.

But if I use findOne().$ after newDocument, the reference is right

stackblitz: https://stackblitz.com/edit/typescript-6usdjm?file=index.ts

rxdb@11.6.1 can also reproduce the bug

rxdb@10.5.4 is right

### DIFF
--- a/test/unit/bug-report.test.ts
+++ b/test/unit/bug-report.test.ts
@@ -130,4 +130,54 @@ describe('bug-report.test.js', () => {
         db.destroy();
         dbInOtherTab.destroy();
     });
+
+    it('use findOne().$ to subscribe a document, the reference should equal to document which created by newDocument', async () => {
+        // generate a random database-name
+        const name = randomCouchString(10);
+
+        // create a schema
+        const mySchema = {
+            version: 0,
+            primaryKey: 'id',
+            type: 'object',
+            properties: {
+                id: {
+                    type: 'string',
+                    maxLength: 100,
+                },
+                name: {
+                    type: 'string'
+                }
+            },
+        };
+
+        // create a database
+        const db = await createRxDatabase({
+            name,
+            /**
+             * By calling config.storage.getStorage(),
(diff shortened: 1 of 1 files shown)


pubkey:
I can reproduce the problem, thank you for reporting.

In the RxStorage interface, it is expected that an event is emitted **before** the write operation is finished.
Therefore the query emits the data before RxDB can know that the write succeeded and that is why a new RxDocument will be created for the emitted insert event.

In the past, this was not a problem because pouchdb emitted the event much time after the write finished.

I do not see a way to fix this. Also I think the whole temporary document topic should be deprecated.
Many things are not clear, for example what happens if multiple temporary documents (with the same primary key) are inserted at the same time.

Can you explain your use case on why you have to use `.newDocument()` instead of just storing the json data an later insert that into RxDB?

pubkey:
Rel.

```ts
it('#3777 use findOne().$ to subscribe a document, the reference should equal to document which created by newDocument', async () => {
            const c = await humansCollection.create(0);
            const docId = randomString(10);
            let docOuter;
            c.findOne(docId).$.subscribe((doc) => (docOuter = doc));
            const docNew = c.newDocument({
                passportId: docId,
                age: 20,
                firstName: randomString(10),
                lastName: randomString(10)
            });
            await docNew.save();
            assert.ok(docOuter === docNew);
            c.database.destroy();
        })
```

pubkey:
Closing this since now answer.

Temporary documents will be removed in the next major version.
https://github.com/pubkey/rxdb/blob/master/orga/before-next-major.md#remove-temporary-documents

qinyang912:
Sorry for the late reply, I found that the code that used `.newDocument` can be implemented with `.insert`

PR #3786, opened 2022-05-04
fix async issue with createRxCollection

mmouterde:
- A BUGFIX
`await createRxCollection()` is resolved before the collection effectively exists.

see https://github.com/pubkey/rxdb/pull/3785/files#diff-8ea39e301dabaa32bb709cda31fb3daba268eda0f3ca80b9d55496af53edb96dR691
where `wait(100)` is required before testing the `createRxCollection` result.

@pubkey I'm not really sure about underlying impacts of that change.


pubkey:
The state is internally handled as `Promise` to ensure that it never happens that the same state is created multiple times in parallel.
With your change, calling `getDexieDbWithTables()` twice, would create more then one states.

mmouterde:
ok, I understand, sorry.

PR #3788, opened 2022-05-05
report bug : data still there despite remove and destroy calls

mmouterde:
- A BUGFIX

data still there despite remove and destroy calls

### DIFF
--- a/src/plugins/dexie/rx-storage-instance-dexie.ts
+++ b/src/plugins/dexie/rx-storage-instance-dexie.ts
@@ -342,7 +342,6 @@ export class RxStorageInstanceDexie<RxDocType> implements RxStorageInstance<
         await Promise.all([
             state.dexieDeletedTable.clear(),
             state.dexieTable.clear()
-            /* why not dexieChangesTable ? */
         ]);
         return this.close();
     }
--- a/src/plugins/dexie/rx-storage-instance-dexie.ts
+++ b/src/plugins/dexie/rx-storage-instance-dexie.ts
@@ -342,6 +342,7 @@ export class RxStorageInstanceDexie<RxDocType> implements RxStorageInstance<
         await Promise.all([
             state.dexieDeletedTable.clear(),
             state.dexieTable.clear()
+            /* why not dexieChangesTable ? */
         ]);
         return this.close();
     }
--- a/src/rx-database.ts
+++ b/src/rx-database.ts
@@ -586,7 +586,7 @@ export async function _removeAllOfCollection(
     const docs = await getAllCollectionDocuments(rxDatabase.internalStore, rxDatabase.storage);
     const relevantDocs = docs
         .filter((doc) => {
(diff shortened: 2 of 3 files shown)


pubkey:
`destroy` destroys the javascript object, it will  not remove the data. I know that this is a bad naming, maybe we should change that in the future.

`remove` however should really remove the data.

Both, `remove` and `destroy` assume that the collection/database is no longer used. Your second query after the remove will likely return data from the query cache, so it makes you think that the data is still there.
We could add a check to every function call to check if the db/collection is already destroyed, but this would add to much overhead I think.

src/plugins/dexie/rx-storage-instance-dexie.ts
```diff
@@ -342,6 +342,7 @@ export class RxStorageInstanceDexie<RxDocType> implements RxStorageInstance<
         await Promise.all([
             state.dexieDeletedTable.clear(),
             state.dexieTable.clear()
+            /* why not dexieChangesTable ? */
```
mmouterde, line 345: I guess this could be an issue

pubkey: You are right. The changes table must also be cleared.

pubkey: Ah incorrect.
I removed the `dexieChangesTable` in the past.
It was never used so it is not required to be cleared.

pubkey:
I tried to reproduce the problem with this test, but it didn't have any error.

```ts
/**
                 * @link https://github.com/pubkey/rxdb/pull/3788
                 */
                it('#3788 removing the collection should also remove all changes', async () => {
                    if (!config.storage.hasMultiInstance) {
                        return;
                    }

                    const dbName = randomCouchString();

                    const createDb = async () => {
                            name: dbName,
                            storage: config.storage.getStorage(),
                            ignoreDuplicate: true
                        });
                        await db.addCollections({
                            human: { schema: schemas.human }
                        });
                        return db;
                    }

                    const db1 = await createDb();
                    await db1.collections.human.insert(schemaObjects.simpleHuman());
                    const db2 = await createDb();

                    // remove the collection on one database
                    await db1.human.remove();


                    /**
                     * Getting the changes in the other database should have an empty result.
                     */
                    const changesResult = await db2.human.storageInstance.getChangedDocumentsSince(10);

                    console.dir(changesResult);

                    assert.strictEqual(changesResult.length, 0);

                    db1.destroy();
                    db2.destroy();
                });
```

mmouterde:
Also linked to dash in names https://github.com/pubkey/rxdb/pull/3788/commits/427e226c8b3fd37c50346c1e2ef4ce1cdb82b236#diff-e3288a7097a47ea2904142245c2895726f20a7be84914f4f4bb889b2adc4148bR589

pubkey:
Closing this PR. Fixed in ca6ea15615157b83471df990ec61362c6576136b

PR #3791, opened 2022-05-11
ADD data extraction function for GraphQL replication

WonderPanda:
## This PR contains:
 - A NEW FEATURE

Currently the ability to extract data from the GraphQL query result inside the pull handler for GraphQL replication feels very limited. There is no way to handle nested arrays using only data path based extraction

## Todos
- [x] Tests
- [ ] Documentation
- [x] Typings
- [ ] Changelog

### DIFF
--- a/src/plugins/replication-graphql/index.ts
+++ b/src/plugins/replication-graphql/index.ts
@@ -151,9 +151,14 @@ export function syncGraphQL<RxDocType>(
                     }
                 }
 
-                const dataPath = pull.dataPath || ['data', Object.keys(result.data)[0]];
-                const docsData: any[] = objectPath.get(result, dataPath);
-
+                let docsData: any[] = [] 
+                if (!pull.dataPath || typeof pull.dataPath === 'string') {
+                    const dataPath = pull.dataPath || ['data', Object.keys(result.data)[0]];
+                    docsData = objectPath.get(result, dataPath);
+                } else {
+                    docsData = pull.dataPath(result)
+                }
+                
                 // optimization shortcut, do not proceed if there are no documents.
                 if (docsData.length === 0) {
                     return {
--- a/src/types/plugins/replication-graphql.d.ts
+++ b/src/types/plugins/replication-graphql.d.ts
@@ -27,7 +27,7 @@ export interface GraphQLSyncPullOptions<RxDocType> {
      */
     batchSize: number;
     modifier?: (doc: RxDocType | any) => Promise<any> | any;
-    dataPath?: string;
+    dataPath?: string | ((result: any) => any[]);
 }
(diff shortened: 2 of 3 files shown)


pubkey:
It is using objectPath to get the field.
So you already can do


```ts
const replicationState = c.syncGraphQL({
                    url: server.url,
                    pull: {
                        batchSize,
                        queryBuilder: collectionQueryBuilder,
                        dataPath: 'result.data.collectionFeedForRxDBReplication.collection'
                    },
                    deletedFlag: 'deleted'
                });
```

WonderPanda:
> It is using objectPath to get the field. So you already can do
> 
> ```ts
> const replicationState = c.syncGraphQL({
>                     url: server.url,
>                     pull: {
>                         batchSize,
>                         queryBuilder: collectionQueryBuilder,
>                         dataPath: 'result.data.collectionFeedForRxDBReplication.collection'
>                     },
>                     deletedFlag: 'deleted'
>                 });
> ```

Please see the details in the associated issue in #3792. I realize that this particular test could have been implemented using dataPath but there are lots of GraphQL queries depending on the backend which can't be properly extracted using dataPath alone

WonderPanda:
@pubkey I updated `dataPath` here so it could accept either a string for a function to extract from the query result but I could see that it might be better to have a separate config property called for example `dataExtractor` or `dataMapper` so that the API is more obvious. Let me know if you'd like me to change it.

I'm happy to also update the documentation if you're willing to accept this change but its a bit of a blocker for us to be able to adopt RxDB at the moment and no combination of `modifier` and `dataPath` seems to be a valid workaround right now

pubkey:
Maybe I am missing something.
Does your GraphQL API really respond with an array that contains items which can be projected to more then one document?

WonderPanda:
> Maybe I am missing something. Does your GraphQL API really respond with an array that contains items which can be projected to more then one document?

Yes, the query I would like to use has nested documents that I would like to flatten into a single list like in the code sample in #3792. It is possible for me to restructure the query but the existing query patterns play nicer with the existing underlying database indices we have in place.

Regardless, I feel like the flexibility of exposing a transform function instead of (or in addition to) a data path is much better for consumers of RxDB since it provides them more control over extracting the records as well as the possibility to give better type annotations with Typescript.

`dataPath` today basically boils down to a function anyways, its just hidden: `() => object.get('path')`

The change I implemented in this PR allows people to keep using the string based `dataPath` if they wish or the function based approach if they would like the option for better type safety or more complex extraction like my particular use case

WonderPanda:
Okay I can make that change

pubkey:
Closing this.
PR for `responseModifier` is welcomed.

PR #3793, opened 2022-05-11
FIX dev-mode fail (patternProperties with regex)

moofoo:
## This PR contains:
 - A BUGFIX

Dev-mode fails when you use patternProperties as a property attribute with the expected regex
See: [JSON schema](https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.10.3.2)

This change narrowly affects the field name check for a patternProperties object when it's used in a schema to define a property

### DIFF
--- a/src/plugins/dev-mode/check-schema.ts
+++ b/src/plugins/dev-mode/check-schema.ts
@@ -3,550 +3,561 @@
  * to ensure nothing is broken or not supported
  */
 
-import objectPath from 'object-path';
-import {
-    newRxError
-} from '../../rx-error';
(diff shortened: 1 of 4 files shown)


src/plugins/dev-mode/check-schema.ts
```diff
@@ -3,550 +3,561 @@
  * to ensure nothing is broken or not supported
  */
 
-import objectPath from 'object-path';
```
pubkey, line 6: You changed the format of the whole file. This makes the PR unreadable.

moofoo: Can’t believe I overlooked that. My bad.

PR #3794, opened 2022-05-13
FIX dev-mode pattern properties fix

moofoo:
## This PR contains:
 - A BUGFIX

Field name check in dev-mode fails when regex is used in pattern properties attribute

### DIFF
--- a/test/helper/schema-objects.ts
+++ b/test/helper/schema-objects.ts
@@ -404,3 +404,35 @@ export function humanWithCompositePrimary(): HumanWithCompositePrimary {
         }
     };
 }
+
+export interface HumanWithPatternProperties {
+    passportId: string;
+    firstName: string;
+    lastName: string;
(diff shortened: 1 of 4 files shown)


test/helper/schema-objects.ts
```diff
@@ -404,3 +404,35 @@ export function humanWithCompositePrimary(): HumanWithCompositePrimary {
...
+    }
+}
+
+export function humanWithPatternProperties(): HumanWithPatternProperties {
+    return {
+        passportId: randomString(12),
+        firstName: faker.name.firstName(),
+        lastName: faker.name.lastName(),
+        numericallyIndexedJSONData: {
+            1:true,
```
pubkey, line 423: Shouldn't these be strings instead of numbers?

pubkey:
Closing this. Ping me when you want to continue working on that.

PR #3800, opened 2022-05-17
Removing heartbeat effect from homepage

mattgperry:
## This PR contains:
Removed heartbeat effect (animations and sound) from homepage

The heartbeat effect is quite stressful. It distracts from the messaging of the library, which otherwise looks great!

### DIFF
--- a/docs-src/index.html
+++ b/docs-src/index.html
@@ -93,7 +93,6 @@
                     <img
                         src="./files/logo/logo_text.svg"
                         width="290"
-                        class="beating"
                         id="heartbeat-logo"
                         alt="RxDB"
                     />
@@ -121,7 +120,7 @@
                     <div class="left third centered">
                         <img
                             src="./files/logo/logo.svg"
-                            class="beating logo"
+                            class="logo"
                             alt="RxDB"
                         />
                     </div>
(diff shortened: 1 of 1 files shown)


robertpenner:
Thank you; I was going to submit the same PR.

docs-src/index.html
```diff
@@ -327,7 +326,7 @@ <h2>Online <b class="underline">is optional</b></h2>
                 Next.js
             </div>
             <div
-                class="neumorphism-circle-m circle centered beating-second"
+                class="neumorphism-circle-m circle centeredd"
```
keichinger, line 329: There's a small typo here.

```suggestion
                class="neumorphism-circle-m circle centered"
```

keichinger (approved):

pubkey:
Hi.
Sorry for the late answer.
The beating and sound is now disabled by default ac62012691345cb75fe2fbcf358cfe761541de5f

pubkey:
Ok I reverted this now.
I know that playing sound is annoying and people might not like the blinking effect etc.
The problem is that just removing that leave the current landingpage in a state that just looks broken and makes no sense at all.
In the future there might be a full redesign with a page that works without that, but for now I rather have an annoying page then a broken one.

pubkey:
I now added a check to not show the animations to users with `prefers-reduced-motion` set.
This is a temporary fix until we have a new reworked landingpage.
https://github.com/pubkey/rxdb/commit/616928828d829b1925ed29f586e5c2307e716408

PR #3763, opened 2022-04-15
UPDATE pouchdb 7.3.0

pubkey:
## Todos 
- [ ] Tests
- [ ] Documentation
- [ ] Typings
- [ ] Changelog

PR #3807, opened 2022-05-21
upgrade pouchdb to 7.3.0

cetsupport:
## This PR contains:

 - IMPROVED TESTS

CI/CD test fail when upgrade to pouchDB 7.3.0

Fix the issue upgrade to pouchdb 7.3.0


pubkey:
So was [this problem](https://github.com/pubkey/rxdb/pull/3763#issuecomment-1110482174) already fixed or did the master branch get changed so that we do not detect the error?

cetsupport:
This problem was fixed by this PR. The reason is changed the selector from $gt to $ne to avoid race condition in pouchdb.

pubkey:
Ah, I see.
Please ping me in gitter so I can send you the license agreement for the premium plugins.

pubkey:
I also thought about this before I merged the PR.
I think having a race condition on pouchdb 7.3.0 is still better then using 7.2.2.

Many other things are broken in 7.2.2 where queries do not work or `npm install` fails.

Having a query withi `$gt: null` is something that should not happen in production because it is not clear if a string/number type can be greater-then compared with `null`. In the future I do not want to allow mixed types in RxDB, see [here](https://github.com/pubkey/rxdb/blob/master/orga/before-next-major.md#do-not-allow-type-mixing)

Yes we should make a PR with a failing test in the pouchdb repo. But in the past this was not worth the time because pouchdb uses the stale bot which just closes them again.
Check my already opened and ignored issues: https://github.com/pouchdb/pouchdb/issues?q=is%3Aissue+pubkey
So it would be cool if someone else could do that.

PaulMest:
Ahh... thanks for the context. I didn't realize there were known issues and warts with staying on 7.2.2. I guess 7.3.0 with this race condition is the lesser of two evils? I'm leaving on vacation soon and have a million things to do, but I'll see if I can create a repro and post it as an open PR.

pubkey:
> I guess 7.3.0 with this race condition is the lesser of two evils

Yes, at least I think so.
Would be cool if you find the time to report this at PouchDB.

PaulMest:
Ok, here is the repro case #3813. I'll also report this over in PouchDB.

PaulMest:
FYI, PouchDB accepted the fix proposed and has released it as part of [PouchDB 8.0.0](https://github.com/pouchdb/pouchdb/releases/tag/8.0.0).

pubkey:
Yes I know. But the pouchdb storage is deprecated anyways, so I will not update PouchDB.

meabed:
I think most of react native setup uses pouch db, any change to support pouchdb if it worth it.

PR #3813, opened 2022-05-25
Failing test for PouchDB 7.3.0 race condition

PaulMest:
## This PR contains:
 - IMPROVED TESTS

PouchDB 7.3.0 introduced a race condition where two databases open in memory at the same time could cause hanging conditions. This was initially discovered randomly by 2 tests in the test suite. This PR adds a smaller consolidated view of this race condition.

Simple repro:
```
$ npm install
$ npm run test:fast:pouchdb
```

## Todos
- [ ] Changelog
- [ ] Actual fix

### DIFF
--- a/test/unit/pouch-db-integration.test.ts
+++ b/test/unit/pouch-db-integration.test.ts
@@ -299,6 +299,79 @@ config.parallel('pouch-db-integration.test.js', () => {
         });
     });
     describe('BUGS: pouchdb', () => {
+        it('Race condition initially discovered with PouchDB in-memory-adapter 7.3.0', async () => {
+            const func1 = async() => {
+                const pouch1: PouchDBInstance = new PouchDB('func1db', {
+                    adapter: 'memory'
+                });
+                const docId = 'func1doc1';
+
+                // insert
+                await pouch1.bulkDocs({
+                    docs: [{
+                        _id: docId,
+                        value: 1,
+                        _rev: '1-51b2fae5721cc4d3cf7392f19e6cc118'
+                    }]
+                }, {
+                    new_edits: false
+                });
+
+                // update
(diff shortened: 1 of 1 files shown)


cetsupport:
@PaulMest Looks like there have race condition when close one of multiple database.

When execute db.close, will call pouchdb-adapter-leveldb-core/lib/index.js:
From line 1218:

```javascript
  api._close = function (callback) {
    /* istanbul ignore if */
    if (db.isClosed()) {
      return callback(pouchdbErrors.createError(pouchdbErrors.NOT_OPEN));
    }
    db.close(function (err) {
      /* istanbul ignore if */
      if (err) {
        callback(err);
      } else {
        dbStore.delete(name);

        var adapterName = pouchdbUtils.functionName(leveldown);
        var adapterStore = dbStores.get(adapterName);
        var keys = [...adapterStore.keys()].filter(k => k.includes("-mrview-"));
        keys.forEach(key => {
          var eventEmitter = adapterStore.get(key);
          eventEmitter.removeAllListeners();
          eventEmitter.close();
          adapterStore.delete(key);
        });

        callback();
      }
    });
  };
```

The problem happen in below line:
```javascript
        var keys = [...adapterStore.keys()].filter(k => k.includes("-mrview-"));
```
So when we close first database, actually will select the second database views... and disconnect it. This cause problem I think.

pubkey:
I am closing this issue here because there is nothing that can be done on the RxDB side. I will reopen when it is fixed on the pouchdb side.
Related links:
https://github.com/pouchdb/pouchdb/pull/8515
https://github.com/pouchdb/pouchdb/pull/8513

PaulMest:
https://github.com/pouchdb/pouchdb/pull/8513 has been merged. The fix should be in the next version of PouchDB.

PR #3832, opened 2022-06-02
Fix examples electron and electron-remote

hujiulong:
## This PR contains:
1. Upgrade electron to the latest stable version(v19) and refactor old codes.
2. Replace spectron with [playwright](https://playwright.dev/), because spectron is officially deprecated.
3. Refactor electron-remote example with [@electron/remote](https://github.com/electron/remote).

* Cannot run electron examples
* Electron examples test failed


pubkey:
Can you comment in the CI tests for electron?
https://github.com/pubkey/rxdb/blob/master/.github/workflows/main.yml#L395
and
https://github.com/pubkey/rxdb/blob/master/.github/workflows/main.yml#L407

They have been commented out in the past because the CI did not work with the electron version.

hujiulong:
> Can you comment in the CI tests for electron? https://github.com/pubkey/rxdb/blob/master/.github/workflows/main.yml#L395 and https://github.com/pubkey/rxdb/blob/master/.github/workflows/main.yml#L407
> 
> They have been commented out in the past because the CI did not work with the electron version.

Ok, I'll try to fix it.

examples/electron-remote/test/render.test.js
```diff
@@ -17,7 +24,8 @@ module.exports = (function () {
         // issue #587 Icorrect working attachments in electron-render
         await (async function () {
             const db = await createRxDatabase({
-                name: 'foobar587' + new Date().getTime(),
+                // generate simple random ID to avoid conflicts when running tests at the same time
+                name: 'foobar587' + Math.round(Math.random() * 0xffffff).toString(16),
```
hujiulong, line 28: @pubkey I can't reproduce the problem, I guess it's because the current time is used **here↑** to generate the database name, so running render.test.js multiple times at the same time will cause errors. I changed it to a random ID to avoid this error.

pubkey:
Hm I think this randomly stucks https://github.com/pubkey/rxdb/runs/6708937126?check_suite_focus=true

hujiulong:
I made some updates:
1. Upgrade electron to the latest stable version(v19) and refactor old codes.
2. Replace spectron with [playwright](https://playwright.dev/), because spectron is officially deprecated.
3. Remove electron-remote example, because electron remote module was removed in electron 14.

But there is a bug here. I haven't found the reason, @pubkey  do you have any suggestions?
https://github.com/pubkey/rxdb/blob/6ba2e523262f53ac7a5dd30d19de454d52487536/examples/electron/main.js#L66-L69

examples/electron-remote/.npmrc
```diff
@@ -1 +0,0 @@
-package-lock=false
```
pubkey, line 1: We do not use package-locks on purpose.
When installing the package on a fresh install, it must work or our CI should fail so that we can investigate.

pubkey:
> The remote module was deprecated in Electron 12, and will be removed in Electron 14. It is replaced by the [@electron/remote](https://github.com/electron/remote) module.

Shouldn't we switch to that one instead of removing the electron-remote example?

hujiulong:
> > The remote module was deprecated in Electron 12, and will be removed in Electron 14. It is replaced by the [@electron/remote](https://github.com/electron/remote) module.
> 
> Shouldn't we switch to that one instead of removing the electron-remote example?

Ok, I refactor the electron-remote example to use @electron/module, this PR is ready to be merged.

pubkey:
Great, merged. Can you contact me in gitter for your access to the premium plugins?

PR #3852, opened 2022-06-16
FIX(docs) change field value for lastId

maxnowack:
I'm pretty sure, that it has to be the id of the document instead of the name 🙂


pubkey:
Thanks, merged.

PR #3856, opened 2022-06-18
Show atomicUpsert bug with graphql

AntonOfTheWoods:
## This PR contains:
 - IMPROVED TESTS

Commit https://github.com/pubkey/rxdb/commit/9f93496da35f6171017b3da50a10277f582e31d5 introduced a bug whereby a test that all elements of the `_meta` dict are *already* present at the time of the test. Investigation into the state in the actual objects appears to suggest that is not the case at the time of the test but *is* the case later in the database.

Sorry this is not in the `bug-report.test.ts`. Everything required was available in the previous test so in order to keep things manageable I just copy-pasted.

### DIFF
--- a/test/unit/replication-graphql.test.ts
+++ b/test/unit/replication-graphql.test.ts
@@ -2141,6 +2141,79 @@ describe('replication-graphql.test.ts', () => {
                 server.close();
                 db.destroy();
             });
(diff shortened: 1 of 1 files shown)


pubkey:
I fixed this. `RxDocument._saveData()` does not merge the _meta fields instead of replacing them.

PR #3861, opened 2022-06-20
test to recreate issue with `find` query containing "$or" and "$regex"

pete-hotchkiss:
## This PR contains:
Bug with querying database collections

Since upgrading to `12.5.3` from `11.6.0` - when trying to query a collection using both `$or` and `$regex` elements something is barfing meaning code will exit with an error 

`TypeError: userValue.every is not a function`

I've also created a standalone repos that shows this error that if run with the pouchdb deb mode on outputs the following to the console

image

Repos URL if you want it - https://github.com/pete-hotchkiss/rxdb-or-regex-example

### DIFF
--- a/test/unit/bug-report.test.ts
+++ b/test/unit/bug-report.test.ts
@@ -46,17 +46,22 @@ describe('bug-report.test.js', () => {
                     primary: true,
                 },
                 firstName: {
-                    type: 'string'
+                    type: 'string',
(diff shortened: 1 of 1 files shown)


pete-hotchkiss:
indeed - not clear when the fix thats awaiting review is likely to make it into a release. I'll monitor in the mean time.

PR #3872, opened 2022-06-29
bug on vanilla example

fredguth:
## This PR contains:
 - A BUG report 
examples/vanilla stopped working on commit 93e174bbab9515e8f8fe3ed21c58d18cb2095314 (12.0.0-beta.2).  You can check in the previous commit it was working.

> git checkout 12.0.0-beta.2
> cd /examples/vanilla
> npm i
> npm start
===>>> error: Uncaught TypeError: RxDB.getRxStoragePouch is not a function at main.js:35:23

> git checkout HEAD~1
Previous HEAD position was 93e174bb 12.0.0-beta.2
HEAD is now at 3fd2a9e2 FIX docs

> npm start
==> works fine!

### DIFF
--- a/examples/vanilla/main.js
+++ b/examples/vanilla/main.js
@@ -32,6 +32,7 @@ const syncURL = 'http://' + window.location.hostname + ':10102/';
 window.RxDB
     .createRxDatabase({
         name: 'heroesdb',
+        // some change on commit 93e174bbab9515e8f8fe3ed21c58d18cb2095314 made getRxStoragePouch brake
         storage: RxDB.getRxStoragePouch('idb'),
         password: 'myLongAndStupidPassword'
     })


fredguth:
My apologies.  But please understand that the reason I was looking the examples/vanilla was exactly to understand how to import the methods.   I guessed there was a change in the way it was made and took a time to find out which commit presented the change.  I did not want you to waste time, just the opposite.

pubkey:
Yes I see.
The thing is that the vanilla example is not tested in the CI so it has run out of date.
This will happen again and again until someone adds tests to the vanilla example (like the other examples have) and then these tests run in the ci.

When you need at the exact code point is

```ts

import {
  getRxStoragePouch
} from 'rxdb/plugins/pouchdb';
 
/* ... */
storage: getRxStoragePouch('idb')
```

PR #3903, opened 2022-07-22
Fix react-native example based expo

msotnikov:
### This PR contains
Updated example for react-native based on Expo.

Main changes at 01.09.2022:
1) Updated versions expo@46.0.9, pouchdb@7.3.0, react@18.2.0, react-native@0.69.2
2) Fixed sync with couchdb instanance (you need your own)
3) Some refactoring


pubkey:
A main reason for having these example projects, is to know when a change on the rxdb code breaks compatibility.
So I do not want to install rxdb from npm but it should use the build rxdb bundle. Also having a fixed version there would mean that someone has to come and update it on each rxdb release.

msotnikov:
Ok, changed back. Added comment in readme.md.

pubkey:
Can you update this line https://github.com/pubkey/rxdb/blob/master/.github/workflows/main.yml#L450
So that it does use the node version from the `node-version-file: '.nvmrc'` like in the other ci tasks.

msotnikov:
expo-cli doesn't support node 18. https://github.com/expo/expo-cli/issues/4478

pubkey:
I see.
I am not sure if we should merge this then. Because we increase the expo version but decrease the react-native version, which does not lead to any benefit.

msotnikov:
> I see. I am not sure if we should merge this then. Because we increase the expo version but decrease the react-native version, which does not lead to any benefit.

But current code not working any case :-)

Now updated only expo-cli. I leave old expo package and have recommendations https://docs.expo.dev/versions/latest/#each-expo-sdk-version-depends-on-a

I'll try to update all deps

msotnikov:
With jest-expo >= 44 tests passes, but finishes with seg fauilt (yarn and npm same). Now its works.

pubkey:
Replication does not work because you switched from the pouchdb RxStorage to the memory RxStorage.
But the replication code uses the CouchDB replication protocol which only works with PouchDB RxStorage.

examples/react-native/package.json
```diff
@@ -1,36 +1,36 @@
 {
   "name": "rxdb-example-react-native",
-  "version": "0.1.0",
+  "version": "0.2.0",
   "private": true,
   "scripts": {
-    "preinstall": "(cd ../../ && npx yarn@1.13.0 pack ../../ --filename ./examples/react-native/rxdb-local.tgz)",
```
pubkey, line 6: Is there a reason why you removed the pinned version?
One day yarn will publish its version 2.x.x to npm and then this command will break here when the version is not pinned.

msotnikov: If yarn 2.p published, we could update code here.

pubkey: Please pin all dependency versions to make the CI reproducible.

pubkey:
Is there a reason why you switched to the memory adapter?
Because PouchDB is  the only persistent RxStorage for react native, so the whole reason for the react-native example is  to explain how to store data in it with the PouchDB RxStorage.

msotnikov:
> Is there a reason why you switched to the memory adapter? Because PouchDB is the only persistent RxStorage for react native, so the whole reason for the react-native example is to explain how to store data in it with the PouchDB RxStorage.

The entire PouchDb ecosystem is very outdated. And so on. The maintainers think it's okay, but it's not. 
For example, pouchdb-adapter-asyncstorage uses the deprecated async-storage (moved to community package). I had patched that package and fix that - the persistance works, but it doesn't fix sync. I tried to find working packages, but only the sqlite package with a bunch of crutches worked.

So, for this PR, we should deside what requirements are needed:
- persistence?
- sync throught http?
- ... ?

pubkey:
Yes both is needed. The reason for making the react native example was to show that RxDB can work with the react-native SQLite  storage.
We can switch from couchdb-sync to the rxdb replication of course.

I assume this PR is made in context of the premium tasks, is that correct?

msotnikov:
> Yes both is needed. The reason for making the react native example was to show that RxDB can work with the react-native SQLite storage. We can switch from couchdb-sync to the rxdb replication of course.
> 
> I assume this PR is made in context of the premium tasks, is that correct?

In my project I have some success with custom version of `pouchdb-adapter-react-native-sqlite`. But yes, I want to try your version of sqlite connector because, as I mentoin before, the whole pouchdb ecosystem is outdated.

I'll try to change the adapter to sqlite.

msotnikov:
Recent changes:
1) Dependencies updated to current versions - expo@46.0.9, pouchdb@7.3.0, react@18.2.0, react-native@0.69.2
2) Sync fixed and tested with couchdb@3.2.2 instance (not included). Also, tested with pouchdb-server@4.2.2 - doesn't work.

There is no persistance functionality. Pouchdb need a crunches and Sqlite need a native modules to work. I don't want to deal with it for Expo. My suggestion is - rename this module to react-native-expo and leave as is in this PR. And create a new one with bare react-native.

pubkey:
Strange error, I restarted the failed CI test.

msotnikov:
I removed the cache settings. In any case, there is no package-lock file. And if I understood you correctly - you want to always have fresh packages.

The buggy jest and CI. I don't know how to fix it. Lost my patience. I'll be back later.

msotnikov:
So... Now we will test database.

.github/workflows/main.yml
```diff
@@ -484,28 +484,14 @@ jobs:
           run: npm run test:build
 
   react-native:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
pubkey, line 487: Please pin all dependencies, always. This goes for the npm modules and also for docker image versions etc.

msotnikov: fixed

pubkey:
The code looks good to me.
What is the state of the sync now? I see it connects to 'http://admin:mysecret1@localhost:5984' but there is nothing started at this port.
As far as I remember the pouchdb sync did work before, or was it broken anyway?

msotnikov:
> The code looks good to me. What is the state of the sync now? I see it connects to 'http://admin:mysecret1@localhost:5984' but there is nothing started at this port. As far as I remember the pouchdb sync did work before, or was it broken anyway?

Before there was the similar code.
``` const syncURL = 'http://localhost:10102/'; // Replace localhost with a public ip address! ```
I haven't seen this code works. I haven't seen any server side also.

pubkey:
Ah, you are right, sorry havent seen that there never was a server.
So I can merge this now?

msotnikov:
Counting for it :-)

pubkey:
Merged, thank you for your work.
You can ping me in discord for the access to the premium plugins.

PR #3911, opened 2022-07-26
add `age` to heroSchemaLiteral

dharmaturtle:
## This PR contains:
 - IMPROVED DOCS

`heroSchema` is inconsistent without this.

## Todos 
- [x] Documentation

### DIFF
--- a/docs-src/tutorials/typescript.md
+++ b/docs-src/tutorials/typescript.md
@@ -55,6 +55,9 @@ export const heroSchemaLiteral = {
         },
         lastName: {
             type: 'string'
+        },
+        age: {
+            type: 'integer'
         }
     },
     required: ['firstName', 'lastName', 'passportId'],


pubkey:
Can you add the same to the file `test/tutorials/src/typescript.ts` so that we can be sure that it works.

dharmaturtle:
From a certain perspective, [it's already there](https://github.com/pubkey/rxdb/blob/master/test/tutorials/src/typescript.ts#L78-L80). From a _strict_ perspective, it isn't because I added it to `heroSchemaLiteral` and `test/tutorials/src/typescript.ts
test/tutorials/src/typescript.ts` doesn't have a `heroSchemaLiteral`. 

On a related note, Options A, B, and C on https://rxdb.info/tutorials/typescript.html are unnecessary because of [this](https://github.com/pubkey/rxdb/blob/035b86361ae7473cb45abe9aae288b5e78ddd192/docs-src/tutorials/typescript.md?plain=1#L147-L169). I'd remove the latter, but decided against it in the interest of simplicity and `test/tutorials/src/typescript.ts`.

PR #3929, opened 2022-08-09
bug in query results type

chrisdrackett:
I'm pretty sure this type is incorrect. It should be an array of documents.

## This PR contains:

 - IMPROVED typings


pubkey:
You are right. Thanks, merged.

PR #3943, opened 2022-08-20
add undefined to typing of RxDocumentData._deleted

Whoops:

The type of the `_deleted` field on `RxDocumentData<T>` is boolean, however it's observed to also be undefined. When replicated, this is the difference between the deleted flag being sent to the server and omitted. It also means eslint thinks `doc._deleted ?? false` is unnecessary since it incorrectly believes `_deleted` cannot be nullish.

### DIFF
--- a/src/types/rx-storage.d.ts
+++ b/src/types/rx-storage.d.ts
@@ -16,7 +16,7 @@ export type RxDocumentData<T> = T & {
      * Instead the documents are stored with _deleted: true
      * which means they will not be returned at queries.
      */
-    _deleted: boolean;
+    _deleted: boolean | undefined;
 
     /**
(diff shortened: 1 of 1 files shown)


pubkey:
`RxDocumentData._deleted` must be required. Otherwise we had strange bugs in some storages because they index `undefined` different then `false`.

Whoops:
Yuck, OK I see my mistake. I assumed the push query builder was getting an `RxDocumentData<T>`. I have no idea why I assumed that, it's not documented that way anywhere and the actual typing just specifies `docs` as `any`. I have `deletedFlag: deleted`, so RxDB is giving me the document with `_deleted` already renamed to `deleted` and ready to be sent to the server. My bad, should've spent a bit more time debugging this.

PR #3952, opened 2022-08-26
Support JsonSchema for nested additionalProperties

swnf:
## This PR contains:

Fixed types for `additionalProperties` with a corresponding tests

Unable to use a json schema for nested `additionalProperties`


swnf:
By the way, I think at least nodejs 18 is required to directly run the tests because they use `Blob`. You might want to update [contribute.md](https://github.com/pubkey/rxdb/blob/master/docs-src/contribute.md).

PR #3957, opened 2022-08-30
Fix angular example

jeromepochat:
This fix 2 issues that were making angular example not working:
* missing `maxLength` in schema
* wrong object passed to `addRxPlugin` function


pubkey:
Thank you very much, merged!

PR #3967, opened 2022-09-04
Upgrade TypeScript to 4.8.x

### LINKED ISSUE #3964: Loose typing of RxSchema property type
Hey,
in addition to the types defined in the link above [rx-schema.d.ts#L7](https://github.com/pubkey/rxdb/blob/6766596dfaf3f90a8de0e4d34a4268c699e0a3d4/src/types/rx-schema.d.ts#L7) the line adds `| string`, which makes the typing very loose and VSCode doesn't suggest me the specific types. 
Does this looseness serve any purpose?

Current behavior:

My desired behavior (without `| string`):

Zhincore:
## This PR contains:
 - typings

Closes #3964

### DIFF
--- a/package.json
+++ b/package.json
@@ -270,7 +270,7 @@
     "terser-webpack-plugin": "5.3.6",
     "ts-loader": "9.3.1",
     "ts-node": "10.9.1",
-    "typescript": "4.5.5",
+    "typescript": "4.8.2",
     "walk-sync": "3.0.0",
     "watch": "1.0.2",
     "webpack": "5.74.0",
--- a/src/plugins/local-documents/local-documents.ts
+++ b/src/plugins/local-documents/local-documents.ts
@@ -32,7 +32,7 @@ import { getSingleDocument, writeSingle } from '../../rx-storage-helper';
  * save the local-document-data
  * throws if already exists
  */
-export async function insertLocal<DocData = any>(
+export async function insertLocal<DocData extends Record<string, any> = any>(
     this: RxDatabase | RxCollection,
     id: string,
     data: DocData
@@ -67,7 +67,7 @@ export async function insertLocal<DocData = any>(
  * save the local-document-data
  * overwrites existing if exists
  */
-export function upsertLocal<DocData = any>(
+export function upsertLocal<DocData extends Record<string,any> = any>(
     this: any,
     id: string,
     data: DocData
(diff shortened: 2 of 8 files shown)


Zhincore:
All tests pass on my machine. I seem to be unable to reproduce the failing environment.

Zhincore:
Updating TypeScript only makes `test/typings.test.ts` pass, tests that do not have inline code still fail. 
Those can mostly be fixed by following RxDB docs, more specifically:
> It is important to set 'as const' to preserve the literal type
Which means adding `as const` after every schema definition.

Zhincore:
Even though it's stated in docs, I understand that this is potentially a breaking change, so I wouldn't mind if this PR got denied.

pubkey:
I do not fully understand these changes. Does that mean that everybody needs to add `as const` to a all schemas even if the document type is not aggregated from the schema?

Zhincore:
Yes, I assume so, unfortunately...

Zhincore:
I just realized I changed this PR into upgrading Typescript, which can be done with `| string` intact. 
I'm sorry for the mayhem I'm doing here.

Zhincore:
Actually, maybe something from here could be implemented: https://github.com/microsoft/TypeScript/issues/29729

Zhincore:
Changing `| string` to `| (string & {})` works great.

Example: 

Now I get suggestions, but the "looseness" is still there because an empty string isn't underlined as error (it was without `| string`).

Should I revert to the commit before changing all the tests and push with `| (string & {})` or do you prefer the explicit `RxJsonSchema` as it is in my latest commit?

Zhincore:
> I am not sure what is meant with 'explicit RxJsonSchema'.

By that I mean directly writing the type in code.

In my latest commit these two work:
```ts
const heroSchema = { /* ... */ } as const;
```
```ts
const heroSchema: RxJsonSchema<any> = { /* ... */ };
```
(The later is "explicit RxJsonSchema")

Which means it forces users to add stuff to all of their schemas.
So I suppose I'll push the new version that doesn't require any action from users but still gives them the suggestions.

Zhincore:
So now this PR:
- Upgrades TypeScript to newer version and fixes type errors that came with it
- Allow Typescript to suggest values for `type` of schema properties

pubkey:
Thank you very much for your great work.
Merged!

PR #4005, opened 2022-10-02
Array order test case for sortObject

swnf:
## This PR contains:
A test case failing in Firefox but not Chrome

I'm working on a replication library that compares the client's RxSchema hash with the one on the server side. I realized the hash provided by RxDB is not the same across browsers (Chrome, Firefox) if the schema contains an array of objects (an `anyOf` in my case). This could also be a problem in the future if a browser update changes the browser's implementation of the sort algorithm.

## Cause of the issue

src/util.ts has the following lines for sorting arrays:
```ts
.sort((a, b) => {
  if (typeof a === 'string' && typeof b === 'string')
    return a.localeCompare(b);

  if (typeof a === 'object') return 1;
  else return -1;
})
```

It returns `1` if the left operand is an object and `-1` otherwise. The spec has no constraints for the order of `a` and `b` in the compare function. It seems to be different in Chrome and Firefox.

## Possible solutions

- Return `0` if one operand is not a string (`sort` is stable, it will preserve the order in this case)
- Don't sort arrays at all

Both solutions would be a breaking change.

### DIFF
--- a/test/unit/util.test.ts
+++ b/test/unit/util.test.ts
@@ -139,6 +139,11 @@ describe('util.test.js', () => {
             const sorted = sortObject(obj);
             assert.ok(sorted.color.$regex instanceof RegExp);
         });
+        it('should sort object arrays in a deterministic order', () => {
+            const arr = [{a: 1}, {b: 2}];
+            const sorted = sortObject(arr);
+            assert.deepStrictEqual(sorted, [{a: 1}, {b: 2}]);
+        });
     });
     describe('.validateDatabaseName()', () => {
         describe('positive', () => {


pubkey:
Thank you for detecting this.
I think we should not sort arrays for schema normalization because that could change the "meaning" of a schema.
The current behavior is dangerous because a schema with a different index (just another order) could result in the same hash because of the normalization.

At the moment RxDB uses these hashes only for the developers security to not accidentially use an existing data store with a different schema without having done a migration. So it is not that big of a problem that the different browsers result in a different hash as long as they create the same hash each time for themself.

However this should be fixed in the next major version, so we should add it to the todo list: https://github.com/pubkey/rxdb/blob/master/orga/before-next-major.md

For your replication plugin you might want to craft your own schema hashes in the meantime.

swnf:
I agree. If you make sure that `sortObject` recurses into array elements (without sorting the array) the schema hash should be the same in all browsers. Feel free to close this PR until the next major release.

pubkey:
I added this to the todo list.

PR #4011, opened 2022-10-05
may return undefined if `array`  is empty

AlexErrant:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array#access_an_array_item_by_its_index


AlexErrant:
Apologies for the back and forth - clearly I can't figure out how to run tests locally :|

pubkey:
No problem. I mostly use the ensureNotFalsy() method for the tuntime checks, maybe that will make it easier to read.

PR #4058, opened 2022-10-13
Ignored SafariTechPreview in test

Albert-Gao:
To unblock test locally


pubkey:
What does this change and how does it solve your problem?
Removing that string from the list will enable the test runs via `SafariTechPreview` browsers, it will not disable them.

PR #4059, opened 2022-10-14
fixed bug in traverse() - TypeError: Cannot read properties of null (reading hasOwnProperty)

### LINKED ISSUE #4055: rxdb addCollections err TypeError: Cannot read properties of null (reading 'hasOwnProperty')
## for the following schema:

```typescript
import { RxJsonSchema } from "rxdb";
import { UserDocType } from "./types";

export const userSchema: RxJsonSchema<UserDocType> = {
  title: "user schema",
  description: "describes a user",
  type: "object",
  version: 0,

  // re-enable once not use dexie.js engine for the web since it does not support it
  keyCompression: false,

  primaryKey: "id",
  properties: {
    id: {
      type: "string",
      maxLength: 100,
    },
    birthday: {
      type: "string",
      default: null,
    },
    timezone: {
      type: "string",
      default: null,
    },
    createTime: {
      type: "string",
    },
    updateTime: {
      type: "string",
    },
  },
  required: ["id", "createTime", "updateTime"],
};
```

I get an error when creating the database, remove the usage of devplugin "solves" it...

wonder if I am missing something or a bug.

## a reproducible repo here:

https://github.com/Albert-Gao/my-rxdb-app

## Digging
With some debug, I found the issue here in this function,
image

You can see the runtime value for `schemaObj` is `"string"` which is not an object, thus causes the problem, do not know the cause though.

Albert-Gao:
## This PR contains:
fixed #4055

`TypeError: Cannot read properties of null (reading hasOwnProperty)`


Albert-Gao (commented):
need to add the fix later, just need to confirm it's failing on the remote since still have trouble running tests locally

Albert-Gao:
in addition to the 'string' will be passed down to the `checkSchema()` as the 2nd parameter which triggered the bug, `null` is also being passed, and here comes the dark side of js:

image

😂

It obviously bypasses the original check.

pubkey:
Code looks good.
But we should move the test here: https://github.com/pubkey/rxdb/blob/master/test/unit/rx-schema.test.ts#L55
Also do we need to create a database to reproduce, or is it enough to just call `checkSchema()`?

Albert-Gao:
thanks, mate.

Yes, I indeed created a database :)

I moved the test to the location you mentioned :)

PR #4067, opened 2022-10-19
TYPOFIX: Fix a typo.

nirvdrum:


nirvdrum:
Is this pulled from a dependency? I grepped through the project and couldn't find any source that would correspond to the generated file. I had assumed this one just lived within the project, since it's committed and seemingly has no source. But, I can close this out if it should be fixed elsewhere.

pubkey:
> You can fix stuff in the documentation AND in the codebase AND in the tests (docs-src, src and test folders).
Event typos in code comments do count.
Only fixes in generated files (everything from the dist and docs folders) do NOT count.

PR #4088, opened 2022-10-21
Stop retries

Whoops:
## This PR contains:
* A bugfix

At this time it doesn't appear to be possible to stop RxDB from retrying requests after a non-recoverable error (ex 401). A simple example:
```typescript
const state = db.test.syncGraphQL({
    url: { http: "/badurl" },
    pull: {
      queryBuilder: pullQueryBuilderFromRxSchema("test", {
        schema: schema,
        checkpointFields: ["id", "updatedAt"],
      }),
    },
    live: false,
  });

state.error$.subscribe((err) => {
    console.log("error", err);
    void state
      .cancel()
      .catch((err) => console.log("cancel error", err))
      .then(() => console.log("cancel done"));
  });
```
"Cancel done" is logged to the console however a new request is still sent every 5 seconds.

### DIFF
--- a/test/unit/replication-graphql.test.ts
+++ b/test/unit/replication-graphql.test.ts
@@ -576,7 +576,8 @@ describe('replication-graphql.test.ts', () => {
                     replicationState.cancel()
                 });
 
-                const timeout = new Promise((resolve, _) => setTimeout(resolve, 500, 'timeout'));
+
+                const timeout = wait(500).then(() => 'timeout');
 
                 assert.notStrictEqual(await Promise.race([replicationState.awaitInitialReplication(), timeout]), 'timeout',)
 
@@ -930,7 +931,7 @@ describe('replication-graphql.test.ts', () => {
                     replicationState.cancel()
                 });
 
-                const timeout = new Promise((resolve, _) => setTimeout(resolve, 500, 'timeout'));
+                const timeout = wait(500).then(() => 'timeout');
 
                 assert.notStrictEqual(await Promise.race([replicationState.awaitInitialReplication(), timeout]), 'timeout',)
 
@@ -962,7 +963,7 @@ describe('replication-graphql.test.ts', () => {
                     replicationState.cancel()
                 });
 
-                const timeout = new Promise((resolve, _) => setTimeout(resolve, 500, 'timeout'));
+                const timeout = wait(500).then(() => 'timeout');
 
                 assert.notStrictEqual(await Promise.race([replicationState.awaitInitialReplication(), timeout]), 'timeout',)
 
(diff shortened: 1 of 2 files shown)


meabed:
Thanks for fixing this, I was just about to open an issue for it.

After canceling the replication requests being sent from the initial replication start.

test/unit/replication-graphql.test.ts
```diff
@@ -549,6 +549,41 @@ describe('replication-graphql.test.ts', () => {
...
+                    retryTime: 100
+                });
+
+                await replicationState.error$.pipe(
+                    first()
+                ).toPromise().then(() => {
+                    replicationState.cancel()
+                });
+
+                const timeout = new Promise((resolve, _) => setTimeout(resolve, 500, 'timeout'));
```
pubkey, line 579: Please use the `wait()` method from the `async-test-utils` to be consistend with the rest of the code.

pubkey:
@meabed 
> A Question based on the implementation, I saw you return empty array on push, will this cause that record to be success? and will be not be tried when you start replication again?

Yes, returning an empty array means that the backend had no conflicts and all writes suceed.
We should add a test to verify that the replication will still work correctly when being restarted.

Whoops:
I added this test around resuming when cancelled and it passes? Which is surprising to me. It seems like everyone is right and the implementation is broken, and would result in RxDB thinking the rows have been synced  even though the backend didn't receive them, so I have to assume the problem is with the test, but I'm not sure how.

pubkey:
I think this is the case because there are other `this.isStopped()` checks afterwards so that the logic chain stops which leads RxDB to throw away the response.

Whoops:
> I think this is the case because there are other `this.isStopped()` checks afterwards so that the logic chain stops which leads RxDB to throw away the response.

Possibly here? https://github.com/pubkey/rxdb/blob/master/src/replication-protocol/upstream.ts#L79. It seems like `processTasks()` is what's responsible for actually settling the checkpoints(?), but since events are unsubscribed from when the canceled event is received it'll never get called so the checkpoint will never get persisted?

pubkey:
Yes that makes sense. So I would merge this now is that ok?

Whoops:
I think so!

meabed:
Awesome thank you!
I will test it in the release and let you know if it works as expected

pubkey:
I think there was something wrong in the test.
The following code from the test
```ts
assert.notStrictEqual(await Promise.race([replicationState.awaitInitialReplication(), timeout]), 'timeout',)
```
This assumes that `awaitInitialReplication()` fires before the timeout fires. But this is not correct because `awaitInitialReplication()` should never fire on an erroring replication because the "initial replication" never can finish.

PR #4094, opened 2022-10-28
Typo in CRDT Plugin: RxDDcrdtPlugin

jwallet:
## This PR contains:

 - A BREAKING CHANGE, a plugin typo

All plugin have RxDB as a prefix, CRDT plugin did not have it. Also, if you want to merge it, it will probably better to fix the lowercase crtd to all uppercase since its an abbreviation.


pubkey:
Thank you, merged.
Having all uppercase is a problem with the generated error message when the plugin is missing.

PR #4096, opened 2022-10-30
WIP RxQuery.count()

pubkey:
## Todos 
- [ ] Tests
- [ ] Documentation
- [ ] Typings
- [ ] Changelog


kicohen2:
Just an FYI. This should have been included in a major version release. When upgrading from an earlier version of 13 to 13.17, our application won't open anything from RXDB since we have implemented our own count method on almost all of our collections. Since count was added it is now a protected keyword and the application throws on the ORM checks. 

```
rx-error.ts:107 Uncaught (in promise) RxError (COL17): RxError (COL17):
RxCollection.ORM: statics-name not allowed
Given parameters: {
name:"count"}
    at newRxError (rx-error.ts:107:12)
    at check-orm.ts:42:23
    at Array.forEach (<anonymous>)
    at checkOrmMethods (check-orm.ts:18:10)
    at after5 (index.ts:109:17)
    at hooks.ts:86:35
    at Array.forEach (<anonymous>)
    at runPluginHooks (hooks.ts:86:20)
    at rx-collection.ts:1108:17
```


Super handy feature that we are appreciative for but just asking that we be a little more careful with backwards incompatible changes.

pubkey:
@kicohen2  yes you are right, sorry for that.

PR #4104, opened 2022-11-02
TS Error in Plugin "Replicate-GraphQL" : GraphQL Websocket on "isomorphic-ws" import

jwallet:
## This PR contains:
 - A BUGFIX

Cannot compile my work project since v13, we get TS errors on GraphQL-Websocket plugin (we use the package but not the websocket feature... yet). The "isomorphic-ws" is wrongly imported base on TS Errors we got.

We are on `node 16.13.2`
Using `rxdb@13.8.2`
And it's a `react-native@` project

## Todos
- [ ] Changelog

Here are the TS Errors when running `tsc` on our project

```(shell)
error TS2595: 'WebSocket' can only be imported by using a default import.
  3 import { WebSocket as IsomorphicWebSocket } from 'isomorphic-ws';
error TS2497: This module can only be referenced with ECMAScript imports/exports by turning on the 'allowSyntheticDefaultImports' flag and referencing its default export
  3 import { WebSocket as IsomorphicWebSocket } from 'isomorphic-ws';
```

The important part of `TS2497` was "and referencing its default export"
So both errors were resolved after importing the default export.

//


pubkey:
Thank you, merged.

PR #4107, opened 2022-11-03
test: reproduce 412 pouchdb error

Elendiar:
## This PR contains:
 - TEST that reproduced a bug
[(https://github.com/pubkey/rxdb/discussions/4106)](https://github.com/pubkey/rxdb/discussions/4106)

I started **test:browser** with "_--single-run=false --auto-watch_" and checked IndexedDB in debug manually to make sure sure of different digest.

### DIFF
--- a/test/unit/bug-report.test.ts
+++ b/test/unit/bug-report.test.ts
@@ -9,9 +9,9 @@
  * - 'npm run test:browser' so it runs in the browser
  */
 import assert from 'assert';
-import config from './config';
-
 import PouchDBPlugin from 'pouchdb-adapter-idb';
+
+import config from './config';
 
 import { addRxPlugin, createRxDatabase, randomCouchString } from '../../';
 import { RxDBAttachmentsPlugin } from '../../plugins/attachments';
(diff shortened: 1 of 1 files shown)


pubkey:
> RxError (DB9): RxError (DB9):
createRxDatabase(): Adapter not added. Use addPouchPlugin(require('pouchdb-adapter-[adaptername]'));
Given parameters: {
adapter:"idb"}

Please check.

Elendiar:
> > RxError (DB9): RxError (DB9):
> > createRxDatabase(): Adapter not added. Use addPouchPlugin(require('pouchdb-adapter-[adaptername]'));
> > Given parameters: {
> > adapter:"idb"}
> 
> Please check.
> 
> Also pls try to reproduce this with a smaller attachment so that I can use that in the unit tests without add a such long base64 string.

Fix this and changed base64string.

pubkey:
Ok so interestingly this only fails in the browser but runs in node.js

pubkey:
Thanks,
I double checked, it is not working as it should. PouchDB does not create the same hash as RxDB's md5 hashing which is used for attachments.

pubkey:
I decided to refactor how the attachment hashing works.
In the future it should hash inside of the RxStorage, not from the outside by RxDB itself.
This will make the behavior equal to CouchDB: https://github.com/pouchdb/pouchdb/issues/3156#issuecomment-66831010

Branch: https://github.com/pubkey/rxdb/pull/4115

pubkey:
Fixed here: https://github.com/pubkey/rxdb/pull/4115

PR #4110, opened 2022-11-06
GraphQL replication: should stop syncing if we receive less docs than pull.batchSize

jwallet:
## This PR contains:
 - IMPROVED TESTS (fail test)

If we send the initialSync with a batchSize of `5`, the server last response will contain less than the batchSize (let say `2`) and instead of finishing the initial replication, rxdb will always request a last time because it ends only if there are no document in the response. [Actually the batchSize interface tells us, that the replication will stops if the response contains less documents than the batchSize, which is currently not true](https://github.com/pubkey/rxdb/blob/d7c3aca4f49d605ceae8997df32f713d0fe21ee2/src/types/plugins/replication.d.ts#L47-L53)

 It will also be preferrable to resolve this case, cause we also prevent relying on the server to stop the sync process.
Currently we have, at work, a endpoint that need to be fix, it includes in the response the document that fits the checkpoint.updatedAt, so the initialSync() never stop.

- [x] Tests

### DIFF
--- a/test/unit/replication-graphql.test.ts
+++ b/test/unit/replication-graphql.test.ts
@@ -746,6 +746,48 @@ describe('replication-graphql.test.ts', () => {
                 server.close();
                 c.database.destroy();
             });
+             // https://github.com/pubkey/rxdb/blob/d7c3aca4f49d605ceae8997df32f713d0fe21ee2/src/types/plugins/replication.d.ts#L47-L53
+             it('should stop pulling when response returns less documents than the pull.batchSize', async () => {
+                const amount = batchSize + 1;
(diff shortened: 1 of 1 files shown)


jwallet:
sorry I'm not familiar with karma, I didn't find the equivalent of jest.fn() and expect().toHaveBeenCalledTimes(2). so I used a goold old count++

pubkey:
Can you set `live: false` and check if that also fails?
This would ensure that no time dependend other sync-cycle is running.

pubkey:
Fixed here https://github.com/pubkey/rxdb/pull/4116

jwallet:
thanks @pubkey

PR #4111, opened 2022-11-06
WIP try fix #4107

pubkey:
WORK IN PROGRESS


pubkey:
Closing this. Instead of hashing at the RxDB side, we should rely on hashing at the RxStorage.

PR #4117, opened 2022-11-08
test: active$ should emit during replication

maxnowack:
## This PR contains:
 - IMPROVED TESTS

It seems that the behavior of `replicationState.active$` has changed since v13. In v12, `replicationState.active$` emits when replication is running.
I could not find out when `replicationState.active$` emits in v13. It may not be emitted at all.
This PR only contains a test that fails if `replicationState.active$` has not emitted `true` during the initial replication.
I have not provided a bugfix, because I don't yet understand the correlations in the code regarding the replication well enough.

### DIFF
--- a/test/unit/replication.test.ts
+++ b/test/unit/replication.test.ts
@@ -349,6 +349,39 @@ describe('replication.test.js', () => {
             localCollection.database.destroy();
             remoteCollection.database.destroy();
         });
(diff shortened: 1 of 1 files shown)


pubkey:
Fixed here https://github.com/pubkey/rxdb/pull/4124

PR #4131, opened 2022-11-13
Using rxjs operator first() when observing events breaks db.remove() db.destroy()

jwallet:
## This PR contains:
 - IMPROVED TESTS

This failing test shows that using rxjs operator `first()` (which throws a "No elements in sequence" when it founds nothing) in a db events subscription, breaks the `.remove()` or `.destroy()` of the database.

We may want to remove subscriptions on db and collections before starting to destroy the database.

https://github.com/pubkey/rxdb/discussions/4128

### DIFF
--- a/test/unit/rx-database.test.ts
+++ b/test/unit/rx-database.test.ts
@@ -622,7 +622,6 @@ config.parallel('rx-database.test.js', () => {
             db.$.pipe(first()).subscribe(() => {
                 // subscribe to first db event for debug purposes
             });
-            assert.strictEqual(isRxDatabase(db), true);
             await db.destroy();
         });
     });
--- a/test/unit/rx-database.test.ts
+++ b/test/unit/rx-database.test.ts
@@ -22,6 +22,7 @@ import AsyncTestUtil from 'async-test-util';
 import * as schemas from '../helper/schemas';
 import * as humansCollection from '../helper/humans-collection';
 import * as schemaObjects from '../helper/schema-objects';
+import { first } from 'rxjs';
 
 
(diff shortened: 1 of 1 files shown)


test/unit/rx-database.test.ts
```diff
@@ -607,5 +608,22 @@ config.parallel('rx-database.test.js', () => {
...
+            });
+            await db.addCollections({
+                foobar: {
+                    schema: schemas.human
+                }
+            });
+            db.$.pipe(first()).subscribe(() => {
+                // subscribe to first db event for debug purposes
+            });
+            assert.strictEqual(isRxDatabase(db), true);
```
jwallet, line 625: this line is useless

pubkey:
Ok, so I checked a bit.
The error is not thrown on `destroy()` but afterwards.
The throwing thing is the subscription because it still tries to await the first item which is not possible on canceled observables.

Using this test shows that it throws after destroy() is done:

```ts
                name: randomCouchString(10),
                storage: config.storage.getStorage()
            });
            await db.addCollections({
                foobar: {
                    schema: schemas.human
                }
            });
            const prom = db.$.pipe(first()).subscribe(() => {
                // subscribe to first db event for debug purposes
            });
            await db.destroy();

            console.log('destroyed !!!!');
            await prom;
            console.log('PROMISE DONE');
```

So I do not think there is anything on the RxDB side to do here. Even wrapping the `complete()` call into a try catch, would still produce the error at the outside.

jwallet:
no worries, thank you for looking at it.

PR #4133, opened 2022-11-13
Added example config for Vite/Sveltekit

PeterKey:
## This PR contains:
A configuration for Vite for use with Sveltekit. 

`RxDBDevModePlugin` does not work in Vite/Sveltekit.

I was having troubles getting rxdb to work with Sveltekit - so did someone else in the Discord, but to no avail. Having spent some time getting the right polyfill working I thought this might help someone in the future on a similar path.

This is working for me with the newest version of Sveltekit and it should work the same in Vite.

### DIFF
--- a/config/vite.config.ts
+++ b/config/vite.config.ts
@@ -1,0 +1,22 @@
+// Only required for Sveltekit
+// import { sveltekit } from '@sveltejs/kit/vite';
+
+import type { UserConfig } from 'vite';
+import { NodeModulesPolyfillPlugin } from '@esbuild-plugins/node-modules-polyfill';
+
+const config: UserConfig = {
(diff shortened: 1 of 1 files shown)


pubkey:
I think we should rather finish the svelte-kit example: https://github.com/pubkey/rxdb/pull/3877

Adding an example config without having it in the CI, just means that it will be outdated soon and then confuse users even more.

PeterKey:
Fair enough! 
I am in the process of setting up a project with rxdb + sveltekit anyways. I'll look into it the svelte-kit example - maybe I can be of some help.

pubkey:
Closing this. PR with a CI tests example still welcomed
Related: #3877

PR #4136, opened 2022-11-15
fix: emit false on active$ after replication is done

maxnowack:
It seems that the fix for #4117 in #4124 was not complete. However, the provided test case wasn't complete either.
I've extend the test case to also test if `false` is emitted after replication has completed.

### DIFF
--- a/src/replication-protocol/downstream.ts
+++ b/src/replication-protocol/downstream.ts
@@ -93,15 +93,17 @@ export function startReplicationDownstream<RxDocType, CheckpointType = any>(
                     useTasks.push(innerTaskWithTime.task);
                 }
 
-                if (useTasks.length === 0) {
-                    state.events.active.down.next(false);
-                    return;
-                }
+                if (useTasks.length === 0) return;
 
                 if (useTasks[0] === 'RESYNC') {
                     return downstreamResyncOnce();
                 } else {
                     return downstreamProcessChanges(useTasks);
+                }
+            }).then(() => {
+                state.events.active.down.next(false);
+                if (!state.firstSyncDone.down.getValue()) {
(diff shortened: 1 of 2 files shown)


pubkey:
Test looks good.
I do not have time to fix this, please investigate.

maxnowack:
I've already tried to. It seems that [these lines](https://github.com/pubkey/rxdb/blob/345bec4d89036082a1db686ff61fb61c55065d76/src/replication-protocol/downstream.ts#L96-L99) will never be executed.
I've tried to investigate further, but I'm to unfamiliar with the replication internals right now.

pubkey:
I think that is the correct part to look at but I have no idea why these lines do not get executed.

maxnowack:
Okay, I'll try to dig into it.

maxnowack:
Maybe I have some potential fix.
I've changed the `addNewTask` method, so that `state.events.active.down.next(false)` gets called always after the task processing inside of the `state.streamQueue.down` promise.
I also had to move `state.firstSyncDone.down.next(true)` in front of it. Otherwise `awaitInitialReplication` would finish before `active$` emits `false` after the first replication.
This has the implication that `firstSyncDone.down` now also emits `true` after `downstreamProcessChanges` is finished first.
Please look over it carefully, as I'm not really sure if that is fine or maybe has some other impact I'm now yet aware of.

pubkey:
I merged this. I am not sure if it is 100% correct but I think it is an improvement compared to the current behavior and as long as the tests are green I am confident that nothing major could be broken.

pubkey:
Added this to the changelog here: https://github.com/pubkey/rxdb/commit/f96d264baad92b4d6c7f0a4b832279ae8763e899

PR #4157, opened 2022-11-29
add proposed middleware updates in types

chrisdrackett:
@pubkey Here is a proposal for updated middleware that will play nicely with typescript.

At the moment I only updated the types to start a discussion. Also I only updated `preInsert` and `preSave` but I think its probably a good idea to update all the hooks so the design is consistent.

## Todos 
- [ ] Code
- [X] Tests
- [X] Documentation
- [ ] Typings
- [X] Changelog

### DIFF
--- a/src/types/rx-collection.d.ts
+++ b/src/types/rx-collection.d.ts
@@ -73,28 +73,45 @@ export type RxCollectionHookCallbackNonAsync<RxDocumentType, OrmMethods> = (
     data: RxDocumentType,
     instance: RxDocument<RxDocumentType, OrmMethods>
 ) => void | any;
-export type RxCollectionHookNoInstanceCallback<
+export type RxPreInsertHookCallback<
     RxDocumentType,
-    OrmMethods
+    OrmMethods,
+    InsertMiddlewareInput
 > = (
-    data: RxDocumentType,
-    instance: RxCollection<RxDocumentType, OrmMethods>
-) => Promise<void> | void | any;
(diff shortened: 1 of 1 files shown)


pubkey:
Can you make an example on how the middleware hooks are used now compared to before?
I do not fully understand.

chrisdrackett:
sure, here is an example:

### Old
```ts
myCollection.preInsert(function(plainData){
    // set age to 50 before saving
    plainData.age = 50;
}, false);
```

### New
```ts
type InputData = {
    newAge: number
}

type RxDocumentType = {
    age: number
}

myCollection.preInsert(function(inputData){
    return {
        age: inputData.newAge
    }
}, false);
```

the types would be optional, but this shows how you have full typing all the way through the hook.

pubkey:
Hi  @chrisdrackett 
Sorry for the late answer.
The proposol looks good, however this will be quite some work to implement because we have to tunnel that types to every place. For example when you call `myRxDocument.collection.insert()` etc.

chrisdrackett:
@pubkey yeah, it isn't ideal. The way I essentially handle this right now is I have the input and output types specified, but internally things are untyped. I'm open to suggestions!

pubkey:
Sorry the stale bot was not meant for PRs.

pubkey:
I am closing this. I have no idea on how to implement that with correct internal typings.
Maybee we have to first change the whole middleware api for that.

PR #4190, opened 2022-12-21
Composite Primary Keys broken on replicated collections

PeterKey:
When documents are added to a collection with a composite primary key through the pull handler of a replication, the resulting composite primary key is accessed before it is generated.

### DIFF
--- a/test/unit.test.ts
+++ b/test/unit.test.ts
@@ -1,70 +1,70 @@
-import './unit/init.test';
-import './unit/util.test';
-import './unit/pouch-db-integration.test';
-import './unit/adapter-check.test';
+// import './unit/init.test';
+// import './unit/util.test';
+// import './unit/pouch-db-integration.test';
+// import './unit/adapter-check.test';
 
 
-/**
- * Helpers that
- * do not fully test RxDB but
- * just single methods
- */
-import './unit/custom-index.test';
-import './unit/query-planner.test';
+// /**
+//  * Helpers that
+//  * do not fully test RxDB but
+//  * just single methods
+//  */
+// import './unit/custom-index.test';
+// import './unit/query-planner.test';
 
 
-/**
- * Move these tests around so that
- * when you focus on one part of RxDB,
- * your relevant tests run first.
- * Do not commit this file if you modified the order.
- */
-import './unit/rx-storage-implementations.test';
-import './unit/rx-storage-query-correctness.test';
+// /**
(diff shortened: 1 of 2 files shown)


PeterKey:
This is a work in progress - I wanted to check if the unit tests actually work. I hope it is ok that I already opened the pull request.

PeterKey:
@pubkey the bug reproduction is finished. 
I tried to look for a possible fix, however I lack a deeper grasp of the replication flow to be confident about my reasoning.

pubkey:
Test looks ok, however this was intentional. RxDB expects the remote to only send documents that exactly match the schema, so the primary key cannot be missing.
I would merge a feature to fill up the primary key if possible, but I will not do the work by myself.
If we fill up missing keys, we also need a check to ensure that an error is thrown if the composite primary key is there but different than how RxDB would create it.

PeterKey:
I find it odd that the behaviour is different for data coming through replication vs. data coming through `insert()`. What is the reasoning behind this?

I thought the logic behind supporting composite primary keys in general was to make it compatible with backend data models - in my case a postgres database - but the composite key behaviour of the backend is not directly compatible with rxdb (concatinating two strings with a separator).
I think it is bad design to expect the remote data/api to reflect an implementation that is configured and only relevant on the client. 

Anyways, I'd be interested to add this if you can give me a little hint where the change would make sense. 
Otherwise I'll make a pull request to add additional documentation on the current state of affairs. Right now it sounds like the key added for the composite primary is solely there for rxdb to fill it in.

pubkey:
You are totally right, the current behavior is not correct. I just wanted to tell you that, so you do not have to look at error at your side.


> Anyways, I'd be interested to add this if you can give me a little hint where the change would make sense.

If you want to fix that, it should be done in the `replication` plugin. There we already transform all data that comes from the server. Atm it only replaces the custom `deleted` field with `_deleted`. This is done with the function `swapdeletedFieldToDefaultDeleted()`. So you can check the parts of the code where this function is used, and there it should also check/fill-up the primary key. The solution will be just a few lines of code, but building the correct tests could be some work to do.

PeterKey:
@pubkey I just wanted to start working on this and saw you closed this with a fix! Thank you! 
Do you still need some help with additional tests?

pubkey:
@PeterKey 
I just added your single test.
We could need an additional test to ensure the composite key is filled when the document comes from the pull.stream$.

PR #4192, opened 2022-12-21
Adding Vite Vanilla Typescript Example

SebasG22:

This PR adds a clear example of how to use the RxDB library over the Vite ecosystem with vanilla TS.
Currently, Vite users do not have a clear demo for their usage.


SebasG22:
Hey @pubkey,
Can you take a look at this PR?

pubkey:
Hi @SebasG22 
I am not sure if I want to maintain more examples inside of this repository. There are already too many and some of them are hardly outdated like the ionic2 example or the vanilla js one.
Also my current goal is to make all example implement the exact same features (hero-list) so that I can unify the end to end test.
So I think you should create a separate repository for the vite example. I added a section [here](https://github.com/pubkey/rxdb/blob/master/examples/README.md#third-party-examples) so that you can link it from there and other can find it.

pubkey:
Ok I understand.
Atm RxDB already has a CI-tested rollup build here: https://github.com/pubkey/rxdb/blob/master/package.json#L118
Would it be possible to do something similar with vite, so that it only tests the build itself without having to maintain the whole js/css/html from the example?

SebasG22:
> Ok I understand. Atm RxDB already has a CI-tested rollup build here: https://github.com/pubkey/rxdb/blob/master/package.json#L118 Would it be possible to do something similar with vite, so that it only tests the build itself without having to maintain the whole js/css/html from the example?

Hey @pubkey,
I was reviewing the test you have for rollup and I consider that the examples folder as it is being used today has two uses mainly:
1. to be a reference for a developer, with any level of experience on how to implement the RxDB library in the desired framework.
2. To ensure with E2E tests that the examples are still valid and working.

Change the natural form of the examples to save time in CI, I think it may work in some cases like this but not in integrations with other libraries, e.g. react, svelte, vue where what really matters is the implementation within each of them.

It would be interesting not to limit the examples part, and that each developer can easily find one.

To improve CI times, we can review the option of having a cache at repository level. That if things have not been affected, certain commands such as test or lint are not executed. Or even always have a cache for the version generated in each build.

By the way, I'm not sure why the check in Github is failing. Do you have any clue how I could fix it

pubkey:
It is more about maintaining more example, less about the CI time.
But however I thought about it and I am ok with merging the vite example.
However the example should have a hero-list, similar to the other examples.

> By the way, I'm not sure why the check in Github is failing. Do you have any clue how I could fix it

The `npm run test:e2e` command is looking for the tests at the `test` folder which is not included in your PR. You should copy the e2e tests from one of the other examples like [here](https://github.com/pubkey/rxdb/tree/master/examples/angular/test).

- You should not include the logo files directly because they can get out of date easily. Instead you can copy them from the logo folder like it is done [here](https://github.com/pubkey/rxdb/blob/master/examples/angular/package.json#L10)

SebasG22:
> ### This branch has no conflicts with the base branch

@pubkey 
Thanks for the feedback.
The PR was updated by changing the example to align with the current heroes' examples.
Additionally, the RXDB logo is now being copied and there are e2e tests for it.

pubkey:
CI errors with `Fail! Modules in package.json not used in code: eslint-plugin-testcafe`.
The examples should not be linted on a root `npm run lint` because that is already so slow.
So I am ok with adding it to the [eslintignore](https://github.com/pubkey/rxdb/blob/master/.eslintignore), the other examples are already ignored anyway.

SebasG22:
> ot be linted on a root `npm run lint` because that is already so slow

Sure, let me revert that change and add it to eslintignore.

pubkey:
Thank you, merged!

PR #4179, opened 2022-12-13
Draft: feat(package): wip: use exports in package.json

L-U-C-K-Y:
fixes #3422

Draft

## This PR contains:
Uses the exports in package.json

## Describe the problem you have without this PR
Better imports

## Todos 

WIP
- [ ] Tests
- [ ] Documentation
- [ ] Typings
- [ ] Changelog

PR #4196, opened 2022-12-22
feat: adding support for esm modules entries

SebasG22:

This PR fixes the issue on some javascript environments (Sveltekit) when you get a resolution error for plugins dependencies as

```
Error [ERR_UNSUPPORTED_DIR_IMPORT]: Directory import 'C:\Users\SebasG\Documents\svelte-kit-demo\node_modules\rxdb\plugins\query-builder' is not supported resolving ES modules
```

### How was fixed?
Adding entries to the package JSON to ensure the validation works correctly. 

After this PR is merged we can add a sveltekit example to test this issue and support it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,139 +35,241 @@
   "sideEffects": false,
   "exports": {
     ".": {
+      "types": "./dist/types/index.d.ts",
       "node": "./dist/lib/index.js",
+      "require": "./dist/lib/index.js",
+      "es2015": "./dist/es/index.js",
       "default": "./dist/es/index.js"
     },
      "./plugins/backup": {
+      "types": "./dist/types/plugins/backup/index.d.ts",
       "node": "./dist/lib/plugins/backup/index.js",
+      "require": "./dist/lib/plugins/backup/index.js",
+      "es2015": "./dist/es/plugins/backup/index.js",
       "default": "./dist/es/plugins/backup/index.js"
     },
      "./plugins/cleanup": {
+      "types": "./dist/types/plugins/cleanup/index.d.ts",
       "node": "./dist/lib/plugins/cleanup/index.js",
+      "require": "./dist/lib/plugins/cleanup/index.js",
+      "es2015": "./dist/es/plugins/cleanup/index.js",
       "default": "./dist/es/plugins/cleanup/index.js"
     },
     "./plugins/crdt": {
-      "node": "./dist/lib/plugins/cdrt/index.js",
-      "default": "./dist/es/plugins/cdrt/index.js"
+      "types": "./dist/types/plugins/crdt/index.d.ts",
+      "node": "./dist/lib/plugins/crdt/index.js",
+      "require": "./dist/lib/plugins/crdt/index.js",
+      "es2015": "./dist/es/plugins/crdt/index.js",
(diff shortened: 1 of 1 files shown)


pubkey:
Related: https://github.com/pubkey/rxdb/pull/4179

pubkey:
Hi.
Since this adds the `exports` fields to the package.json, should that make the `package.json` files for individual plugins in [this folder](https://github.com/pubkey/rxdb/tree/master/plugins) obsolete?
Also should we set `"type": "module"` in the root package.json?

My knowledge arround the whole module/csm/esm/export syntax is lacking, so I might be wrong.

Is this a breaking change? Because for the next major release, there is the [feature/14.0.0](https://github.com/pubkey/rxdb/pull/4181) branch.

SebasG22:
I wasn't able to test this change against the premium plugins due to

> Hi. Since this adds the `exports` fields to the package.json, should that make the `package.json` files for individual plugins in [this folder](https://github.com/pubkey/rxdb/tree/master/plugins) obsolete? Also should we set `"type": "module"` in the root package.json?
> 
> My knowledge arround the whole module/csm/esm/export syntax is lacking, so I might be wrong.
> 
> Is this a breaking change? Because for the next major release, there is the [feature/14.0.0](https://github.com/pubkey/rxdb/pull/4181) branch.

Hey @pubkey,
Additionally, I could not test this change in projects that include the premium plugins because I don't have access to it.
If you want I can do some tests

I am going to do some testing to see if it is possible to remove the package.json from each plugin. Although initially I don't think it would be a good idea, I want to explore it.

I don't consider it a breaking change, since on the user side it shouldn't change the implementation at all. But I'm not sure what standard you are using to consider something minor or major.

pubkey:
>  I could not test this change in projects that include the premium plugins because I don't have access to it

I will fix these afterwards by myself, they anyway run in a different project so it will not fail the rxdb CI.

I think the way rxjs does it is the way to go https://github.com/ReactiveX/rxjs/blob/master/package.json and they also do not have differente package.json files.

I am not sure if this is a breaking change but I would prever to release it on 14.0.0 because the chances are high that it will break some peoples build configuration.

SebasG22:
We are aligned in the way RXJS handles exports in the json package.

Currently removing the package.json inside the plugin folder breaks all the code imports because they are always pointing to dist instead of itself.

I consider that this is a first step on the way to improving the codebase, which can be merged and then make a refactoring where you have a package.json for each plugin for this purpose.

If you merge this to master, can you backport it to the release branch?

swnf:
> I think the way rxjs does it is the way to go https://github.com/ReactiveX/rxjs/blob/master/package.json and they also do not have differente package.json files.

@pubkey I think rxjs does not have multiple `package.json` in their github repo, but the package on npm contains different `package.json` for the paths:

```
./fetch/package.json
./ajax/package.json
./webSocket/package.json
./package.json
./operators/package.json
./testing/package.json
```

pubkey:
@swnf Ah yes indeed they have.
So we should do the same. So the only thing missing is to generate the `plugins/` folder from the content of the root package.json during build time.

pubkey:
Merged, I added the generation of the package.json files on the master branch.

PR #4203, opened 2022-12-29
Add missing mingo array operators

swnf:
## This PR contains:

 - IMPROVED TESTS
 - A BUGFIX

The `$elemMatch` operator did not work for me after upgrading to 13.17.0. It seems like the array operators `$elemMatch`, `$all` and `$size` are not enabled in mingo anymore. But they are part of the RxQuery type definition:

https://github.com/pubkey/rxdb/blob/d828b8181e5c24c08d32429923508fa8d2e1a5ac/src/types/rx-query.d.ts#L20-L22

This PR enables them and adds a corresponding test.

### DIFF
--- a/src/rx-query-mingo.ts
+++ b/src/rx-query-mingo.ts
@@ -6,6 +6,7 @@ import { $sort, $project } from 'mingo/operators/pipeline';
 import {
     $and,
     $eq,
+    $elemMatch,
     $exists,
     $gt,
     $gte,
@@ -19,6 +20,7 @@ import {
     $not,
     $or,
     $regex,
+    $size,
     $type,
 } from 'mingo/operators/query';
 
@@ -44,6 +46,7 @@ export function getMingoQuery<RxDocType>(
         useOperators(OperatorType.QUERY, {
             $and,
             $eq,
+            $elemMatch,
             $exists,
             $gt,
             $gte,
@@ -57,6 +60,7 @@ export function getMingoQuery<RxDocType>(
             $not,
             $or,
             $regex,
+            $size,
             $type,
         } as any);
         mingoInitDone = true;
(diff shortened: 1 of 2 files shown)


pubkey:
Hi @swnf 
Thank you for the PR.
The goal for this was to unify which operators work and which do not work across all RxStorage implementations.
Of course most stuff does not work on PouchDB which is deprecated anyway.
However if an operator does not work on the lokijs storage, I will not add it to RxDB.

I think we should first merge the `$elemMatch` and `$size`.
But for `$all`, we need to first find a workaround to make it work with lokijs.

swnf:
I removed `$all` from this PR. If you want consistent support, it might make sense to just drop support for `$all` and remove it from the `RxQuery` type definition in RxDB 14. I think it is just a fancy way to write `$and` (see [here](https://www.mongodb.com/docs/manual/reference/operator/query/all/#equivalent-to--and-operation)).

pubkey:
I removed `$all` from the types https://github.com/pubkey/rxdb/commit/9d669c1978644a9f7dbda81ef57ecdc1d3497fdb

Strange that the CI fails, @SebasG22 can you check?

SebasG22:
> I removed `$all` from the types [9d669c1](https://github.com/pubkey/rxdb/commit/9d669c1978644a9f7dbda81ef57ecdc1d3497fdb)
> 
> Strange that the CI fails, @SebasG22 can you check?

Looks like it's some wired behavior, can you retry the job for this one and see if that works?
Could be related to a timeout issue that the element does not exist when the test started...
If this is repetitive I can take a deep look, because the changes in the PR are not related to the example at all.

pubkey:
@SebasG22  I fixed it by pinning the vite versions https://github.com/pubkey/rxdb/commit/8387dbd9afa192f72865cf9fe04b04f64ed40b46

PR #4299, opened 2023-01-12
Couchdb replication bug report

Unforgiven-wanda:
Hello,
As requested, here is a pull that reproduces the CouchDB replication bug.
Had to add `node-fetch@2` because replication `fetch` would fail during the test.
Other than that just have a CouchDB instance running, edit the credentials and give it a try.

As always, thank you for your time.

### DIFF
--- a/package.json
+++ b/package.json
@@ -481,6 +481,7 @@
     "mocha": "10.2.0",
     "mocha.parallel": "0.15.6",
     "nconf": "0.12.0",
+    "node-fetch": "2.6.7",
     "node-pre-gyp": "0.17.0",
     "pouchdb": "8.0.0",
     "pre-commit": "1.2.2",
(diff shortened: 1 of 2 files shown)


pubkey:
If the normal node fetch does not work, we should fix that first.
In the tests we normally use the pouchdb server as couchdb endpoint, does it also fail with that?

Unforgiven-wanda:
@pubkey I tested it just now with `pouchdb-server` and received the exact same error.

pubkey:
Pls adapt the test so that the CI reproduces your problem

Unforgiven-wanda:
Done.

pubkey:
Thank you so much. Fixed in 657de005c6b45f4218d03a9c8e79c1cd3738360b

PR #4307, opened 2023-01-16
Correct way to import socket.io-client in P2P Simple Peer Replication

YanDevDe:
Importing `io` was not correct, and it caused the issue for my builds. 

The correct way to import `io` would be:

```
const { io } = require("socket.io-client");
```

and not

```
const io = require("socket.io-client");
```

According to the doc:

https://socket.io/docs/v4/client-initialization/

And this worked just fine at my side.


pubkey:
Thank you, merged.

Issue #4285, opened 2023-01-08
ANNOUNCEMENT: Test RxDB 14.0.0 BETA


pubkey:
RxDB version 14.0.0 is now it beta and you should try it out.
Bugs found related to the beta version, will be fixed by the maintainer.
The new version will be in beta for about **4 weeks**.
If you have problems after the 14.0.0 non-beta release, you will have to fix them by yourself.

Release notes: https://github.com/pubkey/rxdb/blob/master/docs-src/releases/14.0.0.md

You can find the latest beta version here: https://www.npmjs.com/package/rxdb?activeTab=versions

swnf:
@pubkey I think it might be necessary for RxDB users to increment their schema version if a schema contains arrays (like `required`). Otherwise, an application might work in one browser and break in another browser because the new hash does not match. Maybe this should be mentioned in the release notes?

pubkey:
@meabed No, PouchDB will no longer be maintained, [see](https://rxdb.info/questions-answers.html). It just makes no sense and nothing on PouchDB really worked.

@nickchomey I normally do not fix other peoples bugs. This is an open source project.

@swnf Yes you are right, I will put that into the release notes.

Unforgiven-wanda:
As of beta.18 the replication to CouchDB is still broken. Upstream replication will occasionally and randomly get `ensurenotfalsy() is falsy` on [this line](https://github.com/pubkey/rxdb/blob/master/src/plugins/replication-couchdb/index.ts#L137)

Unforgiven-wanda:
Hello again,
I'd like to report that as of beta.25 the **live replication** on CouchDB is broken. (Yes, live is set to true in `replicateCouchDB`)
Upstream replication (edits on the client) function perfectly.
But downstream syncs (edits on the CouchDB instance in the server) do not happen. No error is thrown.
Please see the following for an example: [Video](https://streamable.com/nbyjxf)
I'd like to create a test for this, but I honestly have no idea how, in this particular case.
Any insight of yours is most welcome.
As always, thank you for your time.

pubkey:
@Unforgiven-wanda to reproduce the problem, you could just create 2 collections and replicate them with one server, like [here](https://github.com/pubkey/rxdb/blob/master/test/unit/replication-couchdb.test.ts#L128). Then do the document writes on either side of the collections.
If that does not reproduce the problem, you'd have to modifiy the document state by runnning POST requests against the PouchDB server that is used in the tests.

[...]

PR #4319, opened 2023-01-20
CouchDB Replication

Unforgiven-wanda:
Hello again,
CouchDB replication is essentially broken in this state because it has issues that do not appear when using Pouchdb-server in tests.
To test this pull request you'll have to have an instance of CouchDB running, either locally or not.
I've been at this for a few days, and I can tell you there are three main issues when replicating with CouchDB:

1. The one detailed in this test: If the database you replicate to contains `deleted` documents, the initial replication will fail with `Cannot read properties of null (reading 'id')`. In the test present in this pull request, we create a pristine CouchDB database, sync to it 3 created documents, delete two and then destroy the DB. Upon running the test again however, this error will throw on initial replication.
2. `Ensurenotfalsy` that we thought fixed in https://github.com/pubkey/rxdb/commit/657de005c6b45f4218d03a9c8e79c1cd3738360b is still present, but ever more randomly, I haven't been able to reproduce this reliably with CouchDB replication, but as far as I can tell, it doesn't happen anymore with PouchDB-server in tests. It might occur after 2 edits or a hundred, and honestly I am stumped when it comes to this one.
3. Downstream replication will not occur if the document has been edited locally (**Always in CouchDB replication, not Pouchdb-server**). Edits on CouchDB's Fauxton (or other users) are replicated fine on app start, but the second you edit a document, it ceases to receive downstream updates. I have inquired about this [here](https://github.com/pubkey/rxdb/issues/4285#issuecomment-1386765617).

In its current state, CouchDB replication has too many issues that do not pop up in unit tests, because those use PouchDB-server. What the differences between the two are I do not know, but there has to be some glaring ones because they behave completely differently in the issues I have pointed out above.
I will try to create tests for the two other issues once we hopefully fix the first one.
As always, thank you for your invaluable time.

### DIFF
--- a/test/unit/replication-couchdb.test.ts
+++ b/test/unit/replication-couchdb.test.ts
@@ -258,6 +258,12 @@ describe('replication-couchdb.test.ts', () => {
             await replicationState.awaitInitialReplication();
             return replicationState;
         }
+		
+		function delay(time: number): Promise<void> {
+            return new Promise<void>((resolve) =>
+                setTimeout(resolve, time * 1000)
+            );
+        }
 
         it('should stream changes over the replication to a query', async () => {
             const server = await SpawnServer.spawn();
@@ -310,6 +316,34 @@ describe('replication-couchdb.test.ts', () => {
             await doc2.incrementalPatch({ age: 22 });
             await awaitInSync();
             assert.strictEqual(doc1.getLatest().age, 22);
+			
+			// Edit on the server and wait for it to be pulled
+            const doc = await fetchWithCouchDBAuth(
+                server.url + '/' + doc1.passportId,
+                {
+                    method: 'GET',
+                }
+            );
+
+            const docJson = await doc.json();
+            fetchWithCouchDBAuth(server.url + '/' + doc1.passportId, {
+                body: JSON.stringify({
+                    ...docJson,
+                    firstName: 'edited remotely',
+                }),
+                method: 'PUT',
+            });
+
+            await awaitInSync();
+            // Delay for a second to make sure the change is pulled
+            // Even though it should be instanenous
+            // But just in case
+            await delay(3);
+            const editedDoc = await c1.findOne().exec(true);
+            assert.strictEqual(
+                editedDoc.getLatest().firstName,
+                'edited remotely'
+            );
 
             c1.database.destroy();
             c2.database.destroy();


pubkey:
@Unforgiven-wanda I tried running the current tests against a real couchdb server. You are right, nothing works. Give me some time to fix the basics, then we can check if your issues are also solved.

pubkey:
@Unforgiven-wanda I added a test run `npm run test:replication-couchdb` which runs the couchdb replication tests against a native couchdb server. Please add your tests there so we can reproduce your problem.

Unforgiven-wanda:
@pubkey It seems the issues listed above only occur on CouchDB 3.x. When testing against CouchDB 2.3.1 the only issue that remains is the one where downstream replication doesn't happen. Switch docker to CouchDB 3.2.2 and it occurs again.
So our culprit here was CouchDB 3.x all along.

I have added a `test:replication-couchdb3` npm command for you look at.

I will adapt the test for failing downstream replication for Couchdb2 as soon as I can.

pubkey:
@Unforgiven-wanda  I updated the tests to use couchdb v3 here: https://github.com/pubkey/rxdb/pull/4336
But there it works, so I have no idea what you are doing wrong.

pubkey:
I added some more tests. I think that is all I can do.

Unforgiven-wanda:
Please give me a couple days before closing this eventually. I've been bombarded with unexpected work and haven't had time to look through this.
As always, thank you for your time.

pubkey:
@Unforgiven-wanda No hurry. The v14 will be in beta for 3-4 more weeks. I can leave this open.

Unforgiven-wanda:
Small addition to `should stream changes over the replication to a query` test, demonstrating how edits directly on the CouchDB instance aren't reflected on the local database.

pubkey:
@Unforgiven-wanda 
I think I found the root problem. In v13 the assumption was that RxDB creates the exact same md5 hash revisions as CouchDB would create. But this is not true for v14. I refactored the whole revisision and conflict handling for the CouchDB replication here https://github.com/pubkey/rxdb/pull/4348

Unforgiven-wanda:
Downstream replication is fixed. And I haven't been able to reproduce the other issues since, so I'm assuming (here's hoping) they're also fixed.
Closing this then.
Please allow me to thank you once again for your fantastic work and time invested. You are a prince among open-source maintainers.
Have a wonderful week.

pubkey:
@Unforgiven-wanda great. Thank you for the investigation. Releasing a broken CouchDB replication while removing PouchDB support would be painful :)

mbret:
I am using the v14 with a 3.2 couchdb server and I still have the downstream replication broken together with upstream (randomly). Another big problem is as soon as you get offline, the replication will continuesly loop on request retry and do a thousands of requetes per seconds.

pubkey:
Make  PR with a failing test.

PR #4394, opened 2023-02-03
Fix mingo $type import and add test

swnf:
## This PR contains:
 - IMPROVED TESTS
 - A BUGFIX

Using `$type` in a query throws an exception


pubkey:
Thank you, merged.

Issue #4347, opened 2023-01-25
Firebase replication throwing error


ravi-poonia:
RXDB version : 13.17.1

Firebase replication throws the following error:

```
RxError (RC_PULL): RxError (RC_PULL):
RxReplication pull handler threw an error - see .errors for more details
Given parameters: {
checkpoint:null
errors:[
  {
    "code": "invalid-argument",
    "customData": null,
    "name": "FirebaseError"
  }
]
direction:"pull"}
    at newRxError (/Users/ravipoonia/code/transport-js/node_modules/rxdb/src/rx-error.ts:107:12)
    at /Users/ravipoonia/code/transport-js/node_modules/rxdb/src/plugins/replication/index.ts:204:57
    at processTicksAndRejections (node:internal/process/task_queues:96:5) {
  code: 'RC_PULL',
  parameters: {
    checkpoint: undefined,
    errors: [ [FirebaseError] ],
    direction: 'pull'
  },
  rxdb: true
}
```

As you can see, its impossible to figure out what the issue is from the above error message. Finally i was able to figure out something after trying out the `test:replication-firestore` test with real firebase config. 

```
{
    "code": "RC_PULL",
    "parameters": {
        "checkpoint": {
            "id": "qbzieljdgnqt",
            "serverTimestamp": "2023-01-25T11:18:02.068Z"
        },
        "errors": [
            {
                "name": "FirebaseError",
                "message": "Order by clause cannot contain a field with an equality filter serverTimestamp",
                "code": "invalid-argument",
                "stack": "FirebaseError: Order by clause cannot contain a field with an equality filter serverTimestamp"
            }
        ],
        "direction": "pull"
    },
    "rxdb": true
}
```

After digging some more into it, I think the culprit is this line
https://github.com/pubkey/rxdb/blob/7607a8b9b2346f5de962eac9d2e04045a36d599c/src/plugins/replication-firestore/index.ts#L135

After this also it kept throwing some errors about indexes, I'm not sure about those. I'll have to look into it some more, but the above mentioned is a sure issue.

pubkey:
The firestore dev mock does not support the changestream, so this is not testable in the CI.
Also I do not have a firebase account atm, so please make a PR with a proposed fix.

ravi-poonia:
[...]

PR #4405, opened 2023-02-06
Fixed Firebase Replication Issues

ravi-poonia:
#4347 

Added fix for the Following
- [IN query only allows 10 in a batch ](https://stackoverflow.com/questions/61354866/is-there-a-workaround-for-the-firebase-query-in-limit-to-10)
- firebase errors getting missing while using in live app (better way to stringify errors)


src/plugins/replication-firestore/firestore-helper.ts
```diff
@@ -57,3 +58,19 @@ export function stripPrimaryKey(
...
     return docData;
 }
+
+// https://stackoverflow.com/questions/61354866/is-there-a-workaround-for-the-firebase-query-in-limit-to-10
+export function getContentByIds<RxDocType>(ids: string[], getQuery: GetQuery<RxDocType>): Promise<QueryDocumentSnapshot<RxDocType>[]> {
+    const batches = [];
+
+    while (ids.length) {
+        // firestore limits batches to 10
+        const batch = ids.splice(0, 10);
```
pubkey, line 68: Here we can use the batchArray method from the utils https://github.com/pubkey/rxdb/blob/master/src/plugins/utils/utils-array.ts#L24

PR #4441, opened 2023-02-16
FIx react-native-database.md

Sukikiroi:
i did remove word  "that" has being added by mistake

i did removed a word that have being added by mistake


pubkey:
Thanks, merged!

PR #4454, opened 2023-02-19
feature request: populate multiple paths like in mongoose

darkowic:
## This PR contains:
**Feature request: populate multiple paths like in mongoose**
https://mongoosejs.com/docs/populate.html#populating-multiple-paths

There are cases where you want to easily populate fields for multiple find results. Current option is only to run this as separate async array looping and its not convenient to use.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ release-body.md
 /test-db-*
 dist/typings
 *-debug.log
+.vscode
--- a/test/unit/population.test.ts
+++ b/test/unit/population.test.ts
@@ -270,6 +270,19 @@ config.parallel('population.test.js', () => {
             });
         });
     });
+    describe('RxQuery().populate()', () => {
+        describe('positive', () => {
+            it('populates all top-level-field', async () => {
+                const col = await humansCollection.createRelated();
+                const docs = await col.find().populate('bestFriend').exec(true);
+                const doc = docs[0];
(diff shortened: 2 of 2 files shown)


darkowic:
Basically, the final goal would be to have an option to synchronously access the related document properties.

pubkey:
Hi @darkowic 
I understand the use case but I think this can be better implemented as own RxDB plugin.
We can link to it from [here](https://github.com/pubkey/rxdb/blob/master/docs-src/third-party-plugins.md)

darkowic:
Working on some workarounds yesterday to make it work, I also realized that calling `populate` is not modifying the document, but just returns the populated document. In Mongoose: https://mongoosejs.com/docs/populate.html#population the document is updated with all the fields and can be directly accessed from the parent. This is sooo much easier to use. Do you think its all possible via plugin? Together with handling obserability etc? (I'm really new user of this library)

pubkey:
RxDB works on the plain documents data so that it can de-duplicate them in the cache.
That is also why there are no "select" queries.
Of course you can build whatever plugin you like which also mutates the plain document data, but as said, that should be done in your plugin.

When I started RxDB some years ago, a big goal was to let it behave similar to mongoose. But over the time it became clear that mongoose has many problems and tracking their behavior is often not the best solution. Especially when it comes to have correct typescript support.

PR #4496, opened 2023-02-27
blocked 'e' from team size input

DDeerrpp:

### DIFF
--- a/docs-src/premium.html
+++ b/docs-src/premium.html
@@ -296,7 +296,7 @@
                                     min="1"
                                     max="100"
                                     required
-                                    onkeydown="return event.keyCode !== 69"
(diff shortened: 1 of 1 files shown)


DDeerrpp:
Pretty sure it really doesn't do a difference but just to block the letter 'e' from being input as it doesn't seem like it's needed.

pubkey:
Nice :)
Should we also block the minus symbol `-` ?

DDeerrpp:
Yea I forgot, lemme add see if it can block the period and minus symbols too

DDeerrpp:
Alright it should block the keycodes 'e', '.', and '-' now in that input number type!

pubkey:
Thank you, merged.

Issue #4113, opened 2022-11-06
GraphQL replication: WebSocket client headers support


nirvdrum:
As part of my RxDB 11 -> 13 upgrade, I've tried to move away from the old style of GraphQL subscriptions via custom WebSockets to RxDB's built-in support. I was confused by why the headers were passed to the `streamQueryBuilder` at first and thought maybe that was just a way to pass state through the system. But, it looks like RxDB is expecting the headers to be part of the GraphQL subscription query. Unfortunately, that's not how all GraphQL servers work.

In my case, I'm using Hasura. It expects the header to be a true HTTP header and to the best of my knowledge, there's no option to pass the token as part of the subscription itself. Instead, the header should be passed as part of the `createClient` call. My old code for doing so looks like:

```ts
export function createWebSocketClient(
  tokenFetcher: Auth0ContextInterface['getAccessTokenSilently']
): Client {
  return createClient({
    url: GRAPHQL_WS_URL,
    connectionParams: async () => {
      const token = await tokenFetcher();

      return {
        headers: { authorization: `Bearer ${token}` }
      };
    }
  });
}
```

Obviously, we wouldn't need any of the Auth0 stuff. RxDB already has the headers in `mutableClientState.headers`. I think it just needs an option to pass the header through in `getGraphQLWebSocket`.

pubkey:
Yes sounds good. You can make a PR if you need this.

L-U-C-K-Y:
> As part of my RxDB 11 -> 13 upgrade, I've tried to move away from the old style of GraphQL subscriptions via custom WebSockets to RxDB's built-in support. I was confused by why the headers were passed to the `streamQueryBuilder` at first and thought maybe that was just a way to pass state through the system. But, it looks like RxDB is expecting the headers to be part of the GraphQL subscription query. Unfortunately, that's not how all GraphQL servers work.
> 
> 
> 
> In my case, I'm using Hasura. It expects the header to be a true HTTP header and to the best of my knowledge, there's no option to pass the token as part of the subscription itself. Instead, the header should be passed as part of the `createClient` call. My old code for doing so looks like:
> 
> 
> 
> ```ts
> 
> export function createWebSocketClient(
> 
>   tokenFetcher: Auth0ContextInterface['getAccessTokenSilently']
> 
> ): Client {
> 
>   return createClient({
> 
>     url: GRAPHQL_WS_URL,
> 
>     connectionParams: async () => {
> 
>       const token = await tokenFetcher();
[...]

PR #4533, opened 2023-03-09
ADD 'includeWsHeaders' property for GraphQL replication

johnnyvdberg:
## This PR contains:
 - A NEW FEATURE

I ran into a similar issue as https://github.com/pubkey/rxdb/issues/4113. Some GraphQL backends only support authenticating web sockets by setting an HTTP header. To circumvent the current behaviour, I implemented my own replication strategy, but with this small change we can get back to the official replication strategy. 

Proposed use:
```ts
{
      ...
      headers: {
        Authorization: <Bearer token>,
      },
      pull: {
        ...
        includeWsHeaders: true,
      }
}
```

## Todos 
- [x] Tests
- [x] Documentation
- [x] Typings

Let me know what you think

### DIFF
--- a/docs-src/replication-graphql.md
+++ b/docs-src/replication-graphql.md
@@ -317,7 +317,8 @@ const replicationState = replicateGraphQL(
         pull: {
             batchSize: 100,
             queryBuilder: pullQueryBuilder,
-            streamQueryBuilder: pullStreamQueryBuilder
+            streamQueryBuilder: pullStreamQueryBuilder,
+            includeWsHeaders: false, // Includes headers as connection parameter to Websocket.
         },
         deletedField: 'deleted'
     }


pubkey:
Code looks good. But we also need a test to ensure this will not break in the future.

johnnyvdberg:
Added a test, I struggled to get the tests running on an M1 machine, so could contain an error.

pubkey:
This test only ensures that the headers exist on the client state, not that they reach the backend.
The test should reproduce your initial problem.

johnnyvdberg:
I think the test might be a bit flaky, I wasn't able to reproduce the same error when testing locally.

pubkey:
Thank you, merged!

PR #4547, opened 2023-03-11
example with quasar-framework

TobyMosque:
## This PR contains:
an new example


pubkey:
We need some end2end tests. See the [angular example](https://github.com/pubkey/rxdb/blob/master/examples/angular/test/e2e.test.js) for reference.

pubkey:
Any update on this?

TobyMosque:
I'm trying to expand the current example, if u prefer u can close that for now, I'll reopen when i finish the improvements/tests/cicd.

TobyMosque:
@pubkey done.
image
image

pubkey:
Sorry for not mentioning this before. We should use the testcafe module for testing. In the future the goal is to reuse the test-runner for all examples and run the exact same tests on all of these.
Also please do not commit you yarn.lock file.

PR #4549, opened 2023-03-11
commented on ignoreDuplicats in svelte example

ianengelbrecht:
RxDB throwing repeatedly during Vite dev builds with Svelte. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Typings
- [ ] Changelog

!!!Please advise on how this should be managed and I will attempt to fix in the code / docs-src, depending on where most appropriate.

### DIFF
--- a/examples/svelte/src/store.js
+++ b/examples/svelte/src/store.js
@@ -20,6 +20,13 @@ const _create = async () => {
     storage: wrappedValidateAjvStorage({
       storage: getRxStorageDexie(),
     }),
+    /*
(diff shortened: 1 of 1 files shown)


pubkey:
> To be honest I think createRxDatabase
      should just return the database if it exists, like Dexie does, without throwing an error

Dexie has no form of caching so they can do this. In the past, many people accidentially created duplicated RxDatabase instances which caused big performance problems and also problems with the leader election.

Please do not abuse PRs to ask questions.

PR #4579, opened 2023-03-20
fix: typescript schema to allow nullable fields

johannesjo:
Otherwise any schema with a type like:
```
   myfield: {
      type: ["string", "null"],
    },
```
will produce an error.

## Todos 
- [ ] Tests
- [ ] Documentation
- [ ] Typings
- [ ] Changelog

### DIFF
--- a/src/types/rx-schema.d.ts
+++ b/src/types/rx-schema.d.ts
@@ -5,7 +5,7 @@ import { StringKeys } from './util';
 /**
  * @link https://github.com/types/lib-json-schema/blob/master/v4/index.d.ts
  */
-export type JsonSchemaTypes =  'array' | 'boolean' | 'integer' | 'number' | 'null' | 'object' | 'string' | (string & {}) | ['null', 'array'] | ['null', 'boolean'] | ['null', 'integer'] | ['null', 'number'] | ['null', 'object'] | ['null', 'string'];
+export type JsonSchemaTypes =  'array' | 'boolean' | 'integer' | 'number' | 'null' | 'object' | 'string' | (string & {});
 
 export type CompositePrimaryKey<RxDocType> = {
     /**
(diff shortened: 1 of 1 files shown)


pubkey:
I am sorry but this is wrong at many points.`(string & {})` is allowed which is not reflected in your array. Also there can be more then 2 types which is not reflected by that change. Also there is no test.

Notice that it is recommended to use `undefined` for optional fields, not a mixed type with `null`. By default all fields in json-schema are optional.

johannesjo:
Sorry for the imperfect code. Here is the minimal case to reproduce the underlying problem:
``` 
export const TEST_SCHEMA_LITERAL = {
  title: "test",
  version: 0,
  primaryKey: "id",
  type: "object",
  properties: {
    id: {
      type: "string",
    },
    completion_date: {
      type: ["string", "null"],
    },
  },
  required: [
    "id",
  ],
  indexes: [],
} as const;

const SCHEMA_TYPED = toTypedRxJsonSchema(TEST_SCHEMA_LITERAL);
export type RxSubtaskDocumentType = ExtractDocumentTypeFromTypedRxJsonSchema<
  typeof SCHEMA_TYPED
>;

export const TEST_SCHEMA: RxJsonSchema<RxSubtaskDocumentType> = TEST_SCHEMA_LITERAL;
//  => Throws TS Error
```
Which throws the error ` The type 'readonly ["string", "null"]' is 'readonly' and cannot be assigned to the mutable type 'JsonSchemaTypes[]'`:
```
TS2322: Type '{ readonly title: "test"; readonly version: 0; readonly primaryKey: "id"; readonly type: "object"; readonly properties: { readonly id: { readonly type: "string"; }; readonly completion_date: { readonly type: readonly ["string", "null"]; }; }; readonly required: readonly [...]; readonly indexes: readonly []; }' is not assignable to type 'RxJsonSchema<{ id: string; completion_date?: string | null | undefined; }>'.   The types of 'properties.completion_date.type' are incompatible between these types.     Type 'readonly ["string", "null"]' is not assignable to type 'JsonSchemaTypes | JsonSchemaTypes[] | undefined'.       The type 'readonly ["string", "null"]' is 'readonly' and cannot be assigned to the mutable type 'JsonSchemaTypes[]'
``` 

A possible workaround is casting the type:
```
["string", "null"] as JsonSchemaTypes[]
// or
["string", "null"] as string[]
```

pubkey:
Looks good now. Merged!

PR #4590, opened 2023-03-24
feature(replication-firestore): add filtered replication

KhashayarMirMS:
## This PR contains:
 - A NEW FEATURE

1. We only want to sync a subset of our Firestore database with each client.
2. We only want to sync a subset of client's database with Firestore.

As an example for case 1, we have a user system and want to add ownership concept to documents. In that case we would need to apply a filter/query when fetching data from Firestore.
For case 2, we want our users to be able to choose which documents are synced with the cloud. So we need to filter the local documents and only replicate a subset of them to Firestore.

Here's my proposal for these problems:

```ts
const replicationState = replicateFirestore(
    {
        collection: myRxCollection,
        firestore: {
            projectId,
            database: firestoreDatabase,
            collection: firestoreCollection
        },
        pull: {
            filter: [
                where('ownerId', '==', userId)
            ]
        },
        push: {
            filter: (item) => item.syncEnabled === true
        }
    }
);
```

You can use `pull.filter` and `push.filter` options to limit the data fetched from Firestore and sent to it.

## Todos 
- [x] Tests
- [x] Documentation
- [x] Typings
- [ ] Changelog

### DIFF
--- a/docs-src/replication-firestore.md
+++ b/docs-src/replication-firestore.md
@@ -106,3 +106,30 @@ allDocsResult.forEach(doc => {
 ```
 
 Also notice that if you do writes from non-RxDB applications, you have to keep these fields in sync. It is recommended to use the [Firestore triggers](https://firebase.google.com/docs/functions/firestore-events) to ensure that.
+
+## Filtered Replication
+
+You might need to replicate only a subset of your collection, either to or from Firestore. You can achieve this using `push.filter` and `pull.filter` options.
+
+```ts
+const replicationState = replicateFirestore(
+    {
+        collection: myRxCollection,
+        firestore: {
+            projectId,
+            database: firestoreDatabase,
+            collection: firestoreCollection
+        },
+        pull: {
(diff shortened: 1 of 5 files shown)


pubkey:
Great work. Merged!

chrisdrackett:
@KhashayarMirMS I'm new to Firestore, is there anything stopping a user from capturing requests and adjusting them to access other users data here?

KhashayarMirMS:
> is there anything stopping a user from capturing requests and adjusting them to access other users data here?

If you only filter the data on the client side, no there isn't. But you can define security rules in Firestore that specify who can access the data in each collection. You can read more about it [here](https://firebase.google.com/docs/firestore/security/get-started).
You should also note that these rules don't act as filters, but as gates to the data. So if you set a rule that allows users to access only a subset of the data and query the whole collection, you'll get an empty response. That's why you need to set both security rules and pull/push filters on your replication.

PR #4606, opened 2023-04-03
Export ReplicationOptions type

marceljuenemann:
Thanks for the great library! I just created [rxdb-supabase](https://github.com/marceljuenemann/rxdb-supabase) on top of it to replicate to supabase. 

The only problem I ran into was that the following types don't seem to be exported:
* ReplicationOptions
* ReplicationPullOptions
* ReplicationPushOptions

Of course, I could just duplicate all the fields in the Options type of my library, which arguably is not a terrible thing. But whenever you add a field to ReplicationOptions, it would require all downstream libraries to pass that field through as well. So I was wondering whether that's a deliberate design decision, or whether you want to add those types to the exports?

**NOTE:** I'm not saying this PR is the way to solve it, it probably exposes a lot of types that are implementation details. But it would be one way to solve my problem :)


pubkey:
Can you import them from `rxdb/plugins/replication` ? This should be possible.

marceljuenemann:
No, that gives me `Module '"rxdb/plugins/replication"' declares 'ReplicationOptions' locally, but it is not exported.ts(2459)`

I updated the PR to add the specific types to the list in index.ts, which also solves the issue.

PR #4612, opened 2023-04-07
FIX resume couchdb replication after connection error

davo22:
## This PR contains:
 - A BUGFIX

I am using couchDB replication in my project. I noticed that after an internet outage, the replication breaks down. Neither checkpoint iteration nor event observation works. To resume replication, the application had to be refreshed.

According to me, the cause of the event observation not working was an error issue on `pullStream$`. After throwing an error, the `Subject` goes into a `completed` state and stops emitting further values. Taking a cue from the `replication-firestore` plugin, I replaced the error emission on `pullStream$` with value emission on `replicationState.subjects.error` subject. This fixed the first problem.

After fixing the event observation, the device was still out of sync, because it did not know about the events emitted at the time of the outage. To fix this, I emit the `RESYNC` event at the time the error occurred.

In addition, I noticed that `retryTime` is not taken into account when the connection is broken and the application creates a huge number of requests, which unnecessarily increases the traffic. I fixed this by waiting for the time specified in the `retryTime` before the next connection attempt.

### DIFF
--- a/src/plugins/replication-couchdb/index.ts
+++ b/src/plugins/replication-couchdb/index.ts
@@ -47,6 +47,8 @@ import {
 export * from './couchdb-helper';
 export * from './couchdb-types';
 
+const DEFAULT_RETRY_TIME =  1000 * 5;
+
 export class RxCouchDBReplicationState<RxDocType> extends RxReplicationState<RxDocType, CouchDBCheckpointType> {
     constructor(
         public readonly url: string,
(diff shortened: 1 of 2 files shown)


pubkey:
Hi @davo22 
The CI fails, please check.

pubkey:
Please run the tests locally.

pubkey:
@davo22 the CI is still red, can you check?

pubkey:
Closed since no answer and CI is failing.

pubkey:
@davo22 CI still red, please check.

pubkey:
I fxed the retry-time waiting here https://github.com/pubkey/rxdb/commit/d8c6d3e9ba7d79df82072791d71ad439e9b2681a

For everything else pls fix the ci run first.

PR #4614, opened 2023-04-11
Fix default value of eventReduce option in documentation

russelldavis:
## This PR contains:
- Improved docs

The actual default value in the code is `false`, but docs had it as `true`


pubkey:
The default should be true, let me check.

pubkey:
I added a todo so this will be changed in the next major version as it should have been before.
Thank you for finding this.

PR #4630, opened 2023-04-13
Move z-schema to peer dependencies.

troywweber7:
Listing z-schema as an optional peer dependency has 2 benefits:
1. Users who don't use z-schema validation don't need this dependency.
2. Pnpm won't hoist z-schema to the top level node_modules; pnpm users who wish to register a schema format validation function are now protected from accidentally installing the incorrect version.

## This PR contains
 - SOMETHING ELSE

Pnpm won't hoist z-schema to the top level node_modules. As a result, pnpm users who wish to register a schema format validation should explicitly install z-schema. In doing so, they might install the wrong version, and then registering a new validation format will fail.

## Todos
- [ ] Documentation
- [ ] Changelog

We may need to document that if someone wants to use the z-schema wrapper function, they should first install z-schema. The changelog may also need to be updated.

It is not within the scope of this pull request, but a similar change could be made for other plugin dependencies such as AJV.

### DIFF
--- a/package.json
+++ b/package.json
@@ -365,7 +365,13 @@
     "lint"
   ],
   "peerDependencies": {
-    "rxjs": "^7.8.0"
+    "rxjs": "^7.8.0",
+    "z-schema": "5.0.5"
+  },
+  "peerDependenciesMeta": {
+    "z-schema": {
+      "optional": true
+    }
   },
   "dependencies": {
     "@babel/runtime": "7.21.0",
@@ -402,8 +408,7 @@
     "socket.io-client": "4.6.1",
     "unload": "2.4.1",
     "util": "0.12.5",
-    "ws": "8.13.0",
-    "z-schema": "5.0.5"
+    "ws": "8.13.0"
   },
   "devDependencies": {
     "@babel/cli": "7.21.0",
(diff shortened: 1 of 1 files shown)


pubkey:
This would be a breaking change.
Also it looks like this is more a pnpm problem then a rxdb problem. Because following that logic would mean we should move next to all dependencies into the peer dependencies.

pubkey:
To be honest, a big benefit of node.js compared to java (and others) was that subdependencies do not break each other when they are needed in different versions.
If you use a package manager like pnpm that re-introduces that problem, you need a different strategy to handle these kind of problems.

Having a scope `@rxdb/` creates different problems like when you use a dependency bot etc. also it adds complexity and I do not want to spend my time on maintaining that.

pubkey:
Deleted my previous comment because I missunderstood the problem.

So moving all plugins into a `@rxdb/` scope was proposed before. I already registered that scope on npm. But I do not want to spend the timing on migrating that, you can make a PR for that if you want.

RxDB could export a method that returns the `ZSchema` object so you can use that to get the same instance as RxDB will be using.

troywweber7:
@pubkey I like the idea of exporting a method which provides access to the z-schema object for configuration. We should reject this PR in favor of that solution.

PR #4686, opened 2023-04-23
Add better support for `as const` schemas

gtarsia:
## This PR contains:
IMPROVED typings

From combining the "typescript" and "composite primary key" sections, I got this error:
```
        Types of property 'fields' are incompatible.
          The type 'readonly ["app_id", "budget_id"]' is 'readonly' and cannot be assigned to the mutable type 'string[]'.
```

This error happens because adding the const modifier makes arrays be readonly, which have a more restrictive typechecking than readonly props.
I was looking to push a pull request, but I wasn't sure if this was a proper solution (it did fix the issue for me and I imagine it is backwards compat, although I'm not a typescript expert)
```
export type CompositePrimaryKey<RxDocType> = {
    ...
    fields: readonly (StringKeys<RxDocType> | string)[] | (StringKeys<RxDocType> | string);
    ...
};
I also added `readonly` modifier to other props inside the RxJsonSchema.

### DIFF
--- a/src/types/rx-schema.d.ts
+++ b/src/types/rx-schema.d.ts
@@ -19,7 +19,7 @@ export type CompositePrimaryKey<RxDocType> = {
      * the fields must be required and final
      * and have the type number, int, or string.
      */
-    fields: (StringKeys<RxDocType> | string)[];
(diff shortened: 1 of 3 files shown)


pubkey:
Thank you, merged.

PR #4698, opened 2023-04-26
Add test for firestore replication bug

chrisdrackett:
When adding items quickly while using firestore replication, only the first change makes it onto the server. Also, in this case changes from the server don't automatically make it back to the client.

### DIFF
--- a/test/unit/bug-report.test.ts
+++ b/test/unit/bug-report.test.ts
@@ -9,13 +9,16 @@
  * - 'npm run test:browser' so it runs in the browser
  */
 import assert from 'assert';
-import AsyncTestUtil from 'async-test-util';
 import config from './config';
 
 import {
     createRxDatabase,
     randomCouchString
 } from '../../';
+import { collection, doc, getDoc, getFirestore } from 'firebase/firestore';
+import { initializeApp } from 'firebase/app';
+import { getAuth, signInAnonymously } from 'firebase/auth';
+import { replicateFirestore } from '../../plugins/replication-firestore';
 
 describe('bug-report.test.js', () => {
     it('should fail because it reproduces the bug', async () => {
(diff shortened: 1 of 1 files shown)


pubkey:
Can you reproduce the problem with a non-firestore replication?

chrisdrackett:
@pubkey hm, I just tried this in a graphQL replication test and I'm pretty sure I'm seeing the same thing. I feel like I'm stumbling around these tests a bit, so here is the code I used:

```
                await c.insert(schemaObjects.humanWithTimestamp({id: 'a123456', name: '1'}));

                await waitUntil(() => pushCount === 1);

                await c.incrementalUpsert({id: 'a123456', name: '2'})
                await c.incrementalUpsert({id: 'a123456', name: '3'})

                await waitUntil(() => pushCount === 3);

                const docsOnServer = server.getDocuments()[0];
                assert.strictEqual(docsOnServer.name, '3');
```

in this the assert is getting `1` vs. the expected `3`.

```
AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:

'1' !== '3'
```

chrisdrackett:
@pubkey curious if you've been able to replicate this issue. I'm attempting to work around it or figure out what is going on. I want to make sure this is an RxDB issue and not something I'm doing in my code (mostly surprised I'm the first to run into this)

pubkey:
No I had no time working on this.
The first step would be to reproduce this with the plain non-firebase replication-protocol to ensure that this is not a firebase issue.

pubkey:
Can you make a PR with a failing test that uses graphql instead of firestore?

pubkey:
Ok, this helps to understand.
Please notice that firestore replication is on beta, so I will likely not fix that by myself.

chrisdrackett:
@pubkey oh, is there no plan for official firestore support? I figured the beta tag was because it was still in active development and not ready for production. Is there anyone else I can ping about this?

pubkey:
There is a plan for official support. But it is an open source project, so it is assumed that the community fixes the bugs, not solely the maintainer.

chrisdrackett:
totally get it, I spent a little over a day trying to fix this but didn't get very far as the way replication works in RxDB is over my head :D

pubkey:
I am closing this. I cannot fix it without reproduction. Please make a PR with a failing test case.

chrisdrackett:
@pubkey this pr is a failing test case?

pubkey:
@chrisdrackett I am so sorry. You are right. Reopened.

pubkey:
I think in your test you forgot to run `await replicationState.awaitInSync();` after the last document update. At the time you check the correctness of the server docs, you cannot assume that the replication is done already. It might be still transfering data.

https://github.com/pubkey/rxdb/pull/5240

pubkey:
@chrisdrackett  can you check if this fixes your problem?

chrisdrackett:
@pubkey I think this did it. I'm also going to submit a PR to update the docs for `incrementalUpsert()` as that is where I got the non-awaited code.

PR #4701, opened 2023-04-26
add note about `getDatabaseConnection`

### LINKED ISSUE #4700: update `getSQLiteBasicsQuickSQLite` to take an open database vs. the `openDatabase` function
right now `getSQLiteBasicsQuickSQLite` from `rxdb-premium` takes the `openDatabase` function from `react-native-quick-sqlite` as its input. I assume it then opens the database internally.

The downside here is if other code needs to use the database for whatever reason (for me, currently, its debugging) you need to open a second connection. In my experience this works only until both connections try and access the DB at the same time, and then everything falls apart.

I'm wondering if this could be fixed by changing the input for `getSQLiteBasicsQuickSQLite`

so go from:

```
// create database
const myRxDatabase = await createRxDatabase({
    name: 'exampledb',
    multiInstance: false, // <- Set multiInstance to false when using RxDB in React Native
    storage: getRxStorageSQLite({
        sqliteBasics: getSQLiteBasicsQuickSQLite(open)
    })
});
```

to: 

```
// open database
const db = open('databaseName')

// create database
const myRxDatabase = await createRxDatabase({
    name: 'exampledb',
    multiInstance: false, // <- Set multiInstance to false when using RxDB in React Native
    storage: getRxStorageSQLite({
        sqliteBasics: getSQLiteBasicsQuickSQLite(db)
    })
});
```

chrisdrackett:
adding info about `getDatabaseConnection`

fixes #4700

### DIFF
--- a/docs-src/rx-storage-sqlite.md
+++ b/docs-src/rx-storage-sqlite.md
@@ -142,6 +142,22 @@ const myRxDatabase = await createRxDatabase({
 });
 ```
 
+## Database Connection
+
+If you need to access the database connection for any reason you can use `getDatabaseConnection` to do so:
+
+```ts
+import { getDatabaseConnection } from 'rxdb-premium/plugins/storage-sqlite'
+```
+
+It has the following signiture:
+
+```ts
+getDatabaseConnection(
+    sqliteBasics: SQLiteBasics<any>,
+    databaseName: string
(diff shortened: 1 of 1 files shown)


pubkey:
Thank you, merged.

PR #4716, opened 2023-04-29
Fix pushing null document with custom deleted field

Poiuytrezay1:
## This PR contains:
 - A BUGFIX

As per documentation, with a push modifier: "Returning null will skip the document."

However, if the deleted field is not the default one, it will call this method and return "newDocumentState: { CUSTOM_FIELD_FOR_DELETED: false/true }" instead of "newDocumentState: null". This cause a failure for the push afterwards.

## Todos 
- [ ] Tests
- [ ] Documentation
- [ ] Typings
- [ ] Changelog

### DIFF
--- a/src/plugins/replication/replication-helper.ts
+++ b/src/plugins/replication/replication-helper.ts
@@ -13,7 +13,7 @@ export function swapDefaultDeletedTodeletedField<RxDocType>(
     deletedField: string,
     doc: WithDeleted<RxDocType>
 ): RxDocType {
-    if (deletedField === '_deleted') {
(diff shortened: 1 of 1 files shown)


pubkey:
Change looks ok. Pls notice that each change needs at least one unit test.

pubkey:
Any update on this?

Poiuytrezay1:
After checking around, the original bug is not completely fixed (the "Returning null will skip the document" not working). This needs more work, and I need to understand how the replication works in detail to not destroy anything.

PR #4741, opened 2023-05-04
Fix checkSchema for anyOf and items array

swnf:
## This PR contains:
 - IMPROVED TESTS
 - A BUGFIX

The dev mode plugin throws `"fieldnames do not match the regex"` for `fieldName:"0"` if the schema contains `anyOf` or an `items` array.

My patch seems to break the (negative) test for this schema. I'm not sure how this test was supposed to work. It seems like it only worked before because `checkSchema` had bugs with arrays.

https://github.com/pubkey/rxdb/blob/397ecf2a30fb8b58c705bd963d16aac6db0a1aff/test/helper/schemas.ts#L625-L650

### DIFF
--- a/src/plugins/dev-mode/check-schema.ts
+++ b/src/plugins/dev-mode/check-schema.ts
@@ -175,7 +175,7 @@ export function validateFieldsDeep(rxJsonSchema: RxJsonSchema<any>): true {
                 !currentObj.properties &&
                 schemaObj &&
                 typeof schemaObj === 'object' &&
-                !currentPath.endsWith('oneOf')
+                !Array.isArray(currentObj)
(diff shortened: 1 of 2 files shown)


pubkey:
This breaks the CI, please check.

pubkey:
Hi @swnf 
I checked and you are right. These tests are outdated. I removed them in https://github.com/pubkey/rxdb/commit/63f6b016a53c42721b0315ee54ece36ef6846989

swnf:
Thank you @pubkey. I have rebased my PR, the CI checks now pass successfully.

pubkey:
Awesome, merged.

PR #4745, opened 2023-05-17
Update before-next-major.md

GabrielDelepine:
## This PR contains:
 - SOMETHING ELSE

Incorrect / confused naming 

`myRxReplicationState.send$.subscribe` works only if the sending is successful. Therefore, it should be named `sent$`, not `send$`.

Interestingly, `received$` has been named correctly


pubkey:
Yes you are right, merged.

PR #4749, opened 2023-05-25
Fix graphql replication docu

Jeremy-Hill-02:
## This PR contains:

 - IMPROVED DOCS

Some ternaries from the `pullHuman` resolver in the GraphQL replication documentation are incorrect.
Colons are where question marks should be.


pubkey:
Thank you, merged.

PR #4752, opened 2023-05-29
Fix prevent delete _rev from row.doc

gigasource:
remoteRevById.set(row.id, row.doc._rev);
doc._rev should not be undefined


pubkey:
Thanks, merged.

Issue #4755, opened 2023-06-02
count() is incorrect


adam-lynch:
In our web app, we do this:

```js
const rxQuery = collection.count({
    selector: {
      id: {
        $eq: 'persistentStateBlob',
      },
    },
  });
const result = await.exec();
```

Expected: `result` is `0`.
Actual: `result` is `1`.

In fact, if I use the following code instead...

```js
const rxQuery = collection.find({
    selector: {
      id: {
        $eq: 'persistentStateBlob',
      },
    },
  });
const result = (await.exec()).length;
```

... `result` is `0`. (That's the exact same query)

## Data

The collection does contain one item with an ID of `persistentStateBlob2`.

## Schema

```js
{
    primaryKey: 'id',
    properties: {
        id: { type: 'string', maxLength: 100 },
        value: { type: 'string' },
    },
    required: ['value'],
    type: 'object',
    version: 0,
}
```

## Versions and storage

- rxdb: 14.11.1
- rxdb-premium: 14.11.1
- storage: premium IndexedDB storage via shared storage worker

### With our app

- Chromium and Firefox browsers.
- Disabling `eventReduce`.
- Disabling `multiInstance`.
- Removing the key compression plugin (this collection doesn't use it anyway).
- Disabling the shared storage worker wrapper.
- Using the memory storage (without the shared storage worker).
- Deleting the IndexedDB databases in browser devtools.

All behave the same.

### RxDB tests

- I created a test in `test/unit/bug-report.test.ts` and ran it with node+dexie; it passed.
- I changed the `count matching only` test in `test/unit/rx-collection.test.ts` and ran it with node+dexie; it passed.

## Other notes

- `count(...)` (with a selector) correctly returns `0` if the collection is empty.
- `count()` without any argument seems correct.

pubkey:
We need a PR with a test for that. This would make it easy to check if this is a dexie.js problem or if other storages are also affected.

adam-lynch:
#4775

adam-lynch:
keep open

adam-lynch:
This is still an issue as far as I know. I can't today but I'll test again with the latest version when I get a chance

pubkey:
Hi @adam-lynch 
I have reproduced and analyzed the problem.
I am working on a fix, but it will take some time and very likely it will require a major RxDB release.

pubkey:
Fix is released, please check the newest RxDB version.

adam-lynch:
[...]

PR #4775, opened 2023-06-12
Create count with selector test

adam-lynch:
A test to reproduce #4755, but as I said there, it actually passes.


pubkey:
Fixed here https://github.com/pubkey/rxdb/pull/4825

adam-lynch:
I can run my tests again once rxdb-premium@14.15.0 is released

PR #4827, opened 2023-07-18
`replicationState.active$` observable does not emit on replication failure

aiven715:
Reproducing a bug where `replicationState.active$` observable does not emit a `false` value on replication failure.

## This PR contains:
- REPRODUCED BUG

According to the docs, `replicationState.active$` observable should emit a `false` value when the replication cycle is not running. But when replication is failed (when either pull or push handlers throw an error), it does not emit any value, leaving the observable with the `true` value last emitted.

It might be a change, introduced in the new replication protocol since that issue does not exist in the RxDB@12

### DIFF
--- a/examples/graphql/client/index.js
+++ b/examples/graphql/client/index.js
@@ -213,6 +213,9 @@ async function run() {
             console.error('replication error:');
             console.dir(err);
         });
+        replicationState.active$.subscribe(active => {
+            console.log(`${db.hero.name} replication active: ${active}`);
+        })
     }
 
 
--- a/examples/graphql/server/index.js
+++ b/examples/graphql/server/index.js
@@ -85,6 +85,8 @@ export async function run() {
     // The root provides a resolver function for each API endpoint
     const root = {
         pullHero: (args, request) => {
+            throw new Error('server failure')
+
             log('## pullHero()');
             log(args);
             authenticateRequest(request);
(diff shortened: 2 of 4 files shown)


pubkey:
Hi @aiven715 
Can you make a test in the test suite, so that the CI run proves that something is wrong?
I recommend to edit this file https://github.com/pubkey/rxdb/blob/master/test/unit/bug-report.test.ts

aiven715:
Hey @pubkey 

I moved a test to the mentioned file

aiven715:
@pubkey any update on that issue? Thanks

pubkey:
> According to the docs, replicationState.active$ observable should emit a false value when the replication cycle is not running.

"Replication cycle is running" means that the client is trying to either pull or push data. It will only emit "false" when all data is replicated in both directions. The current behavior is correct, I think we should clarify that in the docs.

aiven715:
> "Replication cycle is running" means that the client is trying to either pull or push data.

In the described case (when the push handler throws an error), RxDB neither pulls nor pushes, so I believe having a `true` value in the `active$` observable for that case looks confusing.

pubkey:
In the described case, the replication cycle is still running. It is in the state of awaiting the retry time so that it can push again. The replication is active.

pubkey:
I am closing this.
As workaround you can wrap the pul/push handlers with your own Subjects.next() calls to observe if they are running or not.

PR #4828, opened 2023-07-18
Add a feature request to control replication start on offline → online change

aiven715:
This is a feature request to prevent automatic re-sync when detecting a switch from offline to online mode via the `navigator.onLine`.

It would be helpful for applications that require full control over when the synchronization should be triggered in such cases.

## This PR contains:
- A FEATURE REQUEST

In RxDB@12, our application had full control over the replication cycle during the offline → online switch.

But, during the migration to RxDB@14, we encountered a breaking change where replication started automatically when the user went online, which affected our logic.

### DIFF
--- a/src/types/plugins/replication.d.ts
+++ b/src/types/plugins/replication.d.ts
@@ -134,11 +134,15 @@ export type ReplicationOptions<RxDocType, CheckpointType> = {
     /**
      * Time in milliseconds after when a failed backend request
      * has to be retried.
-     * This time will be skipped if a offline->online switch is detected
-     * via `navigator.onLine`
      * @default 5000
      */
     retryTime?: number;
+    /**
+     * When this is set to true, the retryTime setting will be skipped if offline->online
+     * switch is detected via `navigator.onLine` and replication cycle will be started immediately.
+     * @default true
+     */
+    retryOnOnline?: boolean;
     /**
      * When multiInstance is `true`, like when you use RxDB in multiple browser tabs,
      * the replication should always run in only one of the open browser tabs.


pubkey:
`navigator.onLine` is only used as an optimization at the `awaitRetry()` method. I do not see a real use case for not skipping the retry time and instead fully await it, especially when it is known that the client was offline before.
Can you describe your exact use case?

aiven715:
@pubkey it produces a side-effect in the form of replication and makes retry/recovery logic less predictable and harder to test.

We would like to fully handover control of replication to the app and do not want RxDB to perform auto-retry.

According to the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/onLine):
> In Chrome and Safari, if the browser is not able to connect to a local area network (LAN) or a router, it is offline; all other conditions return `true`.

In our application, “offline” means unable to talk to our back-end. When `navigator.isOnline === false`, it doesn't always mean the application is offline.

User can be disconnected from the internet, but still able to connect to the back-end via localhost, meaning it's online from the perspective of our app.

aiven715:
@pubkey, just a gentle reminder to take a look at this. Thanks!

pubkey:
> In our application, “offline” means unable to talk to our back-end. When navigator.isOnline === false, it doesn't always mean the application is offline.

Yes, it does not always mean that the client is connected to the replication server. This is why navigator.isOnline is only use as an optimziation shortcut so that it will skip the retry time and directly retry.

I do not understand in which test case it is helpfull to deactivate this shortcut.

jakubroztocil:
> I do not understand in which test case it is helpfull to deactivate this shortcut.

@pubkey the app we chose to use RxDB for requires a completely custom logic for retries. Our implementation is robust, provides detailed feedback to the user, and has an extensive test suite. That requires that we have full control over the retries.

If we throw in RxDB automatically retrying based on the computer’s online status — which has no relation to what our app considers as having the backend available and so is pretty much an unrelated random event— as another variable impacting the behavior of the system, it immensely complicates it, and forces us to invest a lot of resources into keeping it fully controllable and testable by finding workarounds for that behavior. 

For example, it prevents us from testing that our app correctly schedules a retry if RxDB triggers it immediately based on some unrelated external resource (the internet) happening to be currently available.

pubkey:
> the app we chose to use RxDB for requires a completely custom logic for retries

If you need a complete custom behavior, please fork the replication plugin or patch the pull/push method. That would be two lines of code.

jakubroztocil:
Regardless of our specific needs, the hard coding of un-mockable external resources like this is not a good practice.

PR #4851, opened 2023-07-31
Feature request: isMangoQuerySatisfiedByIndex

adam-lynch:
## This PR contains:

A feature request. The code changes are just to show different options for where the function could live / what the signature could be.

From reading the updated [`rxQuery.count` docs](https://rxdb.info/rx-query.html#count), my understanding is that we should ideally only use `.count` if it's satisfied by an index, otherwise we should use `.find` and count the amount of RxDocument results.

In codebases like mine, I can't predict every query. I'd like to be able to do the optimal operation based on what the input is. This is why I'm proposing a function which receives a `MangoQuery` and returns a boolean indicating whether the `MangoQuery` is satisfied by an index or not.

## Ideas

As you can see from the code changes, I think it could be any of the following:

1. `rxCollection.isMangoQuerySatisfiedByIndex(mangoQuery)`
2. `isMangoQuerySatisfiedByIndex(collectionName, mangoQuery)` (standalone util)
3. `isMangoQuerySatisfiedByIndex(rxSchema, mangoQuery)` (standalone util)
4. `rxQuery.isSatisfiedByIndex()`

I don't have a preference except that number 4 is a little cumbersome because if the result is `false`, we would need to create a new `RxQuery`. It's an extra step, plus creating two `RxQuery`s instead of one would involve some unnecessary work under the hood I assume.

### Specific index

Side note: an index could be included in the `MangoQuery`. If given, it would check against that index; if omitted, it would check against all indexes.

## Dev mode

I found that the dev mode plugin exports this:

```
export function areSelectorsSatisfiedByIndex<RxDocType>(
    schema: RxJsonSchema<RxDocumentData<RxDocType>>,
    query: FilledMangoQuery<RxDocType>
): boolean {
    // ...
}
```

This doesn't cover it for two reasons:

1. It requires a `FilledMangoQuery` instead of a `MangoQuery`. I don't know how to fill one; I assume there is an internal function for this.
2. I want to use this in production and I won't be using the dev-mode plugin in production.

## Should users have to worry about this?

I also wonder if this is a problem I should have to worry about at all. Maybe RxDB's `.count` should do this under the hood and return me a number.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -3,6 +3,8 @@
  * for when the you call "import from 'rxdb'".
  */
 
+import { MangoQuery, RxDocumentData, RxJsonSchema } from './types';
+
 export { addRxPlugin } from './plugin';
 
 export * from './rx-database';
@@ -28,6 +30,20 @@ export * from './plugin-helpers';
 export * from './plugins/utils';
 export * from './hooks';
 export * from './query-cache';
+
+export function isMangoQuerySatisfiedByIndex<RxDocType>(
+    collectionName: string, // There is probably a better type for this
+    mangoQuery: MangoQuery<RxDocType>
+): boolean {
+
+}
+
+export function isMangoQuerySatisfiedByIndex2<RxDocType>(
+    schema: RxJsonSchema<RxDocumentData<RxDocType>>,
+    mangoQuery: MangoQuery<RxDocType>
+): boolean {
+
+}
 
 /**
  * TODO use export type * from './types';
--- a/src/rx-collection.ts
+++ b/src/rx-collection.ts
@@ -279,6 +279,10 @@ export class RxCollectionBase<
         return this.getDataMigrator().migratePromise(batchSize);
     }
 
+    isMangoQuerySatisfiedByIndex(mangoQuery: MangoQuery<RxDocumentType>): boolean {
+
+    }
+
     async insert(
         json: RxDocumentType | RxDocument
     ): Promise<RxDocument<RxDocumentType, OrmMethods>> {
--- a/src/rx-query.ts
+++ b/src/rx-query.ts
@@ -340,7 +340,9 @@ export class RxQueryBase<
             });
     }
 
+    isSatisfiedByIndex(): boolean {
 
+    }
 
     /**
      * cached call to get the queryMatcher


pubkey:
Some thoughts I have about your proposal:

You are assuming that all RxStorage implementations use the RxDB query planner. But this  is not true and should not be required. For example the lokijs storage does not use it and in the past the pouchdb storage did not use it. Also in the future we might have a mongodb RxStorage which will use the mongodb internal query planner.
Therefore checking `isMangoQuerySatisfiedByIndex` is not always possible.

The benefit of using .find() for count-queries that need to fetch the doucments anyway (aka slow-count), is only that the document data is written into the RxDB document cache and can be reused. But there can be use cases where this is not wanted. For example if the storage runs in a webworker, you do not always want to send the document-data between the worker and the main thread, you might only need the count-result instead to save performance.

adam-lynch:
> You are assuming that all RxStorage implementations use the RxDB query planner. But this is not true and should not be required. For example the lokijs storage does not use it and in the past the pouchdb storage did not use it. Also in the future we might have a mongodb RxStorage which will use the mongodb internal query planner.
Therefore checking isMangoQuerySatisfiedByIndex is not always possible.

This made me think that having RxDB figure it out internally made even more sense....

> But there can be use cases where this is not wanted

...but this is a problem alright. I assume there are even more cases than you say. If it was figured out by RxDB internally, there would have to be a parameter to override it. But I understand if you just want to close this.

> For example if the storage runs in a webworker, you do not always want to send the document-data between the worker and the main thread, you might only need the count-result instead to save performance.

This kind of thing should be in the docs IMO. If I hadn't created this PR, I would've refactored our app to use `.find` everywhere, but we use a shared worker on one platform so that would've been bad. (This is based on reading the `rxQuery.count` docs, if it's somewhere else I missed it)

adam-lynch:
Thanks, I'll close this.

I'm going to think out loud here for a sec and any advice you have would be welcome...

So to recap, for an app like mine with different platforms, storages, and no guarantee of knowing every query:

- Using remote storage?
  - Yes.
    Could there be a lot of matching documents?
    - Yes. 
      Do we know at write/dev-time that a particular query is satisfied by an index?
      - Yes: use `.count` (and `allowSlowCount: true` to be safe), *unless* we know that it would be worth it to use the cache (read from it or write to it for later) despite the fact a lot of results would be sent from remote storage to main.
      - No: use `.find`.
    - No: use `.find`.
  - No: use `.find` and `allowSlowCount: false`.
  
> Could there be a lot of matching documents?

Putting aside not knowing what "a lot" is,  it could be a tough call to say how many docs could be matched.

Maybe I'll do this: if non-remote storage, always use `.find` (and `allowSlowCount: false`). If remote storage, default to using `.count` (and `allowSlowCount: true`) unless we know a particular query / call becomes an issue, then we can use `.find` for that call.

Or I could just use `.count` (and `allowSlowCount: true`) until we feel this entire question is definitely worth worrying about 🤔

pubkey:
You should use `.count` with `allowSlowCount: true`.

Filling the document cache  (by using .find) will only benefit you if you do a find-by-id afterwards on a document that is in the cache.

Normal `.find` will call the storage anyway which will return the document data again so there is no real performance benefit if you have filled the cache before.

adam-lynch:
Thank you! That's particularly good to know (cache only benefitting find-by-id)

Issue #4877, opened 2023-08-11
Multiple calls to requestIdlePromise end up in a serial execution


costingeana:
RxDB Version: 14.15.1

Hello,
I just observed the changes applied on the `requestIdlePromise` function ( #4813 ) have some undesired effects. I'll explain below.

The previous version of `requestIdlePromise` returned a new `Promise` for each function's call; in this case, the consumer's code would wait only for the resolving of this sole `Promise`.

On the other hand, the current version returns indeed a new `Promise` for each `requestIdlePromise` function's call, BUT the consumer's code will have to wait for the resolving of the **entire Promise's chain** because the `requestIdlePromise` function returns the `idlePromiseQueue` itself. 
This ends up being similar to a serial execution, and the effect is the degradation of the performance in certain scenarios.

For example, I created 100 RxDB databases, and then I dispose them (dbInstance.remove()). Before the changes from #4813 (i.e. version = 14.14.1) the time spent to dispose all the databases was T. Now, after these changes (i.e. version >= 14.14.2) the time is about 10 x T.

IMHO, the previous implementation of the idle promises' queue was correct, so I altered it below to include the fix for the initial problem, namely the one related to the `requestIdleCallback`:

```
/**
 * If multiple operations wait for an requestIdlePromise
 * we do not want them to resolve all at the same time.
 * So we have to queue the calls.
 */
var idlePromiseQueue = PROMISE_RESOLVE_VOID;
export function requestIdlePromise(timeout = undefined) {
  return new Promise(res => {
    idlePromiseQueue = idlePromiseQueue.then(() => {
      /**
       * Do not use window.requestIdleCallback
       * because some javascript runtimes like react-native,
       * do not have a window object, but still have a global
       * requestIdleCallback function.
       * @link https://github.com/pubkey/rxdb/issues/4804
       */
      if (typeof requestIdleCallback === 'function') {
        requestIdleCallback(() => res(), {
          timeout
        });
      } else {
        promiseWait(timeout).then(res)
      }
    });
  });
}
```
Thank you.

_Originally posted by @costingeana in https://github.com/pubkey/rxdb/issues/4804#issuecomment-1673461967_

pubkey:
I am not sure if I understood it correctly.
The returned `idlePromiseQueue` object each time is not the same on each function call because it will be reasinged with `idlePromiseQueue = idlePromiseQueue.then...`

From the code docs
```js
/**
 * If multiple operations wait for an requestIdlePromise
[...]

PR #4884, opened 2023-08-16
Implement an IdleQueue that executes the tasks during the idleness of the main thread (see requestIdleCallback)

costingeana:
1. Implement an `IdleQueue` similar to the one exposed from the `custom-idle-queue` package. 
The difference is that all the queued items are executed as background tasks (i.e. when the main thread is idle) due to the usage of the `requestIdleCallback` function. I want to highlight that more than one background task may be executed during the idleness of the main thread (a.k.a the idle period); the remaining tasks are executed on the next idle period (so, when the main thread is idle again) and so on.
2. Replace the usages of the `requestIdlePromise` from `utils/utils-promise` with `IdleQueue.requestIdlePromise`; the replacements were done especially in the files related to the  lokijs rx storage.
3. For the moment, I wrap the functions (`requestIdlePromise`, `cancelIdlePromise`, `clear`) in a object (namely, the `IdleQueue`) to avoid the collisions with the current code (see the `requestIdlePromise` from `utils/utils-promise`). This may change if you consider this pull requests is useful.

## This PR contains:
 - A NEW FEATURE

#4877

### DIFF
--- a/src/plugins/storage-lokijs/loki-save-queue.ts
+++ b/src/plugins/storage-lokijs/loki-save-queue.ts
@@ -1,7 +1,7 @@
 import type { LokiDatabaseSettings } from '../../types';
 import {
     PROMISE_RESOLVE_VOID,
-    requestIdlePromise
+    IdleQueue
 } from '../utils';
 
 /**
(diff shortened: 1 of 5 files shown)


pubkey:
>  I want to highlight that more than one background task may be executed during the idleness of the main thread

I do not want this. RxDB is a database for UI apps and the purpose for the idle queue is to make it as less likely as possible for background tasks to affect the responsiveness of the application for the user.

pubkey:
Is this PR still relevant?
Which value to RxDB do these changes provide? Because from my point of view this just makes the code more complex.

pubkey:
Thank you for clarifying.

PR #4899, opened 2023-08-23
Migration of assumed master documents

aiven715:
## This PR contains:
 - A NEW FEATURE
 
Here's a pretty rough implementation of the assumed masters' migration feature.

I would like to validate that the direction I'm moving is correct and get feedback on the further steps.

Note, at the point the code is pretty dirty for the sake of simplicity, it would be cleaned up later once we'll agree on the direction.

### DIFF
--- a/src/plugins/migration/data-migrator.ts
+++ b/src/plugins/migration/data-migrator.ts
@@ -44,7 +44,8 @@ import type {
     RxDocumentData,
     RxStorageInstanceCreationParams,
     InternalStoreCollectionDocType,
-    RxStorageInstance
+    RxStorageInstance,
+    RxDocument
 } from '../../types';
 import {
     RxSchema,
@@ -66,8 +67,8 @@ import {
 import { normalizeMangoQuery } from '../../rx-query-helper';
 import { overwritable } from '../../overwritable';
 import { MIGRATION_DEFAULT_BATCH_SIZE } from './migration-helpers';
-import { getComposedPrimaryKeyOfDocumentData } from '../../rx-schema-helper';
 import { docStateToWriteDoc } from '../../replication-protocol';
+import { RX_REPLICATION_COLLECTION_FLAG } from '../replication/replication-helper';
 
 export class DataMigrator {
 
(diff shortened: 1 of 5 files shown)


src/plugins/migration/data-migrator.ts
```diff
@@ -457,6 +460,49 @@ export async function _migrateDocuments<RxDocType>(
...
+        database.storage.statics,
+        database.internalStore
+    );
+    const oldCollectionMetaDoc = allCollectionMetaDocs.find(
+        doc => doc.data.name === oldCollection.newestCollection.name &&
+        doc.data.version === oldCollection.version
+    );
+    const replicationStorageSpec = oldCollectionMetaDoc?.data.connectedStorages.find(s => (
+        // TODO: find a better way to identify replication storage
+        s.collectionName.includes('replication')
```
pubkey, line 475: The code that creates the meta collection name is:

```const metaInstanceCollectionName = this.collection.name + '-rx-replication-' + this.replicationIdentifierHash;```

We should extract `const RX_REPLICATION_COLLECTION_FLAG='rx-collection'` into a variable and then you can do ` s.collectionName.includes(RX_REPLICATION_COLLECTION_FLAG)`

Comment (aiven715) [reply src/plugins/migration/data-migrator.ts]: Sounds good, added it

File: src/plugins/migration/data-migrator.ts [hunk 2]
```diff
@@ -457,6 +460,49 @@ export async function _migrateDocuments<RxDocType>(
      */
     const previousDocumentData = clone(documentsData);
 
+
+    const database = oldCollection.database;
+    const allCollectionMetaDocs = await getAllCollectionDocuments(
```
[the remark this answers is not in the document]
aiven715: Oh right, that makes sense. I moved migration of the assumed masters before the fork documents

aiven715:
@pubkey I've made changes according to your comments, please take a look. I've realized that tests fail for Dexie and some other storages. For some reason, `assumedMasterState` is undefined in the push handler. Do you know why it could happen?

aiven715:
@pubkey just a gentle reminder to look into that 👀

Issue #4948, opened 2023-09-19
Issues getting data when object field has `additionalProperties: true`


adam-lynch:
Related but separate: #4926

adam-lynch:
I forgot to link to the PR; #4949

PR #4949, opened 2023-09-19
Test for #4948 (RxDocument.get() on additonalProperty)

adam-lynch:
## This PR contains:
 - IMPROVED TESTS

This contains a test for #4948

### DIFF
--- a/test/unit/bug-report.test.ts
+++ b/test/unit/bug-report.test.ts
@@ -9,13 +9,13 @@
  * - 'npm run test:browser' so it runs in the browser
  */
 import assert from 'assert';
-import AsyncTestUtil from 'async-test-util';
 import config from './config';
 
 import {
     createRxDatabase,
     randomCouchString
-} from '../../plugins/core';
+} from '../../';
+import { RxJsonSchema } from '../../src';
 
 describe('bug-report.test.js', () => {
     it('should fail because it reproduces the bug', async () => {
@@ -36,7 +36,7 @@ describe('bug-report.test.js', () => {
         }
 
         // create a schema
-        const mySchema = {
+        const mySchema: RxJsonSchema<any> = {
             version: 0,
             primaryKey: 'passportId',
             type: 'object',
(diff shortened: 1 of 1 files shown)


pubkey:
Hi @adam-lynch 
I am not sure if this test is what you want it to be:
The test breaks because `assert.strictEqual(myDocument.get('tags'), tags);` throws.
But the `tags` object is never `strictEqual` to the thing you get from `get('tags')`, because RxDB will add getters for the population and observable fields.

Checking if the content is the same, works

```ts
 assert.deepStrictEqual(myDocument.get('tags').hello, tags.hello);
assert.deepStrictEqual(myDocument.get('tags').world, tags.world);
```

Also this is why `toJSON().tags` is strict equal, because toJSON() returns the plain data object, without any RxDB specific getters/setters.

adam-lynch:
@pubkey ah yes, I've improved the test now. While what you said is true, it's still failing now; both `.get('tags')` and `.tags` have unexpected extra fields (`additionalProperties` and `type`).

pubkey:
Can you rebase so that the CI can run?

pubkey:
RxDB uses a prebuild constructor (per collection) to create the RxDocument from the documentData that comes out of the storage.
This constructor knows all getters/setters, ORM methods and so on. This makes creating the RxDocument much faster compared to cloning the data and adding these things to each document object.
Therefore it does not add getters for fields that are not known at collection-creation time (=fields that are not in the schema).

The only other options would be the proxy API, but the Proxy API is [so slow](https://thecodebarbarian.com/thoughts-on-es6-proxies-performance) that I do not want to use it for accessing document properties, because the performance is so important at this.

So atm `toJSON()...` is the only way to get your additional properties. Atm I am not sure how this could be fixed.

adam-lynch:
OK I understand; if we want to use `additionalProperties: true`, we must access it via `toJSON()` / `toMutableJSON()`.

pubkey:
Yes, atm you need `toJSON()`.
I see some options on how to fix that in the future, please leave this PR open.

pubkey:
I did some performance tests and refactorings. This will be fixed in the next major version.

Issue #4926, opened 2023-09-07
.get() doesn't work for field defined via definitions


adam-lynch:
I'm not sure this is a bug or not. I'm not sure if what I'm trying to do is supported.

I've defined a field in the schema like this:

```
tags: {
  type: 'object',
  additionalProperties: {
    $ref: '#/definitions/tag',
  },
},
```

And in the schema (on the same level as `properties`), I have:

```
definitions: {
  tag: {
    properties: {
      created: {
        type: 'integer',
      },
      name: {
        type: 'string',
      },
    },
    required: ['created', 'name'],
  },
},
```

If I fetch an `RxDocument` and do `rxDocument.toJSON().tags`, it's as expected. But if I do `rxDocument.get('tags')` or `rxDocument.tags`, I get the schema mixed in the result (e.g. `tags.additionalProperties` exists).

I have created a pull-request to show this behaviour: #4927

pubkey:
Using $ref to reference other schemas is not possible. This is intentional. Parsing the $refs is big burden which would slow down initial page load and database creation time.


Also it is [not clear](https://json-schema.org/understanding-json-schema/structuring.html#base-uri) for a frontend code on how to find the other json schemas that are references. In node.js you would ask the filesystem but in the browser you would have to do more http-requests.

> A schema can reference another schema using the $ref keyword. The value of $ref is a URI-reference that is resolved against the schema’s [Base URI](https://json-schema.org/understanding-json-schema/structuring.html#base-uri).


I recommend to normalize the $ref schemas at build time, not at runtime. There are multiple npm packages that do that for you.

adam-lynch:
The only reason I did this was because our existing data contains an object with unknown keys, each with an object as a value matching:

```
    {
      created: {
        type: 'integer',
      },
      name: {
        type: 'string',
      },
    }
```

Could you please confirm whether it's possible or not to achieve this in RxDB?

According to ChatGPT, I could do this with JSONSchema...

```
{
  "type": "object",
  "additionalProperties": {
    "type": "object",
    "properties": {
      created: {
        type: 'integer',
      },
      name: {
        type: 'string',
      }
    },
  }
}
```

...but I think RxDB doesn't support using `additionalProperties` like this

pubkey:
I do not fully understand the use case.
RxDB only takes the fields from the properties object. `additionalProperties` has boolean values, not an object.
All fields are optional by default, so couldnt you just define the "created" and "name" field like normal and not having it required?

[...]

PR #4951, opened 2023-09-19
Add test to show that you can't use `patternProperties` in schema

adam-lynch:
## This PR contains:
 - IMPROVED TESTS

This is for #4926. It shows that if you use `patternProperties` to define the value of unknown keys, that it'll error due to schema key validation.

### DIFF
--- a/src/plugins/dev-mode/check-schema.ts
+++ b/src/plugins/dev-mode/check-schema.ts
@@ -70,7 +70,8 @@ export function validateFieldsDeep(rxJsonSchema: RxJsonSchema<any>): true {
         if (
             typeof fieldName === 'string' &&
             typeof schemaObj === 'object' &&
-            !Array.isArray(schemaObj)
+            !Array.isArray(schemaObj) &&
+            path.split('.').pop() !== 'patternProperties'
         ) checkFieldNameRegex(fieldName);
 
         // 'item' only allowed it type=='array'
--- a/test/unit/bug-report.test.ts
+++ b/test/unit/bug-report.test.ts
@@ -9,13 +9,13 @@
  * - 'npm run test:browser' so it runs in the browser
  */
 import assert from 'assert';
+import AsyncTestUtil from 'async-test-util';
 import config from './config';
 
 import {
     createRxDatabase,
     randomCouchString
 } from '../../plugins/core';
-import { RxJsonSchema } from '../../src';
 
 describe('bug-report.test.js', () => {
     it('should fail because it reproduces the bug', async () => {
@@ -36,7 +36,7 @@ describe('bug-report.test.js', () => {
         }
 
         // create a schema
-        const mySchema: RxJsonSchema<any> = {
+        const mySchema = {
             version: 0,
             primaryKey: 'passportId',
             type: 'object',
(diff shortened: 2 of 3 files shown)


adam-lynch:
I've just pushed a fix for that but it's still failing (the result includes `patternProperties` and `type`)

adam-lynch:
How can we move this forward?

adam-lynch:
I've synced this with master and improved the test so it's more clear

adam-lynch:
Can we please move this forward?

adam-lynch:
> Can we please move this forward?

To be clear, I'm willing to put in work on this if needed (as I have done), I just need some direction if so.

pubkey:
What is the state of this PR?
I think the proposed solution `path.split('.').pop() !== 'patternProperties'` looks good.
Can you move the test to somehwere else, like the rx-schema tests?

The line that atm fails the CI should be removed. Accessing the properties like that is not possible
`assert.deepStrictEqual(Object.keys(myDocument.get('tags')), Object.keys(tags), 'Object.keys(myDocument.get(\'tags\'))');`

adam-lynch:
This PR's initial aim was to prove that `patternProperties` can't be used. Now it has become a PR to fix that.

I fixed the initial regex issue. But in my opinion it still doesn't work because the `tags` object retrieved has extra keys (`patternProperties` and `type`). It seems you disagree with that;

 > The line that atm fails the CI should be removed. Accessing the properties like that is not possible
`assert.deepStrictEqual(Object.keys(myDocument.get('tags')), Object.keys(tags), 'Object.keys(myDocument.get(\'tags\'))');`

It seems like you're saying that you can't do `Object.keys(...)` of `myDocument.get('tags')`. I don't understand why; it gives the correct data, plus some extra unnecessary stuff from the schema. If it didn't return those extra properties, it would be perfect.

If you're right, then I can remove the lines that break the test and we can merge it, but I wouldn't want to use `.get(...)` in our project. And generally it would mean that there is no 100% reliable way to get individual properties from an `RxDocument`; `.toJSON()[key]` must always be used.

adam-lynch:
@pubkey we're pushing hard to rewrite all of our app to use RxDB and this is a bit of a blocker / an unknown that we'd like to resolve. Could I please get an answer on the above?

pubkey:
Using .get() on unknown properties will work in the next major RxDB version.
Please remove this line from the test so we can check if everything else works.

adam-lynch:
> Using .get() on unknown properties will work in the https://github.com/pubkey/rxdb/pull/4949#issuecomment-1747760592 major RxDB version.

Ok I had no idea that applied to the use of `patternProperties` (as well as `additionalProperties`); I would have guessed `patternProperties` aren't "unknown". The point in using `patternProperties` vs `additionalProperties` was to give them schema.

I'll update this test.

adam-lynch:
Done

pubkey:
Thank you for your work. I merged this into the v15 branch.

pubkey:
The beta will be release this week. Non-beta release will be at least 4 weeks. Maybe longer depending on which problems occur.

adam-lynch:
@pubkey ok thanks. I assume that means there will be matching `rxdb` and `rxdb-premium` versions released.

pubkey:
Yes, premium will also be released in the matching beta version.

PR #5124, opened 2023-10-20
Docs edits

ianengelbrecht:
- IMPROVED DOCS

Before merging please check what I said below the code example for destroy() is correct.

Also the function names destroy() and remove() are confusing. There is also RxCollection.remove(), which does something different (drops a collection entirely, unless I've misunderstood RxDatabase.remove()). For the next major version may I suggest the following:

- instead of destroy() have RxDatabase.quit() to do what destroy currently does (shut down and remove the database instance)
- instead of remove(), RxDatabase.clearCollections(). I'm not 100% sure what remove() is doing, emptying a collection or dropping the collection. If dropping entirely then perhaps removeAll() to be consistent with RxCollection.remove().

I'm not 100% sure I've understood these two functions, and RemoveRxDatabase() completely correctly, and when you would use each one. Destroy is a pretty serious thing though, I would expect nothing left after I destroy something :-P


ianengelbrecht:
With a bit more work I'm finding other issues around RxDatabase that I would like to resolve before making this pull request.

PR #5174, opened 2023-10-27
Fix replication primitive pull checkpoint type

swnf:
## This PR contains:
 - IMPROVED typings

- If there is no checkpoint yet, the pull handler's checkpoint is always `undefined`. According to the type definitions, the checkpoint is always a `CheckpointType`. I've updated the parameter types to be `CheckpointType | undefined`.
- It seems like the pull handler can also return no checkpoint if it did not pull anything and the replication just started. In the replication. In the replication plugin [code](https://github.com/pubkey/rxdb/blob/8b7e6c9197830bb1bcea5e9b2a412e9c5e6db401/src/plugins/replication/index.ts#L201-L204) it seems like the returned checkpoint should be `null` in this case. I've added `null` the the pull handler's return type, but I'm not sure if this is correct.

### DIFF
--- a/src/plugins/replication-firestore/index.ts
+++ b/src/plugins/replication-firestore/index.ts
@@ -123,7 +123,7 @@ export function replicateFirestore<RxDocType>(
     if (options.pull) {
         replicationPrimitivesPull = {
             async handler(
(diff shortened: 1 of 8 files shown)


pubkey:
Thank you, merged.

PR #5180, opened 2023-10-28
Update README.md

vaibhav-009:
## This PR contains:

IMPROVED typings

There are certain documents link available in the readme file, but one section is not linked with its respective one in the image.
screenshots: 
image


pubkey:
The link you have added is pointing to the pouchdb adapters. This page is outdated and no longer valid. I think we should just remove that list item because cordova/phonegap it ancient technologie anyways.

> The PouchDB RxStorage [is deprecated](https://rxdb.info/questions-answers.html#why-is-the-pouchdb-rxstorage-deprecated) and should no longer be used in new projects.

PR #5187, opened 2023-10-30
replaceAll is not a function

jwallet:
## This PR contains:
- a bug fix

issue with string replaceAll on react-native with expo (targets: "ESNext") not supported.
replace with a regex produces the same result.

https://stackoverflow.com/questions/13340131/string-prototype-replaceall-not-working


pubkey:
Thank you, merged.

PR #4950, opened 2023-09-19
Add test to show that you can't get data for object field defined with `additionalProperties: { ... }`

adam-lynch:
## This PR contains:
 - IMPROVED TESTS

## Describe the problem you have without this PR
This is for #4926. It shows that if you use `additionalProperties` with an object value to define the value of unknown keys, that you can't then pull out the data.

PR #5197, opened 2023-11-02
Add test to show that you can't get data for object field defined with `additionalProperties: { ... }`

adam-lynch:
This replaces #4950. While I was off work, someone accidentally deleted our fork which closed the old PR.

### DIFF
--- a/test/unit/bug-report.test.ts
+++ b/test/unit/bug-report.test.ts
@@ -125,9 +125,9 @@ describe('bug-report.test.js', () => {
          * assert things,
          * here your tests should fail to show that there is a bug
          */
-        assert.strictEqual(myDocument.toJSON().tags, tags);
-        assert.strictEqual(myDocument.get('tags'), tags);
-        assert.strictEqual(myDocument.tags, tags);
+        assert.deepStrictEqual(myDocument.toJSON().tags, tags);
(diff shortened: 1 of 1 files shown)


test/unit/bug-report.test.ts
```diff
@@ -83,54 +97,39 @@ describe('bug-report.test.js', () => {
...
          */
-        assert.strictEqual(myDocument.age, 56);
-
-        // you can also wait for events
-        const emitted = [];
-        const sub = collectionInOtherTab.mycollection
-            .findOne().$
-            .subscribe(doc => emitted.push(doc));
-        await AsyncTestUtil.waitUntil(() => emitted.length === 1);
+        assert.strictEqual(myDocument.toJSON().tags, tags);
```
pubkey, line 128: shouldn't this be assert.deepStrictEqual()?

adam-lynch: Updated

pubkey:
It is really strange that this only fails on bun but works on Node and Deno

pubkey:
Added the test to the test suite. https://github.com/pubkey/rxdb/pull/5244
I had to use `deepEqual()`, somehow the assert module of bun is broken.

PR #5245, opened 2023-11-12
Feature/angular 17

pubkey:


pubkey:
Broken with `Module '"/workspace/rxdb/examples/angular/src/main.server"' has no default export. Did you mean to use 'import { bootstrap } from "/workspace/rxdb/examples/angular/src/main.server"' instead?ts(2613)`

Issue #5256, opened 2023-11-14
PushModifier applied to pre-change legacy document, resulting in old document sent to endpoint


alex-hladun:
In our project up to Rxdb 13, we have used the pushModifier as a means to control whether a document is sent to the server or not, as [returning null is supposed to skip the document.](https://rxdb.info/replication-graphql.html). However, as of rxdb 13, the pushModifier is run twice, once on the new document, as well as the original document.

Example commit which highlights this issue [is here](https://github.com/pubkey/rxdb/commit/5db08ed990095bef9920adf5d04b3ddf13b3020d). After adding a hero and marking the 'doNotSync' field as true, the [pushModifier is applied to the new document, and then the original document](https://github.com/pubkey/rxdb/blob/316f06508e1998cf9cef9396232608ea9fb1c639/src/plugins/replication/index.ts#L258). 

This results in un-wanted behaviour, as the original document is sent to the server, despite doNotSync being marked as true. 

I see this as a bug, as we have no way to make changes to a local document and **prevent it from being sent to the server.** Any possible solutions would be welcome, as the only fix I can think of right now is to copy the replication plugin and modify that line of code.

pubkey:
Yes I think this a bug, let me fix it.

alex-hladun:
Ok, i think the bug I am refrerring to is still present. Working on a test to reproduce!

alex-hladun:
Fixed in https://github.com/pubkey/rxdb/issues/5256

PR #5259, opened 2023-11-15
FIX PushModifier applied to pre-change legacy document, resulting in old document sent to endpoint

pubkey:
https://github.com/pubkey/rxdb/issues/5256

### DIFF
--- a/src/plugins/replication/index.ts
+++ b/src/plugins/replication/index.ts
@@ -32,6 +32,7 @@ import type {
 } from '../../types/index.d.ts';
 import { RxDBLeaderElectionPlugin } from '../leader-election/index.ts';
 import {
+    arrayFilterNotEmpty,
     ensureNotFalsy,
     errorToPlainJson,
     flatClone,
@@ -273,7 +274,7 @@ export class RxReplicationState<RxDocType, CheckpointType> {
                             return row;
                         })
                     );
-                    const useRows: RxReplicationWriteToMasterRow<RxDocType>[] = useRowsOrNull.filter(row => !!row) as any;
+                    const useRows = useRowsOrNull.filter(arrayFilterNotEmpty);
 
                     let result: WithDeleted<RxDocType>[] = null as any;
 
--- a/src/plugins/replication/index.ts
+++ b/src/plugins/replication/index.ts
@@ -255,9 +255,12 @@ export class RxReplicationState<RxDocType, CheckpointType> {
                         collection: this.collection
                     });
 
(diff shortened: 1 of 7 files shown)


src/plugins/replication/index.ts
```diff
@@ -255,9 +256,12 @@ export class RxReplicationState<RxDocType, CheckpointType> {
...
                     });
 
-                    const useRows = await Promise.all(
+                    const useRowsOrNull = await Promise.all(
                         rows.map(async (row) => {
                             row.newDocumentState = await pushModifier(row.newDocumentState);
+                            if (row.newDocumentState === null) {
+                                return null;
+                            }
                             if (row.assumedMasterState) {
```
alex-hladun, line 265: Is this row required? It was added in 13. If i'm not mistaken, this will always result in the old document state being sent to the server prior to any document modifications (update, incrementals, even remove)

pubkey: Yes I think this is correct.
It prevents the pushModifier from being called on the assumedMasterstate, in case it returned null on the newDocumentState.
Otherwise creating a correct pushModifier would be way more complex because you would have to decide which of the calls it is and if you should return null or not.

alex-hladun: I reproduced the issue we are seeing [in this branch, within the examples/graphql demo](https://github.com/alex-hladun/rxdb/tree/pushReplicationExample-Issue). I wasn't able to write a test, but it can be reproduced by adding a hero, and then clicking 'Do Not Sync' on the hero in the list. The doc is sent to the server initially as expected. 
But after toggling 'doNotSync' to true: 
1. the document goes through the push modifier, null is returned, and the doc is skipped, as expected.
2. The **original** document also goes through the pushModifer, and is sent to the server. This is un-expected behaviour. 

image

alex-hladun: Update: See now that we're filtering out rows with null, I didn't have the latest version built properly. This looks great, thanks for the prompt fix!

alex-hladun (approved):

src/plugins/replication/index.ts
pubkey: I added another test but cannot reproduce your problem: https://github.com/pubkey/rxdb/pull/5259/commits/b1fe251ae10a9808047afaa7798c03cf37da559e

pubkey: @alex-hladun haha this was bad timing :) I will merge this PR and release the new beta.

PR #5272, opened 2023-11-17
Bug: compound index strings calculate wrong bounds

eliias:
## This PR contains:

A test demoing a bug in at least the Memory and IndexedDB storage (most likely applies to all storage plugins that have a similar implementation, especially determining bounds).

There seems to be a bug in some RxDB storages (IndexedDB, Memory) when a compound index covers more than all fields of a selector (one or more additional fields are present in comparison to the fields used in the query selector). This can lead to situations where the query planner sets the `queryPlan.selectorSatisfiedByIndex` to true, but the bounds (upper/lower) might be wrong (in our case a boolean integer was set to a range between 0-9). When the additional index field is more significant than the actual fields used in the selector, the upper bound is too high, and the number of documents returned is greater than the expected number.

The storage implementations assume that `queryPlan.selectorSatisfiedByIndex=true` implies that the range defined by the bounds includes all valid documents, and skips the `queryMatcher` logic.

I debugged by logging the `endKeys` provided to the `getStartIndexStringFromUpperBound` function, and by checking the resulting string. For the example in this PR, the `endKeys` might look like this:

```
_deleted  |  exists  |  hasHighlights  |  lastOpenedAt
0                 1              0                         1600000000000
0                 1              9                         1650000000000
```

As you can see, the specificity of the `hasHighlights` key is overruling the otherwise correct upper bound of the timestamp. Therefore, the query returns 2 docs after execution instead of one. The actual problem is that `hasHighlights` is not provided to the selector, therefore the key is set to `min/max`. IMO there is no way this index is going to work without the `hasHighlights` value being provided to the query (selector), and therefore it shouldn't be determined as the best index.

I don't have a fix yet, the most simple one would be to just run the `queryMatcher` every time, but that obviously has some severe cost. I think the right solution would be to not use the index in such cases. That would require a change in the query-planner itself.

## Todos
- [x] Tests
- [ ] Implement fix

### DIFF
--- a/test/unit/bug-report.test.ts
+++ b/test/unit/bug-report.test.ts
@@ -9,7 +9,6 @@
  * - 'npm run test:browser' so it runs in the browser
  */
 import assert from 'assert';
-import AsyncTestUtil from 'async-test-util';
 import config from './config.ts';
 
 import {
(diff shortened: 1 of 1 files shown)


pubkey:
Fixed [see here](https://github.com/pubkey/rxdb/issues/5273)

PR #5305, opened 2023-11-22
Feature/replication p2pcf

pubkey:
Add [p2pcf](https://github.com/gfodor/p2pcf) to the WebRTC replication plugin.
This will be the easiest way to provide a WebRTC singaling servers for all RxDB users so they do not have to setup their own one for testing and development (only for production they should use their own server).


pubkey:
Depends on https://github.com/gfodor/p2pcf/pull/24

pubkey:
Closing this. Instead the WebRTC plugin has been improved and we have a default public signaling server.

PR #5322, opened 2023-11-27
FIX observe query returns outdated results

pubkey:


pubkey:
This did not work. JavaScript Date() precission is not good and we should not rely on timestamps to filter events in any case.

Issue #5358, opened 2023-12-05
Migrate update operation to mingo library


kofrasa:
The [mingo](https://www.npmjs.com/package/mingo) library now has full support for update operators and can replace [modifyjs](https://www.npmjs.com/package/modifyjs) which is lagging behind.

Relevant files.

https://github.com/pubkey/rxdb/blob/master/src/plugins/update/index.ts
https://github.com/pubkey/rxdb/blob/master/src/plugins/crdt/index.ts
https://github.com/pubkey/rxdb/blob/master/package.json

pubkey:
Hi @kofrasa 
Yes I have seen that already. Great feature :+1: 

I will work on that.

pubkey:
@kofrasa I have some questions:

- Shouldn't `updateObject()` return the new, mutated object? How can I get the updated document after calling the method?
- Does `updateObject` mutate the input document? I have seen there is a cloneMode but from the code it looks like that only clones the operations, not the document.

kofrasa:
I don't think `$pushAll` exists. If you meant `$pullAll` that is implemented. To your earlier question, the update function updates the object in-place which was carefully considered as chosen as the preferred API. So the input object gets modified. As correctly observed, the clone mode is for how to treat the arguments. Cloning the entire input object all the time would be inefficient and is left up to the user of the library.

kofrasa:
What is not implemented is `$setOnInsert` since it doesn't make sense when working on a single object.

kofrasa:
With regards to the typings, the update is expression is typed [here](https://github.com/kofrasa/mingo/blob/main/src/updater.ts#L15) to only allows a single operator as specified in MongoDB.

pubkey:
Ah, pushAll is deprecated
> Deprecated since version 2.4: Use the [$push](https://www.mongodb.com/docs/v2.6/reference/operator/update/push/#up._S_push) operator with [$each](https://www.mongodb.com/docs/v2.6/reference/operator/update/each/#up._S_each) instead.

https://www.mongodb.com/docs/v2.6/reference/operator/update/pushAll/

pubkey:
Closing this, changes applied here https://github.com/pubkey/rxdb/pull/5360
Thank you again @kofrasa for your wonderful library and the hint about the new update feature.

PR #5360, opened 2023-12-06
rEFACTOR use mingo for updates

pubkey:
https://github.com/pubkey/rxdb/issues/5358


kofrasa:
+1 for the quick update.

Issue #5382, opened 2023-12-14
`hasOwnProperty is not a function` in plugins/utils-object.ts


emil14:
This is how object in a collection was changed:

```js
{
    type: 'doc',
    content: [
        {
            type: 'text',
            content: '#'
        }
    ]
}

->

{
    type: 'doc',
    content: [
        {
            type: 'heading',
            attrs: {
                level: 1
            },
            content: []
        }
    ]
}
```

For doing so I did

```ts
await existingInDB.incrementalModify((doc: RxDocument<Document>) => ({
	...doc,
	data: newContent,
}))
```

I also tried `modify` and `update`, nothing helps.

Despite this error everything looks working but when I then delete this document like this

```ts
await collection.docs.findOne(id).remove()
```

I got conflict

```
rx-error.ts:112 Uncaught (in promise) RxError (CONFLICT): RxError (CONFLICT):
Document update conflict. When changing a document you must work on the previous revision
Given parameters: {
collection:"docs"
id:"e2ba1ff0-79f6-48ea-9818-f4ee196e70ba"
writeError:{
  "isError": true,
  "status": 409,
  "documentId": "e2ba1ff0-79f6-48ea-9818-f4ee196e70ba",
  "writeRow": {
    "document": {
      "id": "e2ba1ff0-79f6-48ea-9818-f4ee196e70ba",
      "data": {
        "type": "doc",
        "content": [
          {
            "type": "paragraph",
            "content": [
              {
                "type": "text",
                "text": "#"
              }
            ]
          }
        ]
      },
      "_meta": {
        "lwt": 1702569841972.01
      },
      "_deleted": true,
      "_attachments": {},
      "_rev": "13-wzlhzxuhoy"
    },
    "previous": {
      "id": "e2ba1ff0-79f6-48ea-9818-f4ee196e70ba",
      "data": {
        "type": "doc",
        "content": [
          {
            "type": "paragraph",
            "content": [
              {
                "type": "text",
                "text": "#"
              }
            ]
          }
        ]
      },
      "_meta": {
        "lwt": 1702569260031.01
      },
      "_deleted": false,
      "_attachments": {},
      "_rev": "12-wzlhzxuhoy"
    }
  },
  "documentInDb": {
    "id": "e2ba1ff0-79f6-48ea-9818-f4ee196e70ba",
    "data": {
      "type": "doc",
      "content": [
        {
          "type": "heading",
          "attrs": {
            "level": 1
          },
          "content": [
            {
[...]

PR #5391, opened 2023-12-15
fix(src:plugins:utils): properly call hasOwnProperty in deepFreeze util

emil14:
## This PR contains:

 - A BUGFIX

#5382

### DIFF
--- a/src/plugins/utils/utils-object.ts
+++ b/src/plugins/utils/utils-object.ts
@@ -6,7 +6,7 @@ export function deepFreeze<T>(o: T): T {
     Object.freeze(o);
     Object.getOwnPropertyNames(o).forEach(function (prop) {
         if (
-            (o as any).hasOwnProperty(prop)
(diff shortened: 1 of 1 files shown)


pubkey:
Hi @emil14  The fix looks good to me. Do you think it is possible to reproduce the problem in a unit test? Then we could be sure that this does not break again in the future.

emil14:
@pubkey hey there

Unfortunately I don't I have time for that now, this will require diving into the structure of the repository.

I suggest you enable https://eslint.org/docs/latest/rules/no-prototype-builtins this linter, this problem could arrive not only in that piece of code but literally everywhere you use methods defined on `Object` because of how JavaScript works. One unit test won't protect anyway

pubkey:
@emil14  thanks for the suggestion. I am working on enabling the linting rule here https://github.com/pubkey/rxdb/pull/5398

pubkey:
Closed in favor of https://github.com/pubkey/rxdb/pull/5398

Issue #5370, opened 2023-12-13
Electron : Object has been destroyed


pwi-informatique:
Hi,

I'm facing a small problem when using RxDB with Electron.
For ease of use, I've recreated a small project that can be accessed [here](https://github.com/pwi-informatique/electron-rxdb-test), this project is based on [RxDB Electron example](https://github.com/pubkey/rxdb/tree/ea2632dd69950747c9db089eef66022ee0dfc101/examples/electron).

#### Problem description

After launching the application, if you close one of the two windows and try to add a hero, the following error appears in the log / in an error popup :

```
(node:35380) UnhandledPromiseRejectionWarning: TypeError: Object has been destroyed
    at WebContents.send (node:electron/js2c/browser_init:2:79151)
    at XXXXXXXXXXXXXXXXXXXX\electron-rxdb-test\node_modules\rxdb\dist\lib\plugins\electron\rx-storage-ipc-main.js:32:14
    at Set.forEach (<anonymous>)
    at Object.send (XXXXXXXXXXXXXXXXXXXX\electron-rxdb-test\node_modules\rxdb\dist\lib\plugins\electron\rx-storage-ipc-main.js:31:19)
    at Object.next (XXXXXXXXXXXXXXXXXXXX\electron-rxdb-test\node_modules\rxdb\dist\lib\plugins\storage-remote\remote.js:185:18)
(Use `electron --trace-warnings ...` to show where the warning was created)
(node:35380) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
```

#### Workaround

I've slightly modified the code of the `exposeIpcMainRxStorage` function to remove windows that have been destroyed, from the `openRenderers` list.
The modified Typescript code of the `exposeIpcMainRxStorage` function can be consulted [here](https://github.com/pwi-informatique/electron-rxdb-test/blob/7a9d26a0805e051565cd6555339f26036c92d19d/rx-storage-ipc-main.custom.ts).

You can follow the steps listed in the README file to reproduce the bug and test the workaround.

pubkey:
Hi @pwi-informatique 
thank you for investigating. The proposed change look good to me, can you do a PR?
[...]

PR #5395, opened 2023-12-16
fix(src:plugins:electron): Object has been destroyed

pwi-informatique:
## This PR contains:
A BUGFIX

[#5370](https://github.com/pubkey/rxdb/issues/5370)


pubkey:
Thank you, merged.

PR #5418, opened 2023-12-21
Operation name graphql

jwallet:
## This PR contains:
 - A NEW FEATURE
- ~breaking the helper function~

GraphQL `operationName` is required with some tools to valid the gQL request

We used `patch-package` to change this in `rxdb` but we will like to be able to upgrade without it.


src/plugins/replication-graphql/query-builder-from-rx-schema.ts
```diff
@@ -96,15 +101,15 @@ export function pushQueryBuilderFromRxSchema(
...
 
     const variableName = collectionName + prefixes.pushRow;
 
-
     const returnFields: string[] = Object.keys(input.schema.properties);
 
     const builder: RxGraphQLReplicationPushQueryBuilder = (pushRows) => {
         const query = '' +
-            'mutation ' + prefixes.push + ucCollectionName + '($' + variableName + ': [' + ucCollectionName + 'Input' + prefixes.pushRow + '!]) {\n' +
+            'mutation ' + operationName + '($' + variableName + ': [' + ucCollectionName + 'Input' + prefixes.pushRow + '!]) {\n' +
```
jwallet, line 112: I don't know if this a internal helper, but it never worked for us, the operationName does not match the queryName and the input are forced by some lib like HotChocolat to end with `Input` on the backend side so here it will be "CollectionInputPushInput"

pubkey:
Hi @jwallet 
Do you think this is a breaking change?

jwallet:
I can remove the changes made to the helper functions to make it non
breaking, but it looks like these functions does not follow the required
graphql recommandations found in apollo or hotchocolate documentation

It can stay anoptional param

jwallet:
the change the helper functions queryBuilderFromRxSchema to prevent a breaking change with the pullStream.

jwallet:
Also, if someone use a query like this one below he can't just ignore the "operationName" param, but the operationName must match the name after query/mutation

```graphql
query($checkpoint: CheckpointInput, $limit: Int!)
                    {
                        collectionFeedForRxDBReplication(checkpoint: $checkpoint, limit: $limit) {
                            collection {
                                documents {
                                    id
                                    name
                                    age
                                    updatedAt
                                    deleted
                                }
                                checkpoint {
                                    id
                                    updatedAt
                                }
                            }
                        }
                    }
```

pubkey:
Thank you, merged.

PR #5421, opened 2023-12-22
Custom fetch to support axios for the graphQL replication

jwallet:
## This PR contains:
- A non-breaking new feature to the graphql-replication plugin to make it possible to support a custom fetch that fits the `typeof fetch`
- it always default to `window.fetch` otherwise it will be the responsibility of the lib user to make sure its custom fetch fits the fetch API even for the `then()` and the `catch()`

We imported the graphql-replication plugin and customized it to make the helper `graphQLRequest` support `axios` which is a must-to-have lib for our project, since we use it everywhere to auto manages the bearer refresh token (renewal) and makes sure all requests get the updated access token before making a new request with the `axios.interceptors.request.use` and we added custom logic to the `axios.interceptors.response.use` to handle graphql [`extensions.code`](https://www.apollographql.com/docs/apollo-server/data/errors/#built-in-error-codes)  that rxdb doesn't handle.

### example with `axios` as a custom fetch
```ts
// options...
fetch: async (request: Request) => axios({
    method: request.method || 'POST',
    url: request.url,
        data: await request.json(),
        // axios.interceptors.request.use can reinject all headers as well as the Authorization Bearer token
        headers: request.headers,
    })
    .then(res => async () => Promise.resolve(res.data))
    .catch(err => {
        throw err.response.data;
    })
// other options...
```

### DIFF
--- a/src/plugins/replication-graphql/helper.ts
+++ b/src/plugins/replication-graphql/helper.ts
@@ -16,6 +16,7 @@ export type GraphQLErrors = Array<GraphQLError>;
 
 
 export function graphQLRequest(
+    fetchRequest: WindowOrWorkerGlobalScope['fetch'],
     httpUrl: string,
     clientState: RxGraphQLReplicationClientState,
     queryParams: RxGraphQLReplicationQueryBuilderResponseObject
@@ -33,7 +34,8 @@ export function graphQLRequest(
             credentials: clientState.credentials,
         }
     );
-    return fetch(req)
+    
+    return fetchRequest(req)
         .then((res) => res.json())
         .then((body) => {
             return body;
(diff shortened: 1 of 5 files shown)


src/plugins/replication-graphql/helper.ts
```diff
@@ -16,6 +16,7 @@ export type GraphQLErrors = Array<GraphQLError>;
 
 
 export function graphQLRequest(
+    fetchRequest: WindowOrWorkerGlobalScope['fetch'],
```
jwallet, line 19: maybe you will prefer to use `typeof fetch`

jwallet:
Note: another MR will follow to be able to remove and purge a specific document without waiting the replication sync. The backend may refuse the newDocumentState (no assumedMasterState) if the `extensions.code` is one of the case that will always reject the payload, so the replication will never finish, therefore even if we `doc.remove()` the `cleanup()` will never run, except if we change the 'awaitReplicationsInSync' to 'false' for all documents, but that will be a bigger issue.

I will try to add it to the collection: `collection.purgeOne({selector}|string)`

src/plugins/replication-graphql/helper.ts
pubkey: Yes, please use `typeof fetch`

PR #5422, opened 2023-12-22
PurgeDocumentsById on internalStorage

jwallet:
## This PR contains:
 - A new feature for the internalStorage and the document

I added a way to purge a document from the storage, to be able to resolve the replication when one document is bloqued by the server.

- I added the function to RxDocument, but I won't mind if you remove it from there, to prevent users from using it by default. By default it will only purge _deleted=true documents, unless the user forcePurge it.
- I added the function on many storage, but it misses the premium storages
- I did not add it to the cleanup plugin and did not attach it to the collection (...yet?)

Like said in another PR, sometimes not everything is perfect, the server might return a GraphQL Error that will prevent the replication to finish and will always return the same error for the same `newDocumentState` in error. Since the cleanup policy or manual cleanup() is configured to `awaitReplicationsInSync` for all others documents, it will never be able to get rid of this "corrupted" document. Unless we have a function to purge it by id (soft deleted or not).

If the replication only allows update and no insert, we can then `doc.remove()` and then `doc.purge()` or calling directly `collection.internalStorage.purgeDocumentsById([id], true)` if `RxDocument` does not have the function

This is still in draft and has no tests because I was unsure if I was on the right path and that's something you will be interested to merge.

## Todos
- [ ] Tests
- [ ] Missing premiums storage
- [ ] Documentation
- [ ] Changelog

### DIFF
--- a/src/rx-document.ts
+++ b/src/rx-document.ts
@@ -408,7 +408,7 @@ export const basePrototype = {
      * Purge the removed document from the database.
      *
      * This will not wait for the document to be synced with the server,
-     * use remove() if you need to be sure it is deleted and cleanup() to clean up afterwards
+     * use the cleanup plugin to purge all removed documents properly
     */
     purge(this: RxDocument): Promise<void> {
         if (!this.deleted ) {
--- a/src/rx-document.ts
+++ b/src/rx-document.ts
@@ -410,14 +410,14 @@ export const basePrototype = {
      * This will not wait for the document to be synced with the server,
      * use remove() if you need to be sure it is deleted and cleanup() to clean up afterwards
     */
-    purge(this: RxDocument, forcePurge: boolean = false): Promise<void> {
-        if (!forcePurge && !this.deleted ) {
+    purge(this: RxDocument): Promise<void> {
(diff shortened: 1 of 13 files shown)


src/rx-document.ts
```diff
@@ -403,6 +404,21 @@ export const basePrototype = {
                 return this.collection._docCache.getCachedRxDocument(removedDocData);
             });
     },
+    /**
+     * Purge the removed document from the database.
+     *
+     * This will not wait for the document to be synced with the server,
+     * use remove() if you need to be sure it is deleted and cleanup() to clean up afterwards
```
jwallet, line 411:
```suggestion
     * use the cleanup plugin to purge all removed documents properly
```

pubkey:
I do not want to merge this. I spend so much time on making the RxStorage layer less complex and this adds more complexity for the rare edge use case where fixing the backend is not possible.
If your backend server has wrong behavior you should fix that instead. Or fork the GraphQL plugin and patch the behavior in the push handler.

jwallet:
no worries, closing.

Issue #5414, opened 2023-12-21
[!important]: v15 official realease undefined is now treated as null, this broken whole app used rxdb previous version.


1yasa:
`storage-dexie`.

In RxDB v15 official release, undefined is now treated as null, which is inconsistent with the behavior of previous versions. Is there a solution to ensure compatibility with the previous data processing?

1yasa:
`remind_time?:number`

### prev v15.0.0-beta42
set remind_time: undefined => schema check pass

### v15.0.0
set remind_time: undefined => schema check failed

### Temporary solution

define `remind_time?:number|string`, filter update value(undefined => ''), but this may lead to other side effects.

pubkey:
Hi @1yasa 
Can you make a PR with a test case?

1yasa:
> Hi @1yasa This should work exactly like it did in v14. Can you make a PR with a test case?

<img width="988" alt="截屏2023-12-22 16 56 41" src="https://github.com/pubkey/rxdb/assets/127166080/95b738a1-acbe-4ec3-8649-ffb91a4d6e9a">

I'm trying to figure out how to run the test cases for RxDB, but it seems that there are some issues preventing progress, possibly due to something in the `postinstall` process.

<img width="934" alt="截屏2023-12-22 17 00 49" src="https://github.com/pubkey/rxdb/assets/127166080/2aed9dda-f98a-42c6-afcf-1f65a985bdb0">

1yasa:
Here is the test case:

```ts
export type Item = {
        /** @maxLength 30 */
        module: string
        type: 'dir' | 'file'
        /** @maxLength 30 */
        id: string
        name: string
        create_at: number
        pid?: string
        prev_id?: string
        next_id?: string
        icon?: string
        icon_hue?: number
        update_at?: number
        backup_at?: number
}
```
```json
{
          "type": "object",
          "properties": {
                "module": {
                      "type": "string",
                      "maxLength": 30
                },
                "type": {
                      "type": "string",
                      "enum": [
                            "dir",
                            "file"
                      ]
                },
                "id": {
                      "type": "string",
                      "maxLength": 30
                },
                "name": {
                      "type": "string"
                },
                "create_at": {
                      "type": "number"
                },
                "pid": {
                      "type": "string"
                },
                "prev_id": {
                      "type": "string"
                },
                "next_id": {
                      "type": "string"
                },
                "icon": {
                      "type": "string"
                },
                "icon_hue": {
[...]

PR #5423, opened 2023-12-24
Undefined is now treated as null https://github.com/pubkey/rxdb/issue…

1yasa:
## This PR contains:

A unit test of `rxdb@v15.0.0` for inserting a undefined value(this value is optional), rxdb will throw error cause by rxdb tranformed value to null.

https://github.com/pubkey/rxdb/issues/5414


pubkey:
Your test throws a conflict error which is correct because at the second insert the document already exists.

1yasa:
> Your test throws a conflict error which is correct because at the second insert the document already exists.

I tested non-CRDT insertion, insertCRDT in CRDT mode, and regular insert in CRDT mode. The test results indicate that the first two methods meet expectations, but the result of insert in CRDT mode is incorrect.

截屏2023-12-26 15 07 23

pubkey:
Since RxDB v15 we use the [mingo library](https://github.com/kofrasa/mingo) for the nosql-updates. It seems that mingo strips away fields with undefined or replaces the value with null.
So this was causing the dev-mode check to throw an error because the outcoming documents where not equal.

1yasa:
> Since RxDB v15 we use the [mingo library](https://github.com/kofrasa/mingo) for the nosql-updates. It seems that mingo strips away fields with undefined or replaces the value with null. So this was causing the dev-mode check to throw an error because the outcoming documents where not equal.
> 
> I added a fix for the check here: #5445 However you might want to do a PR to the mingo lib to ensure correct merging of undefined fields.

🙏

PR #5429, opened 2023-12-27
Local search - first attempt

klauss194:
## This PR contains:
 - A NEW FEATURE - LocalSearch

Couple of points to take into consideration:
1. Project requires BUILD -> run: npm run build in the root -> THEN -> cd docs-src/ -> npm install , npm run build -> npm run serve 

1. I took lunr-local-search plugin and it won't index files properly -> it would only index /blog files -> So i had to take the plugin code and just add it to the project creating a reference in the documentation. 
2. Modified jsx & css minimally to hide searchbar from homepage
3. Need to further check a bit linking after a deployment -> it works on local
4. I could further clean the plugin folder ( removing unnecessary files )


klauss194:
Just for reference no dist / generated files have been commited unless I missed some - I will double check

PR #5451, opened 2024-01-03
 Test: located the specific reason: The key-compression plugin caused the insertion to incorrectly insert the query index.

1yasa:
## This PR contains:

Unit test, using a optional value in crdt mode, throw `dexieIDBError` error when call collection.count(...).

In CRDT mode, there is an error when using `count`, while in non-CRDT mode, there is no error.


pubkey:
The CI fails atm because the schema is not valid, can you check?

> Fields of type string that are used in an index, must have set the maxLength attribute in the schema

1yasa:
> The CI fails atm because the schema is not valid, can you check?
> 
> > Fields of type string that are used in an index, must have set the maxLength attribute in the schema

Located the specific reason: The key-compression plugin caused the insertion to incorrectly insert the query index. Without key-compression, .count() will be ok.

pubkey:
Is the CI failing at the intended test?

PR #5469, opened 2024-01-05
Bump dexie dep for bugfix, improve supabase example

### LINKED ISSUE #5466: Planning to tackle the "Fix the flaky tests for the "example-supabase" CI" task
Per premium task instructions - creating an issue to indicate I'm going to take a look. Honestly, difficult to say how long it will take, but I spent the last hour doing a quick debug and I'm pretty sure I know what's going on so I'm hoping not more than a day. Will let ya know!

marbemac:
close #5466 

## This PR contains:

- A BUGFIX: upgrade dexie to resolve flakey `Internal error opening backing store for indexedDB.open` error in tests
- SOMETHING ELSE: upgrade the supabase example + hopefully reduce time it takes for the related CI job to complete

**Open questions:**

- The bump from alpha dexie to beta.2. Hopefully it is safe but honestly I did not go through their entire changelog and I don't know enough about rxjs internals to have a gut feel re what changes might be risky.
- Do we want to bump to the latest while we're at it (beta.6) rather than the specific version that includes the fix (beta.2)?


package.json
```diff
@@ -442,7 +442,7 @@
     "broadcast-channel": "7.0.0",
     "crypto-js": "4.2.0",
     "custom-idle-queue": "3.0.1",
-    "dexie": "4.0.0-alpha.4",
+    "dexie": "4.0.1-beta.2",
```
marbemac, line 445: This is the crux of the issue in the flakey test - see the thread in https://github.com/dexie/Dexie.js/issues/543#issuecomment-1831922386 for more information.

Side note - does the premium indexdb plugin also handle this chromium bug? If not, might want to check out dexie's solution :).

Before, during the supabase e2e test (only happens sometimes, hence the flake):

Screenshot 2024-01-04 at 2 17 36 PM

After, during the supabase e2e test (still only happens sometimes, but now handled gracefully by dexie):

Screenshot 2024-01-04 at 3 52 37 PM

examples/supabase/src/index.ts
```diff
@@ -1,28 +1,43 @@
...
+    console.error(
+        `got unhandled error`,
+        e.reason.message,
+        JSON.stringify({
+            // @ts-expect-error ignore
+            frame: e.currentTarget?.ifrm?.id,
+        }),
+        e
+    );
+});
```
marbemac, line 17: It was difficult to get to the bottom of the issue, because the page was throwing unhandled exceptions and test cafe only captures logs, not exceptions. So the test was effectively silently failing, and test cafe was just going into a loop trying to get it to pass until the test finally timed out.

Now if there are any unhandled exceptions on the page, it's logged as an error and the test cafe tests will immediately stop and print out the relevant info - should make future problems much easier to debug.

marbemac (commented):
In summary:

- update dexie to gracefully handle a rare chromium indexdb bug that the supabase e2e test was occasionally triggering (and thus failing)
- while i was at it I updated supabase and modernized the approach to starting it etc. fingers crossed that this has the side effect of reducing CI times

package.json
marbemac: Side note - does the premium indexdb plugin also handle this chromium bug? If not, might want to check out dexie's solution :).

examples/supabase/import-dump.sh
```diff
@@ -1,17 +0,0 @@
...
-# postgres might not be up already
-# so we have to retry
-n=0
-until [ "$n" -ge 20 ]
-do
-   pg_restore -h localhost -d postgres -U postgres dump.sql && break
-   n=$((n+1)) 
-   echo "failed to import, will try again in 5 seconds"
-   sleep 5
-done
```
marbemac, line 15: Ah almost forgot - this loop was always running the max 20 count - it's why the import db step in CI takes ~2m every time.

pubkey:
@marbemac 
Wow, thank you for the investigation and fixes.

Atm the CI fails with `npm ERR! Missing script: "start:supabase"` please check.

> Do we want to bump to the latest while we're at it (beta.6) rather than the specific version that includes the fix (beta.2)?

Yes I think we should directly use beta.6.

> The bump from alpha dexie to beta.2. Hopefully it is safe but honestly I did not go through their entire changelog and I don't know enough about rxjs internals to have a gut feel re what changes might be risky.

I will check the basic behavior to ensure current alpha state is compatible with beta.6 state.

> does the premium indexdb plugin also handle this chromium bug?

Unlikely, the indexeddb premium has a different way of opening storages compared to dexie so it might not be relevant, but I will check.

marbemac:
> Atm the CI fails with npm ERR! Missing script: "start:supabase" please check.
> 
> Yes I think we should directly use beta.6.

Hopefully all set!

PR #5492, opened 2024-01-11
test: .count() with key-compression plugin

1yasa:
## This PR contains:
Unit test case for doc.count() with key-compression plugin.

Throw store error about indexes when use .count() with key-compression plugin.


pubkey:
Thank you, fixed in https://github.com/pubkey/rxdb/pull/5500

PR #5554, opened 2024-01-27
fix(Angular): add process.nextTick() polyfill

DanielKucal:
Without it when using `replicateWebRTC` we'll see console error when peer becomes visible:
 
> ERROR TypeError: process.nextTick is not a function
    at resume (_stream_readable.js:832:13)
    at Readable.resume (_stream_readable.js:824:5)
    at Readable.on (_stream_readable.js:753:39)
    at createPeerConnection (connection-handler-simple-peer.js:101:31)
    at connection-handler-simple-peer.js:140:19
    at Array.forEach (<anonymous>)
    at ensureNotFalsy.onmessage [as __zone_symbol__ON_PROPERTYmessage] (connection-handler-simple-peer.js:136:32)
    at WebSocket.wrapFn (zone.js:755:39)
    at _ZoneDelegate.invokeTask (zone.js:402:31)
    at core.mjs:14369:55

image

### DIFF
--- a/docs-src/docs/install.md
+++ b/docs-src/docs/install.md
@@ -40,6 +40,7 @@ You have to add them by your own, like we do [here](https://github.com/pubkey/rx
 (window as any).global = window;
 (window as any).process = {
     env: { DEBUG: undefined },
+    nextTick: () => {},
 };
 ```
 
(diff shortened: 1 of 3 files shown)


pubkey:
Shouldnt the nextTick method call a callback after awaiting a new promise resolve?

pubkey:
I added a check with a more helpful error message.

DanielKucal:
@pubkey, after investigating, I think the proper polyfill for process.nextTick should be:
```js
process.nextTick = function (fn, ...args) {
    setTimeout(function() {
        fn.apply(null, args);
    }, 0);
};
 ```

pubkey:
Please make a PR

PR #5672, opened 2024-02-23
Fix some incorrect OPFS statements

tomayac:
## This PR contains:

 - IMPROVED DOCS

Someone is wrong on the Internet `;-)`.

### DIFF
--- a/docs-src/docs/rx-storage-opfs.md
+++ b/docs-src/docs/rx-storage-opfs.md
@@ -6,7 +6,7 @@ description: Origin Private File System (OPFS) Database with the RxDB OPFS-RxSto
 
 # Origin Private File System (OPFS) Database with the RxDB OPFS-RxStorage
 
-With the [RxDB](https://rxdb.info/) OPFS storage you can build a fully featured database on top of the Origin Private File System (OPFS) browser API. Compared to other storage solutions, it has a way better performance.
+With the [RxDB](https://rxdb.info/) OPFS storage you can build a fully featured database on top of the [Origin Private File System](https://web.dev/opfs) (OPFS) browser API. Compared to other storage solutions, it has a way better performance.
 
 ## What is OPFS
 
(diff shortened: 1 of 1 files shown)


docs-src/docs/rx-storage-opfs.md
```diff
@@ -17,56 +17,64 @@ OPFS is ideal for applications requiring **high-performance** file operations (*
 
 From the beginning of 2023, the Origin Private File System API is supported by [all modern browsers](https://caniuse.com/native-filesystem-api) like Safari, Chrome, Edge and Firefox. Only Internet Explorer is not supported and likely will never get support.
 
-It is important to know that the OPFS API is **only available inside of a [WebWorker](./rx-storage-worker.md)**.
-It cannot be used in the main thread, an iFrame or even a [SharedWorker](./rx-storage-shared-worker.md).
-If you call the OPFS `getFileHandle()` function in the main thread, it will throw the error `Uncaught DOMException: A requested file or directory could not be found at the time an operation was processed.`.
```
pubkey, line 22: This was added so people will find that page when they google the error message

tomayac: Maybe, but it is factually not true:

pubkey: Oh. Yes, you are right.

tomayac: The OPFS in general is available on the main thread, just the synchronous options, that is, all methods on [`FileSystemSyncAccessHandle`](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemSyncAccessHandle), are restricted to Web Workers. These methods are the most performant ones, but because they're synchronous and potentially blocking, they're not exposed on the main thread.

If you don't need the absolute best performance, then you can perfectly work with the OPFS on the main thread and only ever use [`FileSystemFileHandle`](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemFileHandle). In your library, both might be possible depending on the use cases you need to cover.

pubkey:
@tomayac  Thank you very much. Merged.

PR #5757, opened 2024-03-10
fix: always use slice() for applying offset and limit in loki storage instance, fixing the fuzzing test run

Connum:
…

## This PR contains:
A BUGFIX

The fuzzy testing failed for the LokiJS storage, because sometimes there were less documents returned than expected. After quite some investigation, I found `limit()` and `offset()` to be the culprits, so instead of using those, slice() is now always used for limit and offset values, just as it already was when `mustRunMatcher` was `true`.

## Todos 
- [ ] Changelog


pubkey:
Hi. The changes look good. Can you add one test to the query-correctness tests which contains the broken fuzzed query+documents combination?

Connum:
Test has been added and verified that it fails before the fix and passes after.

Connum:
I guess the failing [example-graphq jobl](https://github.com/pubkey/rxdb/actions/runs/8246730820/job/22554467717?pr=5757#logs) must have been a hiccup during `npm install`, it shouldn't have anything to do with this PR, but a re-run might bring clarity.

pubkey:
Yes I will just retry the CI until it works.

Issue #5835, opened 2024-04-01
WebRTC Replication plugin should pass an RTCPeerConnection configuration to simple-peer


cleech:
I'm making use of WebRTC replication for shared documents between pairs of mobile users.  I'm getting reports that seem to indicate some users that can pair through the signaling server (my own based on the example code) but don't seem to be making peer connections.  I suspect both peers are on mobile networks blocking incoming connections and need a TURN server.

I'd like to be able to pass my own configuration object for the ICE agent, through simple-peer and the webrtc replication API.

pubkey:
I think there is an option already that is not documented. Check the code. Also a PR is welcomed, just make sure to not cause a breaking change.

cleech:
PR #5838 already opened :smile:

PR #5838, opened 2024-04-02
replication-webrtc: allow passing config to simple-peer

cleech:
Users may need to pass an RTCPeerConnection configuration object to simple-peer through the plugin. This simply adds an optional parameter to getConnectionHandlerSimplePeer.

This is mostly likely to be used to pass a custom ICE server list for STUN and TURN servers.

Issue #5835


pubkey:
Hi @cleech 
Code changes look good. Please also add this to the docs and to the changelog.

PR #5876, opened 2024-04-08
Update reactivity.md

maxfriedmann:
## This PR contains:
 - IMPROVED DOCS

2 changes were made: 

A) Wrapping the factory code with `untracked` prevents the NG0602 error as explained here: https://stackoverflow.com/questions/78291012/ng0602-when-using-rxdb-reactivity-factory

B) Adding `rejectErrors: true` prevents endless loops if the observable actually returns an error. In my case, this led to a crashing browser.

## Todos 
- [x] Documentation


pubkey:
Can you also add this to the angular example: https://github.com/pubkey/rxdb/blob/master/examples/angular/src/app/services/database.service.ts#L72

maxfriedmann:
Done. 

One note though: The example is not using angular signals at all: https://github.com/pubkey/rxdb/blob/master/examples/angular/src/app/components/heroes-list/heroes-list.component.ts#L34 So whats the point of providing a `reactivityFactory`? :) I was indeed wondering why your example worked and my local code didn't...

Maybe you could create a new example so that both, the rxjs and the angular signal version, are documented and tested?

pubkey:
It is used one line below for the `heroCount$$` value. I added this so we have at least one CI test for these signals. I am open for any PR to improve test coverage.

Issue #5925, opened 2024-04-23
When replicationWebrtc is used, each collection replication creates a websocket connection and several peers


croatialu:
![image](https://github.com/pubkey/rxdb/assets/22277972/217e8e02-7fb9-4d1c-b60f-46683eecd12a)

![image](https://github.com/pubkey/rxdb/assets/22277972/6600e071-7b33-4421-b935-dfce2ab5dac3)

PR #5926, opened 2024-04-23
Multiple collections share sockets and peers when replicationWebrtc is used for synchronization

croatialu:
## This PR contains:
 - A NEW FEATURE
https://github.com/pubkey/rxdb/issues/5925


croatialu:
You can experience the difference between before and after modification at this link

https://stackblitz.com/~/github.com/croatialu/rxdb-replication-webrtc

croatialu:
Solution: Use closure to save data such as peers and sockets, and then each replicationWebrtc will try its best to reuse existing peers and sockets


How to use:
``` ts
  const creator = getConnectionHandlerSimplePeer({})
  //                        ^ create a creator
  Object.keys(database.collections).map(async (key) => {
    return replicateWebRTC({
      collection:
        database.collections[key as keyof typeof database.collections],
      connectionHandlerCreator: creator,
      //                           ^ use the creator
      topic: dbName,
      pull: {},
      push: {},
    })
  })


```

pubkey:
Hi @croatialu 
Code changes look ok but this breaks the CI, please check.

croatialu:
> Hi @croatialu Code changes look ok but this breaks the CI, please check. Also we need at least one test to check the correctness of the new behavior.

Okay, I'll take care of it right away. Thanks for the reminder!

croatialu:
https://github.com/pubkey/rxdb/blob/48ea5eeacc72c88d5cecfe0c81c350470b7c2737/test/unit/replication-webrtc.test.ts#L161-L162



I found that this piece of code for the unit test uses data synchronization between collections of different names in the same database.  My idea is to synchronize collections of the same name from different databases.


My solution:
1.  When the collection of the database is changed, I will add a collectionName of the changed data to distinguish when sending data to other ends
2.  When the other end receives the data, according to the collectionName, the data is applied only when the collectionName is the same

The problem that the solution wants to solve is that the same collection data is synchronized on different ends.


So in an existing unit test, two different collectionnames cannot synchronize data.

Without modifying the unit tests, it didn't occur to me that different Collection synchronization schemes locally could also support the same Collection synchronization scheme between different ends

To pass the unit test, it will involve modifying the existing unit test method (changing to a different database with the same collectionName), which is likely to cause break change.

Do you have a better idea about that. @pubkey 




translate by youdao

pubkey:
@croatialu What is the status of this? Are you still workin on it?

croatialu:
@pubkey Sorry, I don't have extra time on this PR lately. If you need to copy collections with different names, or copy collections with different local names, then my PR does not meet the requirements. At present, my PR can only copy the same name collection of different hosts

Translated via youdao

pubkey:
Closing this. Please ping me when you want to continue then I can reopen.

Issue #5969, opened 2024-05-06
GraphQL query builder functions do not generate output fields for nested queries


KXLAA:
Faling test [PR](https://github.com/pubkey/rxdb/pull/5968)

You have a RX schema for a document in your collection with nested fields:

```ts
const mySchema = {
    version: 0,
    primaryKey: 'passportId',
    type: 'object',
    properties: {
        passportId: {
            type: 'string',
            maxLength: 100,
        },
        firstName: {
            type: 'string',
        },
        lastName: {
            type: 'string',
        },
        age: {
            type: 'integer',
            minimum: 0,
            maximum: 150,
        },
        updatedAt: {
            type: 'string',
        },
        address: {
            type: 'object',
            properties: {
                street: {
                    type: 'string',
                },
                city: {
                    type: 'string',
                },
                zip: {
                    type: 'string',
                },
            },
        },
    },
};
```
You want to generate GraphQL queries from this schema to support replication using the `pullQueryBuilderFromRxSchema`, `pullStreamBuilderFromRxSchema` & `pushQueryBuilderFromRxSchema` util functions.

Currently, the query builder functions only generate output fields for top-level keys. So given `mySchema` above we get the following pull query when we pass the schema to the `pullQueryBuilderFromRxSchema`:

```gql
query PullHuman($checkpoint: HumanInputCheckpoint, $limit: Int!) {
  pullHuman(checkpoint: $checkpoint, limit: $limit) {
    documents {
      passportId
      firstName
      lastName
      age
      updatedAt
      address 
      _deleted
    }
    checkpoint {
      passportId
      updatedAt
    }
  }
}
```
As you can see, the properties of `address` are not generated and this causes a GraphQL validation error :
```js
GraphQLError: Field \"address\" of type \"HumanModel\" must have a selection of subfields.
```

ideally, the function should be able to generate output fields for nested objects as well: 
```gql
query PullHuman($checkpoint: HumanInputCheckpoint, $limit: Int!) {
  pullHuman(checkpoint: $checkpoint, limit: $limit) {
    documents {
      passportId
      firstName
      lastName
      age
      updatedAt
      address {
        street
        city
        zip
      }
      _deleted
    }
    checkpoint {
      passportId
      updatedAt
    }
  }
}
```

PR #5976, opened 2024-05-08
feat: update GraphQL query builder functions to generate output fields for nested queries

KXLAA:
## This PR contains:
This PR contains A NEW FEATURE

You can see the issue this PR intends to fix [here](https://github.com/pubkey/rxdb/issues/5969). 

## Todos 
- [x] Tests
- [ ] Documentation
- [x] Typings
- [ ] Changelog


pubkey:
Looks like the ci fails because your test does not clean up its databases.

KXLAA:
> Looks like the ci fails because your test does not clean up its databases.

Do you know what is wrong with the supabase tests?

pubkey:
@KXLAA I don't think the breaking supabase test is your fault, it also randomly fails on master at the moment.
I merged this now to have it the next release which will come soon.

PR #6019, opened 2024-05-17
fix(datapath) must be equivalent for pull and push

jwallet:
## This PR contains:
 - IMPROVED DOCS
 - IMPROVED TESTS
 - A BUGFIX
 - BREAKING

If we define a custom `dataPath` for the `graphQL` replication, we end up setting a different datapath for the pull and push rep. Pull rep does not handle the `data.` prefix while push already handles it with `result.data` 

I changed this to make both (pull and push) use the same function helper to extract the proper datapath if defined or not by the user.

Based on the unit tests, I found some that were testing only the pull replication and needed `data` as a prefix so I made the helper function the same as the current `pull.dataPath`, so it will request the user to add `data.` as a prefix, if you prefer the other way around, let me know. However, it makes sense if for any reason the default selector `['data', Object.keys(result.data)[0])]` doesn't work because there is no `data` key returned from the server (exceptional edge case, maybe they used the `customFetch` property) and it breaks the default selector on `Object.keys(result.data)`

I also changed the typing of `dataPath` to accept `string[]` since `getProperty` also accepts it.

```ts
{
  dataPath: ['data', 'foo', 'bar']
}
  // or
{
  dataPath: 'data.foo.bar'
}
```

### DIFF
--- a/src/plugins/replication-graphql/helper.ts
+++ b/src/plugins/replication-graphql/helper.ts
@@ -1,5 +1,5 @@
 import type { RxGraphQLReplicationClientState, RxGraphQLReplicationQueryBuilderResponseObject } from '../../types/index.d.ts';
-import { ensureNotFalsy } from '../../plugins/utils/index.ts';
+import { ensureNotFalsy, getProperty } from '../../plugins/utils/index.ts';
 
 export const GRAPHQL_REPLICATION_PLUGIN_IDENTITY_PREFIX = 'graphql';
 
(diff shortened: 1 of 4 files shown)


pubkey:
Is this a breaking change?

jwallet:
Yes it is for the push replication. If someone used dataPath: 'pushNotifications' he will need to change it to 'data.pushNotifications'. Just like the pull dataPath.

pubkey:
Ok that makes it complicated. Atm there is no major release planned. I added the 16.0.0 label so it will be merged, but this will take a while.

Issue #6022, opened 2024-05-19
Can't resolve types


dongnaebi:
I am using RxDB in TypeScript, but I'm not getting any type hints. I looked in `dist/types/index.d.ts` and found this code:
```
export * from './plugin.ts';
export * from './rx-database.ts';
export * from './rx-error.ts';
export * from './rx-database-internal-store.ts';
// ...
```
It seems that the file names are incorrect. They should be `xxx.d.ts`?

pubkey:
I have no idea what you mean. But there are several unit tests that ensure these things work. For example in the angular RxDB example.

dongnaebi:
My point is: the types specified in `package.json` are `./dist/types/index.d.ts`, but the file names exporting these types in `./dist/types/index.d.ts` are incorrect, causing the editor(e.g. webstorm) to be unable to find these files, as shown in the picture:

![image](https://github.com/pubkey/rxdb/assets/6902417/6b375251-8f21-433e-8e89-da490d8fdad4)

pubkey:
Can you make a PR with a proposed fix?

dongnaebi:
OK, I think for now, the most cost-effective solution is to directly replace the content of `index.d.ts`

```js
// scripts/fix-type.mjs
import { promises as fs } from 'node:fs'

async function main () {
    const file = './dist/types/index.d.ts'
    try {
        let content = await fs.readFile(file, { encoding: 'utf-8' })
        content = content.replaceAll('export *', 'export type *').replaceAll('.ts', '.d.ts')
        await fs.writeFile(file, content)
    } catch (err) {
        console.log(`Fix type error：${err.message}`)
    }
}
main()
```

package.json
```diff
{
  "scripts": {
-     "build:types": "npm run build:version && rimraf -rf --max-retries=3 ./dist/types && tsc --project ./config/tsconfig.types.json && cp -r ./src/types ./dist/types/types",
+     "build:types": "npm run build:version && rimraf -rf --max-retries=3 ./dist/types && tsc --project ./config/tsconfig.types.json && cp -r ./src/types ./dist/types/types && node ./scripts/fix-type.mjs",
  }
}
```

dongnaebi:
Yes, it's work for me

![image](https://github.com/pubkey/rxdb/assets/6902417/828b9fea-869f-41a8-81e1-3dcd7638ca66)

pubkey:
@dongnaebi Thank you for the screenshot.
[...]

PR #6051, opened 2024-05-25
fix(types): modify file suffix after build types so that it can export correct types

dongnaebi:
## This PR contains:

 - A BUGFIX

fix: https://github.com/pubkey/rxdb/issues/6022

## Todos 
- [ ] Tests
- [ ] Documentation
- [ ] Typings
- [ ] Changelog


pubkey:
@dongnaebi  this fails the CI, please check.

dongnaebi:
Sorry I didn't notice this, it seems to be because I only exported the type. I removed the type, and the test:typings command has been tested successfully. I don't know how to test other CI, please run CI again to check.

pubkey:
I am not sure if we are working in the correct direction. Is this still correct typescript code which is generated? Why do we need ts-nocheck?

pubkey:
Ok, I merged this for now. In the future I want to drop the cjs export which simplifies the setup a lot and then maybe there is a better solution.

Issue #6141, opened 2024-06-24
replication websocket bug with composite primary key 


sir-script:
Hello everyone,

I am encountering different behaviors when streaming the replication.
In the first case (simple primary key), the user who modifies the document receives a subscribe event.
In the second case, where a composite primary key is used, the user who modifies the document does not receive a subscribe event. All other users/tabs receive the event.
Couldn't find the exact source code lines where the difference occurs while debugging.
The logic behind both cases is the same. I have checked this thoroughly and am quite sure that it is due to the primary key.
Does anyone have any idea or hint on what could be causing this?

Pseudo Code:
```
const replicationState = replicateGraphQL();
...
replication.received$.subscribe(() =>
  console.log('fired'),
);
```

### Case 1:
servicePlan (primary key with a single field)
```
export const servicePlanSchema = {
  version: 5,
  primaryKey: 'sp_id',
  type: 'object',
  properties: {
    sp_id: {
      type: 'string',
      maxLength: 20,
    },
    description: {
      type: 'string',
    },
    resort_id: {
      type: 'string',
    },
    status_id: {
      type: 'integer',
    },
    owner_id: {
      type: 'integer',
    },
    ts_completed: {
      type: 'string',
    },
    ts_assigned: {
      type: 'string',
    },
    ts_updated: {
      type: 'string',
    },
    sort_id: {
      type: 'integer',
    },
    ts_inserted: {
      type: 'string',
    },
  },
  required: ['sp_id', 'resort_id', 'ts_updated', 'ts_inserted', 'sort_id'],
};
```

Server Response:
```
{
  "id": "c98447b0-78c9-4ebb-9c0c-d5558c2c98c0",
  "type": "next",
  "payload": {
    "data": {
      "streamServicePlan": {
        "checkpoint": {
          "sp_id": "CHECKLIST_XXX_AM",
          "ts_updated": "2024-06-24T11:53:32.194Z",
          "sort_id": 64
        },
        "documents": [
          {
            "deleted": false,
            "sp_id": "CHECKLIST_XXX_AM",
            "description": "test plan",
            "resort_id": "8bbf66d8-d894-47df-a12e-c00f4c1f0e60",
            "status_id": 0,
            "owner_id": 15,
            "ts_assigned": "2024-06-24T11:53:32.194Z",
            "ts_completed": null,
            "ts_updated": "2024-06-24T11:53:32.194Z",
            "ts_inserted": "2024-05-15T13:52:57.130Z",
[...]

PR #6221, opened 2024-07-17
test for issue 6141

sir-script:
PR Test Case for https://github.com/pubkey/rxdb/issues/6141
In my opinion, both clients should receive same data (also client who made the modification, cause server may also update timestamp entries)

### DIFF
--- a/test/replication-couchdb.test.ts
+++ b/test/replication-couchdb.test.ts
@@ -6,13 +6,16 @@ import {
     humansCollection,
     ENV_VARIABLES,
     ensureCollectionsHaveEqualState,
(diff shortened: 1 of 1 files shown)


sir-script:
Hi @pubkey, thanks for reviewing it.

I would like to book a quick consulting session [Quick Consulting Session](https://rxdb.info/consulting/) to get more detailed info about this behaviour and other related concepts from rxdb.
Is it guaranteed that we will have this session with you if I book it online?

Cheers,
Michael

pubkey:
@sir-script Yes these sessions are with me in person.

PR #6227, opened 2024-07-18
Added initial checkpoint to graphql replication

lavaxun:
## This PR contains:
1. Initial checkpoint for GraphQL Replication: pull & push.
2. This PR is inspired by couchdb initial replication from another PR: https://github.com/pubkey/rxdb/pull/5461

1. Currently only the couchdb supported initial replication. However GraphQL replication does not support this.


pubkey:
Thank you for the PR. Can we add a test to that? Also docs must be updated.

pubkey:
Do you install premium for this PR? Because the changes are only for the core stuff and therefore you should not need premium.

PR #6238, opened 2024-07-23
Keep primary fields with an underscore

radxene:
For example, if we were to use couchbase entities/structures with _id property


pubkey:
Hi @radxene 
The change looks good. Please add at least one test to ensure this is not accidentially broken in the future.

radxene:
Hi @pubkey 

I just came up with this issue as a recommendation for future improvement, but unfortunately at the moment I don't have the free time to intervene with more contributions, if you wish you can omit this request and just do this improve yourself.

PR #6267, opened 2024-07-30
Bug example: key compression enabled with boolean indexes on dexie

gizm0bill:
Hello,
I'm opening this PR to demo a bug regarding key compression and boolean indexes with dexie storage.
So when key compression is added and there are some boolean indexes defined in the schema, no results are returned when using that index in a query.
The problem basically boils down to the compression flag and from/toStorage "dance" when inserting and querying documents

### DIFF
--- a/test/unit/bug-report.test.ts
+++ b/test/unit/bug-report.test.ts
@@ -8,134 +8,101 @@
  * - 'npm run test:node' so it runs in nodejs
  * - 'npm run test:browser' so it runs in the browser
  */
-import assert from 'assert';
-import AsyncTestUtil from 'async-test-util';
-import config from './config.ts';
 
-import {
-    createRxDatabase,
-    randomCouchString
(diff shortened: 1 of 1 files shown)


pubkey:
Primary key must always be a string.

gizm0bill:
Ok I have modified it but still fails. You can also look in the browser's indexeddb storage, indexes for `[_deleted+__a+passportId]` and `[_deleted+__b+__c+__a+passportId]` are empty

pubkey:
You did not set maxLength for the primary key. Please read the docs and enable the dev-mode plugin in your setup.

gizm0bill:
I have updated the unit, please take a look now

pubkey:
@gizm0bill Thank you. I can reproduce this now, I will check.

gizm0bill:
Danke!

PR #6315, opened 2024-08-10
WIP: upgrading to eslint 9.9.0

kadarsh11:

I apologize for raising the pull request before its completion. I encountered the following issue and thought someone might be able to assist:

```
Oops! Something went wrong! :(

ESLint: 9.9.0

A config object is using the "extends" key, which is not supported in flat config system.

Instead of "extends", you can include config objects that you'd like to extend from directly in the flat config array.

If you're using "extends" in your config file, please see the following:
https://eslint.org/docs/latest/use/configure/migration-guide#predefined-and-shareable-configs

If you're not using "extends" directly (it may be coming from a plugin), please see the following:
https://eslint.org/docs/latest/use/configure/migration-guide#using-eslintrc-configs-in-flat-config
```


pubkey:
Hi  @kadarsh11 
Thank you for the PR. This fails on the CI please check.

pubkey:
Hi @kadarsh11
Thank you for the PR. This fails on the CI please check.

PR #6348, opened 2024-08-25
feat: unnecessary space is deleted

webmaster1225:
- IMPROVED typings
    I don't think you need that kind of space.


pubkey:
We need this to ensure that the url is clickable in a terminal when the error object is printed as json.

PR #6405, opened 2024-09-18
RxDocument proxy performance issue

djoume:
## This PR contains:

A failing test to illustrate a performance issue with RxDocument

I've noticed a big performance degradation when adding rxdb to my app. A page that use to take a couple dozen ms to render started taking almost 200 ms. I tracked down the issue to the performance of RxDocument.proxy-get.

In my app mapping the list of RxDocuments I'm processing to regular javascript objects using:
```
myDocs.map((doc) => { return {...doc }}); 
```
Brought down rendering time from 200ms back to 30ms

### DIFF
--- a/package.json
+++ b/package.json
@@ -344,6 +344,7 @@
   "scripts": {
     "postinstall": "node scripts/postinstall.js || echo \"ignore\"",
     "test": "npm run test:node && npm run test:browser",
+    "test:bugreport": "npm run transpile && cross-env DEFAULT_STORAGE=memory NODE_ENV=fast mocha --config ./config/.mocharc.cjs ./test_tmp/unit/bug-report.test.js",
     "// test:fast": "run tests in the fast-mode. Most of them will run in parrallel, skips tests that are known slow",
     "test:fast": "npm run test:fast:memory && npm run test:fast:lokijs && npm run test:fast:dexie",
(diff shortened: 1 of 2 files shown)


pubkey:
We have `RxDocument.toJSON()` to get the plain document data as an object.

Javascript proxies have a performance penalty and I am not sure if this can be fixed on the RxDB side. In the past we did not use proxies but plain objects and when I moved to proxies I did some performance tests and everything was fine. Is your problem the case in all browsers?

djoume:
I suspect there might be something wrong with the proxy becasue would have expected a ~20% performance hit on property access but I'm seeing 2x slower on top level fields and 9x on nested field. 

```
Accessing top level field
---------------------------------
    RxDocument: 8.55ms
Regular object: 4.62ms

Accessing nested field
---------------------------------
    RxDocument: 93.47ms
Regular object: 10.53ms
```

I also noticed that the performance of nested field gets worst with the number of iterations.

I've confirmed this behavior on Node, Chrome and Firefox.

pubkey:
I played around with the code. I think I found something that would be faster for most cases. Please check https://github.com/pubkey/rxdb/pull/6406

djoume:
What's the best way to test this code change in my project?

pubkey:
improved here: https://github.com/pubkey/rxdb/pull/6406

PR #6406, opened 2024-09-19
IMPROVE performance of nested property access

pubkey:
This PR tries to improve performance on nested property access: https://github.com/pubkey/rxdb/pull/6405


djoume:
On the problematic page on my app, this change brings the page load time from 750ms down to 150ms. Using toJSON() brings it down to 85ms so that's what I will be using.

pubkey:
Ok thank you for testing @djoume I will still merge this I think it is still an improvement even when toJSON is faster in your case.

PR #6436, opened 2024-09-23
Fixes firestore replication of multiple document with the same serverTimestamp

djoume:
## This PR contains:
 - A BUGFIX

I noticed some documents where missing when replicating from firestore. I traced down the issue to a bug in the `sameTimeQuery` in the firestore plugin where we're querying for a document field instead of querying for a document id.

I'm not sure how to test a change to this library on my project, could you advise on how to do so? I don't think I can use `npm link` because of the build steps required.

### DIFF
--- a/src/plugins/replication-firestore/index.ts
+++ b/src/plugins/replication-firestore/index.ts
@@ -138,8 +138,8 @@ export function replicateFirestore<RxDocType>(
                     );
                     sameTimeQuery = query(pullQuery,
                         where(serverTimestampField, '==', lastServerTimestamp),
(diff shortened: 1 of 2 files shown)


pubkey:
To test this, clone the rxdb project into your `node_modules/rxdb`. Then run `npm run build` inside of it.

djoume:
I finally found what was preventing me for testing changes to rxdb, I had forgot to exclude rxdb from vite's optimizeDeps config.

I have been able to confirm that with this change all documents are being fetched, let me know if there is anything else I can do to help get this bugfix merged.

pubkey:
Thank you, merged.

PR #6458, opened 2024-10-03
Update alternatives.md

beorn:
Add more alternatives: InstantDB, ElectricSQL, SignalDB, PowerSync, SQLSync

## Todos 
- [ ] Tests
- [ ] Documentation
- [ ] Typings
- [ ] Changelog


pubkey:
Some parts of these texts are wrong. Did you use ChatGPT?

pubkey:
Its like everything is just slightly wrong a bit. Please write it by yourself.
As alternative we can just add a link-list at the bottom so that we at least do not missinform readers.

beorn:
I'll go over it again and try to fact-check it more, and maybe make some entries a bit shorter.  I wish there was another more comprehensive list of local-first technologies that you could point to, but this one is actually the most comprehensive I've seen so far :)  In any case, it's useful to compare them to RxDB. It's also fast-moving field so my worry it'll be outdated very quickly.

beorn:
I made another round of edits.  It's somewhat difficult to compare and contrast the alternatives given they are so different and (often) flexible, and giving a pros/cons list is ultimately going to be a bit subjective.  Perhaps there are other things you want to mention or leave out, or emphasize/deemphasize.

beorn:
This comparison maybe interesting as it highlights internals of PowerSync and Electric-Next - https://www.powersync.com/blog/electricsql-electric-next-vs-powersync

pubkey:
Thank you, this looks way better. Merged.

PR #6503, opened 2024-10-15
test: bad rx-state after cleanup

enrico-teterra-maramalabs:
## This PR contains:
a failing test

to reproduce the issue:
- have an rx-state
- make more than 5 writes to ensure _cleanup does its thing
- the state has an additional "" property which causes problems further down the line

### DIFF
--- a/test/unit/rx-state.test.ts
+++ b/test/unit/rx-state.test.ts
@@ -488,29 +488,20 @@ addRxPlugin(RxDBJsonDumpPlugin);
             });
         });
         describe('issues', () => {
-            it('createRxState with cleaned up operations', async () => {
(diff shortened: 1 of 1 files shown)


pubkey:
Can you make the test by only using the "normal" RxDB api without accessing the internal storageInstance?

enrico-teterra-maramalabs:
yep no problem, please take a look at the test now 

the test as it is now fails with:

```
+ expected - actual

       {
      -  "": {
      -    "foo": "bar6"
      -  }
         "foo": "bar6"
       }
```

pubkey:
Thank you for reporting. Fixed here https://github.com/pubkey/rxdb/pull/6564

PR #6516, opened 2024-10-18
fix: position of dev-mode iframe

msotnikov:
This patch removes 5 extra pixels in dev-mode
Screenshot 2024-10-18 at 10 20 22


pubkey:
Thank you, merged.

Issue #6520, opened 2024-10-22
failed to replicate Large document Attachments over WebRTC in browsers


haouarihk:
## Have you found a bug?
WebRTC in browsers is only limited to 16kb,
however this limit is not present in nodejs environment.

## Did you make a unit test?
As far as i can tell, there are no browser based unit test in this project, but i made the test anyway for future implementation.

## Do you have a feature request?
The solution to this problem is to send large contents in chunks, rather than sending them all as one blob. this could also open us for another event listener for "syncing".

if i missed anything, be sure to let me know.

PR: https://github.com/pubkey/rxdb/pull/6521

haouarihk:
here's the test file:
```ts
import assert from 'assert';
import config from './config.ts';
import {
    randomCouchString,
    createRxDatabase
} from '../../plugins/core/index.mjs';
import {
    replicateWebRTC,
    getConnectionHandlerSimplePeer,
    SimplePeerWrtc
} from '../../plugins/replication-webrtc/index.mjs';
import {
    isFastMode,
    isDeno,
    isNode
} from '../../plugins/test-utils/index.mjs';
import { waitUntil } from 'async-test-util';
import { wrappedAttachmentsCompressionStorage } from '../../plugins/attachments-compression/index.mjs';

describe('large-document-webrtc.test.ts', function () {
    this.timeout(1000 * 60); // 1 minute

    if (
        !config.storage.hasReplication ||
        !config.storage.hasPersistence ||
        !config.storage.hasAttachments ||
        isDeno
    ) {
        return;
    }

    let wrtc: SimplePeerWrtc;
    let webSocketConstructor: WebSocket;
    const signalingServerUrl = 'ws://localhost:18006';

    before(async () => {
        if (isNode) {
            const wrtcModule = await import('node-datachannel/polyfill');
            wrtc = wrtcModule.default as any;

            const wsModule = await import('ws');
            webSocketConstructor = wsModule.WebSocket as any;
        }
    });

    it('should replicate large attachments over WebRTC', async function () {
        if (isFastMode()) {
            return;
        }

        const createCollection = async (name: string) => {
            const db = await createRxDatabase({
                name: randomCouchString(10),
                storage: wrappedAttachmentsCompressionStorage({
                    storage: config.storage.getStorage()
                })
            });

            const collections = await db.addCollections({
                [name]: {
                    schema: {
                        version: 0,
                        type: 'object',
                        primaryKey: 'id',
                        properties: {
                            id: {
[...]

PR #6521, opened 2024-10-22
#6520 Implementation for chunking in WebRTC  Plugin

haouarihk:
## This PR contains:
- Added chunking to messages sent using simple-peer to handle large data.

Related: https://github.com/pubkey/rxdb/issues/6520


haouarihk:
if there are any changes you want me to make, please let me know

pubkey:
Hi @haouarihk  . Please also add some tests for this.

pubkey:
@haouarihk Your fix looks good. Can you add some tests?

haouarihk:
As I mentioned in the issue.
Its not possible. Because it needs browser tests.
This issue only appear in browser environment

pubkey:
@haouarihk you are right. So the first step would be to add some browser based tests.

PR #6589, opened 2024-11-15
test: bulk inserting 2 docs with same primary key

enrico-teterra-maramalabs:
## This PR contains:
a failing test

bulk inserting 2 docs with same primary key should throw conflict error
instead, the collection now contains 2 docs with the same primary key

example:
* given a collection: `people` has a primary key: `id`
* when `people.bulkInsert({id: 1, name: bob}, {id: 1, name: jane})` 
* then the bulkInsert is successful and now i have 2 records with the same primary key in the collection
* `people.count().exec()` is 2

### DIFF
--- a/test/unit/rx-collection.test.ts
+++ b/test/unit/rx-collection.test.ts
@@ -252,6 +252,27 @@ describe('rx-collection.test.ts', () => {
                 });
             });
             describe('negative', () => {
+                it('ensures uniqueness of primary-key', async () => {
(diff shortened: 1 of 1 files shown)


pubkey:
Thank you for reporting. I added a fix. I think the correct behavior should be to just throw, RxDB should not be used in that way.

enrico-teterra-maramalabs:
yep throwing an error in this case seems fine! thanks

PR #6592, opened 2024-11-15
fix(rx-query) findByIds not working with modify and patch

HRM:
## Content
* Unit tests that shows the bug
* Fix for the bug
## Description
When you chain "findByIds" with "patch" or "modify" it throws an error like "there is not method named modify on doc". The issue is caused by "findByIds" returning a result in the form of a map witch is not handled in "runQueryUpdateFunction".


pubkey:
Thank you very much @HRM . Merged.

PR #6598, opened 2024-11-18
added possibility to pass graphql-ws options into the replication options

epodgaysky:
This PR contains:
 - IMPROVED typings
 - A NEW FEATURE

In our project we need more flexibility with replication subscription, particularly we want it to try reconnecting until the connection is established, even if server is not responding.

### DIFF
--- a/test/unit/replication-graphql.test.ts
+++ b/test/unit/replication-graphql.test.ts
@@ -1635,6 +1635,67 @@ describe('replication-graphql.test.ts', () => {
                 await server.close();
                 await c.database.destroy();
             });
+            it('should respect pull.wsOptions', async () => {
+                const capturedWSStates = [];
+                const [c, server] = await Promise.all([
+                    humansCollection.createHumanWithTimestamp(0),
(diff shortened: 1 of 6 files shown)


pubkey:
Hi @epodgaysky 
The changes look good. We need at least one test and this should be added to the docs.

test/unit/replication-graphql.test.ts
```diff
@@ -1635,6 +1635,67 @@ describe('replication-graphql.test.ts', () => {
...
+                    },
+                    live: true,
+                    deletedField: 'deleted',
+                });
+
+                ensureReplicationHasNoErrors(replicationState);
+
+                await replicationState.awaitInitialReplication();
+
+                await wait(300);
```
pubkey, line 1682: Please do not do waits by times inside of tests. This will slow down the whole test suite.

pubkey:
Thank you @epodgaysky  merged!

pubkey (commented):
looks good

PR #6643, opened 2024-11-27
Inserting without an optional index field causes unexpected behavior

OskarD:
## This PR contains:
A BUG REPORT

Documents inserted with a (not required) empty index field behave unexpectedly

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Typings
- [ ] Changelog

### DIFF
--- a/test/unit/bug-report.test.ts
+++ b/test/unit/bug-report.test.ts
@@ -9,7 +9,6 @@
  * - 'npm run test:browser' so it runs in the browser
  */
 import assert from 'assert';
-import AsyncTestUtil from 'async-test-util';
 import config from './config.ts';
 
 import {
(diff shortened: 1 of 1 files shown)


test/unit/bug-report.test.ts
```diff
@@ -9,7 +9,6 @@
  * - 'npm run test:browser' so it runs in the browser
```
OskarD, line 9: This script does not exist

pubkey:
Thanks for the PR. Indeed it looks like there is a bug. I will check.

pubkey:
Ok so the query plan looks correct:

```json
"preparedQuery": {
        "query": {
            "selector": {
                "_deleted": {
                    "$eq": false
                }
            },
            "sort": [
                {
                    "id": "asc"
                }
            ],
            "skip": 0
        },
        "queryPlan": {
            "index": [
                "_deleted",
                "numberIndex",
                "id"
            ],
            "startKeys": [
                false,
                -9007199254740991,
                -9007199254740991
            ],
            "endKeys": [
                false,
                "￿",
                "￿"
            ],
            "inclusiveEnd": true,
            "inclusiveStart": true,
            "sortSatisfiedByIndex": false,
            "selectorSatisfiedByIndex": true
        }
    }

```

I played around with the generated keyrange but I am not sure if this is even fixable.

pubkey:
I found out:
As David Fahlander from dexie.js explains [here](https://github.com/dexie/Dexie.js/issues/1019), `undefined` is not an [indexable type](https://github.com/dexie/Dexie.js/issues/1019). So this cannot be fixed. We should add a check to the dexie.js RxStorage to throw a proper error when this type of schema is used. Just returning wrong results is not a correct behavior.

The premium IndexedDB RxStorage from rxdb does not have this problem because it uses [custom index strings](https://rxdb.info/slow-indexeddb.html#custom-indexes), not the composite index keys from native IndexedDB.

Issue #6515, opened 2024-10-18
Websocket hangs after connection is closed by the server


brunovinicius:
GraphQL Websocket Plugin hangs and does not retry connection when server closes connection right after a successful 101 Upgrade. On my scenario this manifests itself as the client try to resume conection after refreshing auth token. WS connection is restablished way before the token is refreshed so it tries to connect using the old, expired token. After this, it gets stuck in this limbo as the server closes connection as soon as it validates the token.

Sequence of events looks similar to this:

=> GET wss://... 
<= 101 Switching Protocols
=> WS Message: connection_init (with the expired token)
<= Server Closes connection

It seems like the client its not expecting this behavior and stays hanging in wait for the server to respond "connection_ack" which will never come. This scenario might arise due to a unfortunately timed network issue, but any directions on how my server could deal with this appropreately would be very much appreciated.

RxDB Version: 15.17

Thanks in advance and sorry for not providing a PR with a test case right away.

brunovinicius:
up

pubkey:
Do you think it is possible to build a unit test to reproduce this?

brunovinicius:
Hey @pubkey hope you are feeling better! 

We went live with our project a couple weeks ago and we are sorting out issues so not a lot of time available for creating this test, but I expect to put some focus on it soon. 

We developed a patch using patch-package that helped a bit, but didnt fully resolve the issue. We can make a PR of this change really quickly if you want to take a look at it. Let me know.

cayres:
Hi @pubkey, I'm working on the same project as @brunovinicius. Our patch is really simple and based on the ws client events we found here: https://the-guild.dev/graphql/ws/docs/modules/client#event.
```
diff --git a/node_modules/rxdb/dist/esm/plugins/replication-graphql/index.js b/node_modules/rxdb/dist/esm/plugins/replication-graphql/index.js
index 56d6c7b..65e8e0e 100644
--- a/node_modules/rxdb/dist/esm/plugins/replication-graphql/index.js
+++ b/node_modules/rxdb/dist/esm/plugins/replication-graphql/index.js
@@ -123,6 +123,17 @@ export function replicateGraphQL({
       wsClient.on('connected', () => {
         pullStream$.next('RESYNC');
       });
+
+      wsClient.on('closed', (error) => {
+        pullStream$.error(error);
+        graphqlReplicationState.cancel()
+      });
+
[...]

PR #6644, opened 2024-11-27
Unit test for reproduction of issue scenario #6515

cayres:
## This PR contains:

Unit test for reproduction of issue scenario #6515

Today there is no error or event published when a client loses its websocket connection. Then we can't know and try to fix the problem, and that user doesn't get live updates anymore.

### DIFF
--- a/test/unit/bug-report.test.ts
+++ b/test/unit/bug-report.test.ts
@@ -9,133 +9,202 @@
  * - 'npm run test:browser' so it runs in the browser
  */
 import assert from 'assert';
-import AsyncTestUtil from 'async-test-util';
 import config from './config.ts';
 
 import {
     createRxDatabase,
-    randomToken
+    randomToken,
+    RxReplicationWriteToMasterRow,
 } from '../../plugins/core/index.mjs';
 import {
-    isNode
+    HumanWithTimestampDocumentType,
+    isNode,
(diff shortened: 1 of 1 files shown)


pubkey:
I understand the use case.
I think this is more like feature request then a bug report.
Please make a PR if you need that feature.

brunovinicius:
Hey Daniel! Thanks for the update. I think you are correct in that adding this event as something that we can use to react its a feature, yes. But I feel the behavior on the RxDB side is still kinda broken, I would argue.

You see, from RxDB perspective, if the websocket connection fails or gets dropped, a new one is created to re-stablish communication. I would argue this scenario is one that should cause the same behavior as when a GraphQL subscription gets dropped.

pubkey:
Dropped subscriptions that reconnect could lead to missed events on the client that happen during this time. Please consider that case when implementing.

cayres:
Hi Daniel, I have started coding this feature and realized that this was solved in the PR #6598. With the exposed wsoptions it is possible to listen to the closed events and reconnect. So I think this issue can be closed.

cayres:
Hi Daniel, I have started coding this feature and realized that this was solved in the PR #6598. With the exposed wsoptions it is possible to listen to the closed events and reconnect. So I think this issue can be closed.

PR #6690, opened 2024-12-17
Storage and Schema Migration Combo Test

tsekiguchi:
## This PR contains:
A failing test that incorporates both storage migration and schema migration.

Updating schema in between storage migrations _without_ updating the version numbers is currently allowed behavior. 

Then, when trying to then implement a normal schema migration strategy afterwards, the migration strategy starts throwing error unknown migration error, DM4.

```ts
// this is where the test error occurs
const v1collection = await db_newStorage_schemav1.addCollections({
    mycollection: {
        schema: mySchemaV1,
        migrationStrategies: {
            1: (d) => d,
        },
    },
});
```

I tried to see if `getOldCollectionsMeta()` was potentially causing this, but it wasn't returning anything in this context. 

## Todos 
- [x] Tests
- [ ] Documentation
- [ ] Typings
- [ ] Changelog

### DIFF
--- a/test/unit/migration-schema.test.ts
+++ b/test/unit/migration-schema.test.ts
@@ -2,7 +2,11 @@ import assert from 'assert';
 import config, { describeParallel } from './config.ts';
 import AsyncTestUtil, { waitUntil } from 'async-test-util';
 
-import { humansCollection, schemaObjects, schemas } from '../../plugins/test-utils/index.mjs';
+import {
+    humansCollection,
+    schemaObjects,
+    schemas,
+} from '../../plugins/test-utils/index.mjs';
 
 import {
     createRxDatabase,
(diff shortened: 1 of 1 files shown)


pubkey:
I tried to add a check but this likely is not possible.
The problem is that there are many reasons to change the schema during storage-migration. For example when you migrate to an encrypted storage or start using key-compression. So some schema changes would be ok and some would not.
I think I will just add a warning about this into the docs.

tsekiguchi:
Got it! What about adding a requirement to increment schema version numbers when migrating storage? That would help prevent collisions and unknown behavior down the road. I mean, technically you're changing all the schema when changing the underlying storage. 

It could be a simple check for migrationNeeded() in all schema? 

And yes, adding more info in the docs about this would be great! Beyond just switching the underlying storage, I think it's important to spell out all the scenarios that require a storage migration - adding attachments for a schema, key compression, etc. - that can't be accomplished by just a schema migration.

pubkey:
Adding attachments or key-compression do not always require a storage-migration. It can be done in a schema migration. But often it makes sense to add this when already doing a storage-migration.

The thing is that the storage-migration does not have to know about the new collections and does not store collection metadata into the new database. This is important because we cannot store old schemas when doing a storage-migration to upgrade between major RxDB versions. The schemas might not be compatible.

For me its important to have the storage-migration as flexible as possible so it works across major versions and all storage combinations.

pubkey:
I checked the code, we already use the new-collections storage instance. This also uses the schema validation.
I will now add a warning to the docs to not change the schema during storage migrations. That should be enough.

PR #6788, opened 2025-01-28
ADD test for #6786

pubkey:


pubkey:
Closed because does not fail, see https://github.com/pubkey/rxdb/pull/6786

PR #6840, opened 2025-02-09
Fix/notification popup

tsekiguchi:
Notification popup key was being stored with Date.getTime() embedded directly in the key. Unfortunately this won't work because every time a user refreshes the page or navigates to a new page, the Date.getTime() is different, and therefore localStorage is unable to find the entry for when the popup was closed, so it thinks that it hasn't happened yet.

Evidence of this is seeing dozens of notification_popup_closed_at values in the local storage
Screenshot 2025-02-09 at 11 50 16

To fix this, we can store the time as a value, we can simply look up the `notification_popup_closed_at` key, and then perform the comparison on the time instead.


pubkey:
Thank you merged. I have no idea why I had it implemented so complicated :)

PR #6876, opened 2025-02-18
Fix foundationdb premium task

nickchomey:
FIxing the premium task for upgrading node-foundationdb to v2.0.1.

Local testing suggests that the only thing that is necessary is installing foundationdb client and server v7.3.xx. You can then specify any apiVersion _except_ 730. 

Ive opened an issue at node-foundationdb about this https://github.com/josephg/node-foundationdb/issues/85

Hopefully the tests will pass with this PR


pubkey:
I have to manually trigger them until one of your PRs has been merged.

PR #6945, opened 2025-03-07
Draft: Turn off node integration

davetorre:
## This PR contains:
A demonstration that nodeIntegration can be false (and contextIsolation can be true) when using RxDB in Electron.
It seems that one reason (hopefully the only reason) it is set to true today is because of the way node_modules are being imported in the renderer.js code. Adding webpack seems to let me turn off nodeIntegration and the example still works.
Can someone more familiar with the RxDB take a look and determine if this is correct?

There are warnings in the Electron docs against enabling nodeIntegration and disabling contextIsolation: https://www.electronjs.org/docs/latest/tutorial/security


pubkey:
Closed because the CI fails and no answer.

Issue #6988, opened 2025-03-21
DB8 Error suggestion


KingSora:
Good day!

We have a React-App where we use RxDB to store user related information. This means each user has its own db (we use the userID inside the db name). The RxDB is initialized and cleaned up inside an `effect`. A simple example code looks like this:

```js
const [rxdbInstance, setRxdbInstance] = useState(null);
useEffect(() => {
  const abortController = new AbortController();
  const rxdbInstancePromise = createRxDatabase({
    name: `prefix-${userId}-suffix`
  }).then((rxdbInstance) => {
    if (!abortController.signal.aborted) {
      setRxdbInstance(rxdbInstance);
    }
  }).catch(() => {
    if (!abortController.signal.aborted) {
      setRxdbInstance(null);
    }
  });

  return () => {
    abortController.abort();
    rxdbInstancePromise.then((rxdbInstance) => rxdbInstance.destroy());
  }
}, [userId])
```

This code runs perfectly fine and each rxdb instance is cleaned up in case the `userId` changes. The problem here is that initialization and destruction is `async`. If the `userId` would quickly change like this:
1. user-a
2. user-b
3. user-a 

Then we would encounter a `DB8` error for the second initialization of `user-a` because the first initialization might not be destroyed yet. This scenario is theoretical in `production` but very real in `development` because React runs effects twice because of its strict mode.

This behavior is a _huge_ DX issue for us. To fix this issue we implemented a tracking system where we always wait for a rxdb with the same name and storage to destroy before a new one can be intialized, but for some time now we observe our users in production getting this error right after deployments.

----

Going back to what the `DB8` error actually means and tries to solve: Having two different instances of the same RXDB being initialized _and being used_ at the same time. This would be easy to track if everything would be snychronous but is harder if initialization and destruction are async, because it actually takes time until one instance is ready / destroyed.

I would have three suggestions to improve this behavior but keep the semantics in place:

1. The first solution would be to again allow the option `ignoreDuplicate` being true without the need of the `dev-mode` plugin. (personally not a fan of this solution, although its the simplest to implement)
2. The `DB8` error doesn't happen anymore on initialization but only if the instance actually being used. (So having two instances at the same time is fine, but only the instance which was initialized later can actually be used without this error happening)
[...]

PR #6995, opened 2025-03-23
feat: db8 error handling

KingSora:
## This PR contains:
 - A NEW FEATURE
 - A BREAKING CHANGE (kind of?)

The problem is described and discussed here: https://github.com/pubkey/rxdb/issues/6988

This PR makes the `DB8` erro obsolete because a new instance with the same db-name and storage will only initialize when no other with the same combination exists:

```js
const db1Promise = createRxDatabase({ name, storage });
const db2Promise = createRxDatabase({ name, storage }); // this promise will only resolve once db1 is closed and db2 is fully initialized
```

In the issue above I've also mentioned to implement a `AbortSignal` parameter which is not contained in this PR yet as I want to first wait for feedback.

## Todos
- [x] Tests
- [ ] Documentation
- [x] Typings
- [ ] Changelog

### DIFF
--- a/dist/types/plugins/dev-mode/error-messages.d.ts
+++ b/dist/types/plugins/dev-mode/error-messages.d.ts
@@ -43,7 +43,6 @@ export declare const ERROR_MESSAGES: {
     DB4: string;
     DB5: string;
     DB6: string;
-    DB8: string;
     DB9: string;
     DB11: string;
     DB12: string;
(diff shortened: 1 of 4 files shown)


pubkey:
I do not understand, this removes the DB8 check completely. What happens if I open the same db twice without ever closing the first one?

KingSora:
> I do not understand, this removes the DB8 check completely. What happens if I open the same db twice without ever closing the first one?

So instead of throwing the DB8 error the `createRxDatabase` promise only resolves as soon as all other instances with the same name + storage are closed. I've posted an example in the description above.

pubkey:
oh, I accientially merged this. Sorry, have to revert first.

pubkey:
From your description, does that mean that if I call `createRxDatabase` on the same db twice, the second call will silently never resolve without showing an error? This would confuse people I think.

KingSora:
@pubkey Thats correct and I've tought about this too.. There would be a few possibilities to solve that:
- In the place of the `DB8` error we could print a `warning` in the console (in dev mode) which describe the behavior
- There is an explicit option for this new behavior which can be enabled, something like the `ignoreDuplicate` option just for this behavior

Obviously this should be documented in the docs.. What do you think?

Issue #6938, opened 2025-03-06
Tauri + SQLite hero-example


jameshclrk:
From the [Premium Tasks](https://github.com/pubkey/rxdb/blob/master/orga/premium-tasks.md).

I'll have a PR up in an hour or two.

pubkey:
Hi @jameshclrk 
We have some problems running this in the CI in the long term:
- Please pin the dependency versions, (no `^`)
- Please add `npm run build` to the CI
- Please do not commit a package-lock.json. If the tauri example is not working on the "normally installed deps", the CI should fail because likely it is broken for other RxDB+tauri users.

PR #7061, opened 2025-04-05
Fix/jameshclrk/tauri example

jameshclrk:
## This PR contains:

Improved Tauri tests

 Should fix issues mentioned in #6938


pubkey:
Can you add the build to the CI script?

jameshclrk:
> Can you add the build to the CI script?

It's run as part of `npm run test`, do you want it on its own before this step? https://github.com/pubkey/rxdb/blob/master/.github/workflows/main.yml#L593-L597


> Also this does not remove the package-lock

I'm not sure what you mean, there's no package-lock in the repo

pubkey:
I see. This looks ok then, merged.

Issue #7069, opened 2025-04-09
rx-state may cause test case errors in extreme cases.


kryon-7:
rx-state.test.ts at the `write with two states at once` case 
This case may occasionally fail, but the probability is relatively low. Therefore, I created a new case to simulate a more extreme scenario.

kryon-7:
However, this issue remains difficult to reproduce. On average, it takes 4-5 attempts to trigger the error. However, multiple runs don't seem to execute sequentially—instead, they appear to run asynchronous functions concurrently, leading to race conditions during collection creation. Do you have any suggestions?

pubkey:
Hi @kryon-7 
Does the test `write with two states at once` occasionally fail on all storages or only on the random-delay storage?

kryon-7:
As I mentioned in #7070, this is a probabilistic event that depends on whether the computer yields execution after this.collection.insert completes. There's an extremely low chance that the computer might not return control to the main process but instead invoke other asynchronous functions. Because rx-state operations execute very quickly, and some storage mechanisms have transactional features, this issue is particularly hard to reproduce. The best approach is to add a 10-millisecond delay with await wait(10) after this.collection.insert finishes. For memory storage, you don't even need a 10-millisecond delay—just await wait() with 0 milliseconds is enough to yield to other asynchronous operations and trigger the error.
The probability is extremely slim because concurrency issues are inherently probabilistic, and the officially provided storage plugin code is quite robust. I think it would be very difficult to reproduce.

kryon-7:
hi! @pubkey 
After careful consideration, I find it quite challenging to accurately reproduce this error without making destructive code changes. First things is We need to reach a consensus on the following points.

1. The following code segment has the potential to pause execution, yielding the main thread to other asynchronous operations (such as state change listeners using pipes that might modify the state during this window):

```typescript
await this.collection.insert({
    id: nextId,
    sId: this._instanceId,
    ops
});
```
2. Assuming we intentionally introduce a delay after successful insertion (e.g., 10ms), should the system maintain correct behavior? In other words, is the following implementation expected to work reliably?

```typescript
await this.collection.insert({
    id: nextId,
    sId: this._instanceId,
    ops
});
await wait(10); // Artificial delay
this._state = newState;
```

pubkey:
What do you mean with "pause execution"?
.insert is async and while it runs, other stuff can happen.

[...]

PR #7070, opened 2025-04-09
Add a test case for an extreme scenario

kryon-7:

Because there is a test case that may fail intermittently (uncertain) #7069 , I am now adding more extreme test cases.

### DIFF
--- a/test/unit/rx-state.test.ts
+++ b/test/unit/rx-state.test.ts
@@ -360,6 +360,20 @@ addRxPlugin(RxDBJsonDumpPlugin);
                 state1.collection.database.close();
                 state2.collection.database.close();
             });
+            it('write with ten states at once', async () => {
+                const databaseName = randomToken(10);
+
+                const list: RxState<TestState, ReactivityType>[] = [];
+                for (let i = 0; i < 10; i++) {
+                    const state = await getState(databaseName);
+                    list.push(state);
+                }
+                await list[0].set('a',()=>0);
+                await Promise.all(list.map(state=> state.set('a', plusOne )));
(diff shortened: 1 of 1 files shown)


kryon-7:
@kryon-7 hello!I have identified the cause. It's because the handling of external modifications and internal modifications are not in the same queue. When creating an rx-state, an Observer is created to monitor external modifications to the collection, while internal modifications are processed using the _writeQueue. The execution order between these two is uncontrollable, which leads to concurrency issues. Theoretically, this problem should occur in all cases, but perhaps the inconsistent speed of processing internal and external modifications in other real storage systems makes the error harder to reproduce.
Below is our expected flow.
image
But Scenario:
Assume StateA first calls insert() to set num=1 and succeeds.

However, we know the insert method's workflow:

First, it actually stores the data in the storage.

Then, it broadcasts a notification.

Finally, it returns.

Race Condition:
Once the data is stored in the storage, other states (e.g., StateB) can modify it.

At this point, StateA’s insert() request might not have returned yet, but StateB has already started modifying the data (e.g., setting num = 1 + 1 = 2).

The storage applies StateB’s change (num=2) and broadcasts the update.

StateA receives the notification and updates its num to 2.

The Problem:
But await insert() in StateA still hasn’t resolved! (This is possible!)

Meanwhile, StateA has already accepted num=2 from StateB’s update.

When StateA’s await insert() finally returns successfully, it reverts num from 2 back to 1, causing inconsistency!

image
I already know how to fix this issue to prevent it from occurring, but accurately reproducing the error will require more time to study the source code and write non-intrusive tests. Sorry I haven't had time recently, but I'll write a proper test case for this bug in the next few days.

Issue #7260, opened 2025-07-01
After adding a new property to an existing schema, no document can be retrieved from the InputStream


alpha-2000:
I'm using rxDB 16.15. A new field was added to the schema of a collection that was migrated with 
```ts
2: (oldDoc) => oldDoc
```
After that, my collection can no longer receive changes via the remote's InputStream. I log my document in the modifier before returning it there. Normally, I should see it in the received field afterward, but nothing happens. No errors, no warnings.
My setup is also very minimalistic. I use the following when creating: 
```ts
db = await createRxDatabase<Database>({
name: dbName,
storage: wrappedValidateAjvStorage({
storage: getRxStorageIndexedDB(),
});,
localDocuments: true,
multiInstance: true,
});
```

I've tried performing the migration automatically or starting it manually with:
```ts
const needed = await col.migrationNeeded();
if (needed) col.startMigration(1);
```

But it doesn't make any difference.

This only happens when I add a new field to the schema. If I change the type of an existing field and migrate it, everything works fine. The new field I add isn't written to the required field either.

alpha-2000:
After switching the schema validator from ajv to z, an error occurred during migration. This is an SNH error in the RxReplicationProtocolMetaData validation errors. 
```ts
validationErrors: Array(1)
  0: 
    description: undefined
    message: "Expected type object but found type undefined"
    path: "#/docData/_meta"
    title: undefined
    [[Prototype]]: Object
  length: 1
  [[Prototype]]: Array(0)
 ```
```diff
document:
  docData: 
    Bezeichnung1: "XXX"
    Email: "XXX"
    GUID_Kategorie: "XXX"
    Ort: "XXX"
    Plz: "XXX"
    Strasse: "XXX"
    Telefon1: "XXX"
    fb_tablename: "YYY"
    guid: "ZZZ"
    id: 000001
    lastUpdate: 1722258246680
    _attachments: {}
    _deleted: false
    _id: "YYY-ZZZ"
-    _meta: undefined
    [[Prototype]]: Object
  id: "YYY-ZZZ|0"
  isCheckpoint: "0"
  itemId: "YYY-ZZZ"
  _attachments: {}
  _deleted: false
  _meta: {lwt: 1750777824641.15}
  _rev:"1-rx_storage_replication_3a47a69335c473af1060a3206368bab33f4129307cd626c10dfd317dce027991"
[[Prototype]]: Object
```

PR #7261, opened 2025-07-01
Create migration-bug.test.ts

alpha-2000:
## This PR contains:
Steps:
- Insert document with v1 schema
- Replicate and update via inputstream before migration
- Migrate to v2 schema and replicate again
- Verify correct document state after each step
- schow wath after make a local update to the doc after migration  fixes the problem

this adds a test for the bug described in #7260

### DIFF
--- a/test/unit.test.ts
+++ b/test/unit.test.ts
@@ -19,6 +19,7 @@ import './unit/internal-indexes.test.ts';
  * your relevant tests run first.
  * Do not commit this file if you modified the order.
  */
+import './unit/migration-bug.test.ts';
 import './unit/rx-storage-implementations.test.ts';
 import './unit/rx-storage-query-correctness.test.ts';
 import './unit/rx-storage-helper.test.ts';
--- a/test/unit/migration-bug.test.ts
+++ b/test/unit/migration-bug.test.ts
@@ -117,7 +117,7 @@ describe('migration-bug.test.js', () => {
         const sub2 = collection
             .findOne('a')
             .$.subscribe((doc) => {
-                emmited.push(doc);
+                emitted.push(doc);
             });
 
         await AsyncTestUtil.waitUntil(() => emitted.length === 1);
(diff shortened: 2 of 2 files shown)


pubkey:
Shouldnt this fail to reproduce your bug?

alpha-2000:
Yes, Case 2 fails. That's the bug. Cases 1 and 3 only serve to demonstrate the normal intended behavior and show that nothing else is wrong with the test scenario.

alpha-2000:
After transpiling I started with the following:
npx cross-env DEFAULT_STORAGE=memory npx mocha ./test_tmp/unit/migration-bug.test.js or
npx cross-env DEFAULT_STORAGE=dexie npx mocha ./test_tmp/unit/migration-bug.test.js
and for both storage layers 1 and 3 are successful and 2 fails as intended

pubkey:
Can you make it fail in the CI?

pubkey:
Hi, it seems that your 3 test cases do not run independently. Only if the first one is running, the `should update document via replication stream AFTER migration` fails. Please create a single minimal test which reproduces exactly what is failing.

pubkey:
Ok I could reproduce, working on a fix.

alpha-2000:
Hi Daniel,

I hope you’re doing well!

I wanted to kindly ask if you have an idea when the bugfix might be merged.
I created this pull request with a test case about three days ago (thank you for reviewing!), but the issue itself has been affecting my workflow for about two weeks now.

At the moment, I’m unable to release a new version of my app, since all recent builds are affected by this bug.
I understand everyone is busy, and I really appreciate your hard work! If you could let me know roughly how long it might take until a fix is available, I’d be very grateful—this would help me to plan my next release.

Thank you so much in advance!

pubkey:
> I’d be very grateful—this would help me to plan my next release.

Hi @alpha-2000 you last message was clearly generated by ChatGPT, please do not do that. We can write like normal humans 😃

I am working on this. Thank you for providing the test case. Your test revealed some deeper laying problem and I want to fix that instead of patching some stuff just to make it work. This will likely take some more days.

alpha-2000:
Hi Daniel, nice to hear from you.
Sorry about the chatGPT thing. I was just trying to save time. But you're right, this isn't the right place for that.

pubkey:
fixed here: https://github.com/pubkey/rxdb/pull/7266

Issue #7264, opened 2025-07-14
Replication pause `ensureNotFalsy()` throws


KingSora:
Good day! :)

Since we update rxdb to `v16.15.0` our sentry is getting spammed with `ensureNotFalsy` errors. This is happening because of the `toggleOnDocumentVisible` option introduced in `v16.5.0` and the replication state queue which was introduced in `v16.13.0`.

To be more specific the error is coming from here: https://github.com/pubkey/rxdb/blob/bc50ccfb2a2a5452ddf7f0864e97ca37294ec017/src/plugins/replication/index.ts#L625 and thrown here: https://github.com/pubkey/rxdb/blob/bc50ccfb2a2a5452ddf7f0864e97ca37294ec017/src/plugins/replication/index.ts#L437

We also can't really control this behavior, because the whole code is running because of the `visibilitychange` event.

I'm assuming that this error "isn't really an error" since the replication still runs normaly and `ensureNotFalsy` is only used so that one can be sure that `this.internalReplicationState` is not falsy and the replication can be paused. I would say that this error is not helping or saying anything besides "there is an error, but everything continues to work normally".

If this behavior is expected and `this.internalReplicationState` can be falsy I would suggest to silently ignore it if it happens in the `visibilitychange` event, since otherwise the error is being spammed for each collection which can be a lot.

If the behavior mustn't happen and `this.internalReplicationState` must be defined in the queue, then this is a bug.

A test with the reproduction is here: https://github.com/pubkey/rxdb/pull/7265

PR #7265, opened 2025-07-14
Replication state Ensure not falsy throw when `pause` is called

KingSora:
## This PR contains:
 - SOMETHING ELSE

A test which reproduces the behavior described in: https://github.com/pubkey/rxdb/issues/7264


pubkey:
fixed here: https://github.com/pubkey/rxdb/pull/7267

PR #7269, opened 2025-07-16
UPDATE mingo

pubkey:


pubkey:
Seems to not work because mingo no longer ships cjs

pubkey:
https://github.com/pubkey/rxdb/pull/7403

Issue #7430, opened 2025-10-03
incrementalPatch does not work as expected for documents with nested fields.


alex-zywicki:
Given a document structure like
```json
{
    "foo": "bar",
    "data": {
        "field1":  1,
        "field2": 2,
    }
}
```

a call to `incrementalPatch` like
```typescript

doc.incrementalPatch({
    data: {
        field1: 3
    }
})
```

will result in a new object state like
```json
{
    "foo": "bar",
    "data": {
        "field1":  3,
        // "field2": 2, This field was removed
    }
}
```

After looking at the implementation https://github.com/pubkey/rxdb/blob/master/src/rx-document.ts#L316

It appears that `incrementalPatch` simply does a "one level" shallow patch instead of a "deep merge" operation so the output would be
```json
{
    "foo": "bar",
    "data": {
        "field1":  3, // The field was changed, but other fields not interfered with
        "field2": 2,
    }
}
```

A simple potential solution would be to leverage a package like https://www.npmjs.com/package/object-deep-merge (I cannot personally vouch for this library, it was what I found when searching) or to implement that functionality manually in RxDB.

That would mean updating the current implementation from
```typescript
    incrementalPatch<RxDocumentType = any>(
        this: RxDocument<RxDocumentType>,
        patch: Partial<RxDocumentType>
    ): Promise<RxDocument<RxDocumentType>> {
        return this.incrementalModify((docData) => {
            Object
                .entries(patch)
                .forEach(([k, v]) => {
                    (docData as any)[k] = v;
                });
            return docData;
        });
    },
```

to something like
```typescript
    incrementalPatch<RxDocumentType = any>(
        this: RxDocument<RxDocumentType>,
        patch: Partial<RxDocumentType>
    ): Promise<RxDocument<RxDocumentType>> {
        return this.incrementalModify((docData) => {
            return merge(docData, patch);
        });
    },
```

I would be happy to provide further context to help prove the existence of the problem, and write a PR with the proposed change if this seems acceptable.

alex-zywicki:
I appears that `patch` has the same underlying issue.

alex-zywicki:
[...]

PR #7432, opened 2025-10-03
Fix patch and incremental patch for nested values

alex-zywicki:
## This PR contains:
 - An updated implementation of `patch` and `incrementalPatch`
 - A new third patty dependency https://www.npmjs.com/package/object-deep-merge
 - Supporting tests.
 
 This PR contains 2 commits that;
1.   Recreate the bug
2.  Fix the bug

[patch`and incrementalPatch can splice out nested values](https://github.com/pubkey/rxdb/issues/7430)

## Todos 
- [ ] Tests
- [ ] Changelog

### DIFF
--- a/package.json
+++ b/package.json
@@ -511,6 +511,7 @@
     "mingo": "6.5.6",
     "mongodb": "6.20.0",
     "nats": "2.29.3",
+    "object-deep-merge": "1.0.5",
     "oblivious-set": "1.4.0",
     "reconnecting-websocket": "4.4.0",
     "simple-peer": "9.11.1",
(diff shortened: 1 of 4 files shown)


alex-zywicki:
After looking at the issues with the CI build related to the "object-deep-merge" package I opted to replace that with "@fastify/deepmerge" which I verified locally should build. The tests as a I wrote them still pass with the changed lib.

I also made sure that the merge behavior for arrays overwrites arrays with the "patch" value as I believe this is the behavior that is intended in patch. A lot of these 3rd party libs want to concat the arrays by default.

alex-zywicki:
@pubkey Can you approve the workflow so we can hopefully move forward with this? Or is there something you would need changed?

pubkey:
closed because no answer: https://github.com/pubkey/rxdb/issues/7430

Issue #7371, opened 2025-09-18
Firestore removed events do not get processed(?)


calvinturbo:
Hi,

I am using RxDB with Firestore replication. I also have a premium license.

I am using replication for the Firestore collection 'shipments' with where filters. 
`where('status', 'in',
          [ShipmentStatusEnum.PENDING, ShipmentStatusEnum.RETRIEVED, ShipmentStatusEnum.DELIVERED, ShipmentStatusEnum.ONROUTE, ShipmentStatusEnum.PLANNEDFORROUTE])`

Therefore, only shipments with above statutes get replicated. 
In my application, you can modify the status of one or multiple shipments at the same time. If you change the status to a status not listed above via Firestore directly, the shipments statutes do not change in the application. If you modify the shipments via RxDB directly, they will change in the clients browser, but not on other clients. 

Example: User A changes shipment 1 and 2 to status 'cancelled'. They change in their browser, but the status remains the same in User B's browser, because they don't process the event coming from Firestore.

https://firebase.google.com/docs/firestore/query-data/listen#view_changes_between_snapshots
shows an example how to correctly process these changes

If a document moves out of the replication scope because it no longer matches the filters, a 'removed' event is received by Firestore client.
I have verified this myself with my application.

<img width="530" height="187" alt="Image" src="https://github.com/user-attachments/assets/77f6f2d4-8266-4407-829c-b8c01b1ba38b" />

RxDB should correctly process 'removed' events and .remove() documents that are no longer in the Firestore replication scope.

pubkey:
Can you make a PR with a test case that reproduces your problem?
You can start with [this template](https://github.com/pubkey/rxdb/blob/master/test/unit/bug-report.test.ts).

calvinturbo:
Ping, pretty busy right now, but will write the test soon

calvinturbo:
Created pull request, added test case to firestore-replication.test file
https://github.com/pubkey/rxdb/pull/7445

PR #7445, opened 2025-10-08
add test #7371 documents moving outside replication scope

calvinturbo:
## This PR contains:
- added test to Firestore Replication test file, tests if replication correctly notices docs moving outside of replication scope

Missing test for issue with Firestore docs moving outside of replication scope https://github.com/pubkey/rxdb/issues/7371


pubkey:
I understand the problem.
But I see no way to fix this on the RxDB side atm. If there is no more permission to the doc, the sync will not know about the move-out-of-scope change to that document.
Atm docs are expected to either stay inside or outside of the scope, not change that.

calvinturbo:
> I understand the problem. But I see no way to fix this on the RxDB side atm. If there is no more permission to the doc, the sync will not know about the move-out-of-scope change to that document. Atm docs are expected to either stay inside or outside of the scope, not change that.

This should be fixable, because Firestore sends a 'removed event', RxDB just needs to process that.

https://firebase.google.com/docs/firestore/query-data/listen#view_changes_between_snapshots
shows an example how to correctly process these changes

If a document moves out of the replication scope because it no longer matches the filters, a 'removed' event is received by Firestore client.
I have verified this myself with my application.

Image

calvinturbo:
I'll check if I can make the necessary modifications to the Firestore Replication plugin, but I don't know if I am sufficiently qualified, will keep you posted

pubkey:
Hi @calvinturbo 
This PR fails on the CI, please check.

calvinturbo:
@pubkey I'll start working on it

pubkey:
Any update @calvinturbo ?

Issue #7444, opened 2025-10-08
CouchDB replication pull handlers fails on deleted document when schema validation is enabled.


alex-zywicki:
I have an RxDB setup to live replication from CouchDB, when a document is deleted from couch an error occurs that is reported by the replication states `error$` subject. 

When a document is deleted in CouchDB a simple stub is returned
```json
{
  "_id": "...",
  "_rev": "...",
  "_deleted": true
}
```

this is then forwarded along the replication chain and seems to be hitting the Ajv validator plugin I have enabled and is failing with a validation error saying that I am missing required fields.

This is the error message (I've omitted parts to keep it short and to remove potentially sensitive info)
My schema requires several top level required fields including the `data_type` the error is referring to.
```
Err:  RxError (RC_PULL): 

Error message: RxReplication pull handler threw an error - see .errors for more details
Error code: RC_PULL
Find out more about this error here: https://rxdb.info/errors.html?console=errors#RC_PULL 

--------------------
Parameters:
writeError: {
  "status": 422,
  "isError": true,
  "documentId": "4e87f787-3336-4df9-a603-38aafad81293",
  "writeRow": {
    "document": {
      "_id": "4e87f787-3336-4df9-a603-38aafad81293",
      "_deleted": true,
      "_meta": {
        "lwt": 1759940096305.02
      },
      "_attachments": {},
      "_rev": "2-yumvvanuif"
    },
  ...
  "validationErrors": [
    {
      "instancePath": "",
      "schemaPath": "#/required",
      "keyword": "required",
      "params": {
        "missingProperty": "data_type"
      },
      "message": "must have required property 'data_type'"
    }
  ],
  ...
}
```

As a temporary work around I have modified my pull handler to work around this
```typescript
export function handlePull(docData: any) {
    ...

    if (docData._deleted === true) {
        docData.data_type = ""; // inject defaults into deleted data to appease the validator
        ...
    }
    ...
    
    return docData;
}
```

After reviewing the CouchDB replication plugin implementation I would propose a fix that looks up the deleted doc in the collection and injects the current state into the event stream with the deletion applied to it. 

I believe these would be the places to do so.
https://github.com/pubkey/rxdb/blob/master/src/plugins/replication-couchdb/index.ts#L127
https://github.com/pubkey/rxdb/blob/master/src/plugins/replication-couchdb/index.ts#L322

The solution would look something like

```typescript
const existing = await collection.findOne({
    selector: {
        primaryKey: change._id
    }
});

[...]

PR #7446, opened 2025-10-09
add test case to replcate #7444 (CouchDB sync _deleted=true)

alex-zywicki:
## This PR contains:
 A test to replicate  #7444 
 A fix for #7444 (TODO)

Replicating a deleted doc from CouchDB with schema validation fails when your schema has required fields.
This happens because CouchDB does not send the full document over the change-feed on delete only the `_id`, `_rev`, and `_deleted` fields.
#7444 

## Todos
- [ ] Tests


alex-zywicki:
I have done some experimentation at potential fixes:

```typescript
// src/plugins/replication-couchdb/couchdb-helper.ts
export async function couchDBDocToRxDocData<RxDocType>(
    collection: RxCollection<RxDocType, unknown, unknown, unknown>,
    couchDocData: any
): Promise<WithDeleted<RxDocType>> {
    const primaryKey = collection.schema.primaryPath;
    let doc = couchSwapIdToPrimary(primaryKey as any, couchDocData);
    // ensure deleted flag is set.
    doc._deleted = !!doc._deleted;

    delete doc._rev;
    if (doc._deleted !== true) {
        return doc;
    }

    const found = await collection.findOne(doc[primaryKey]).exec();
    if (found === null) {
        return doc;
    }
    // no need to worry about a deep merge since doc only contains _id, _rev, _deleted in this case
    return Object.assign(doc, found.toMutableJSON())
}
```
I tried this, but it is looking like there is an async issue going on as the collection seems to be empty at the point where I am observing the deletion (found comes back null).

I don't know RxDB's internals well enough to comment on why that is the case, but my guess would be that in the test case I am developing against the document create and delete are happening in such short succession that the state has not reached the storage completely before the delete comes in.

Assuming that is the case I'm not sure this approach would work without either going back to couch for the previous revision (slow and unreliable) or keeping some sort of cache in the replication plugin itself (waste of space, duplicate data).

The only other approach I can think of would be to tweak the validation plugins themselves to consider deleted documents to be valid or to parameterize that as an option. I tried this and it does work, but I suspect doing so by default would have other repercussions on the system. 

@pubkey what are your thoughts?

pubkey:
Hi @alex-zywicki 
Sorry for the late reply.

Tweaking the validation is not an option for me. We should "fill up" the deleted document data so that it matches the normal schema. This fill-up-data can either be retrieved from the local store or by making an additional couchdb-request.
Both of these options are ok for me.

alex-zywicki:
> Hi @alex-zywicki Sorry for the late reply.
> 
> Tweaking the validation is not an option for me. We should "fill up" the deleted document data so that it matches the normal schema. This fill-up-data can either be retrieved from the local store or by making an additional couchdb-request. Both of these options are ok for me.

What would the "local store" be in this case? I assume you are referring to the underlying storage?

Going back to couch for the data is not a truly viable option as couch is not required to keep previous revisions of documents at all times, so things like compaction can wipe out the revision you need.

I tried going to the collection to get the data, but I hit a case where if you do a create followed by a delete in quick succession the initial create will not have reached the storage before you go to process the delete resulting in a failure to find the doc from the collection. The only way I can think to handle that would be for the couch replication to hold it's own cache in memory.

That in memory cache could be a simple js Map object or even could take the form of a "shadow db" using the in memory storage option. The primary issue is that both of there options would functionally double the required storage.

I suppose it would be possible to write a cache such that it evicts items that it knows have already made it to storage, but I'm not sure what the correct strategy for maintaining that over time would be.

pubkey:
The metadata of a replication already contains the assumed-server-state. Maybe we can use that one? Otherwise I see no other way then fetching it from the couchdb server while assumeing it will always be there at that point in time.

pubkey:
@alex-zywicki any update on this?

alex-zywicki:
> @alex-zywicki any update on this?

I haven't had the time to look into it further yet. So no update at this time, I will try to find some time to try out the assumed server state you had mentioned. How is that accessed?

pubkey:
> I haven't had the time to look into it further yet. So no update at this time, I will try to find some time to try out the assumed server state you had mentioned. How is that accessed?

https://github.com/pubkey/rxdb/issues/7444#issuecomment-3743254596

PR #7482, opened 2025-10-23
add test using composite pull filter on firestore

agarcher:
## This PR contains:
 - IMPROVED TESTS

When using the firestore replication, it is possible to use a composite filter if you use _only one_ filter, but doing so causes typescript errors. The problem is that the [pull options filter constrains to field filters only](https://github.com/pubkey/rxdb/blob/3f4fbc04878f8fb97ece73742cfbde795a2036cb/src/plugins/replication-firestore/firestore-types.ts#L42) and according to firestore a [composite filter is a different thing](https://github.com/firebase/firebase-js-sdk/blob/8209266c625c2c08effd8d0d4c82af72439a0f9c/packages/firestore/src/lite-api/query.ts#L366-L368).

What I don't understand is that the code in the test _does work_ if you throw an `as any` on there to tell typescript it's safe, but then we lose type safety. It would be great if the replication could relax the typing to allow a single composite key, but when I attempt to do that I wind up with type errors in the plugin source code where it performs the query. Some attempts to resolve the type issues led me to dead ends.


pubkey:
Hi @agarcher 
PR is welcomed. We can use `any` internally as long as it is tested. Its only important that the outside facing types are correct.

agarcher:
Thanks for the response. I will look at fixing with an `any` workaround then.

pubkey:
@agarcher any update on this?

PR #7497, opened 2025-10-27
add bug test case dexie indexed db with findOne subscription + exec

willhoney7:
## This PR contains:
- a bug test case

Using dexie (with fake-indexedDb on node or in the browser with real indexedDb), if you subscribe to a certain type of query, then delete the value, then execute the query, it will still return a result. (see the test)

Some notes:
- this bug does not replicate in memory storage. It may replicate with other storages, but I did not test.
- I _think_ the findOne part is relevant but not 100% sure.
- It is relevant that the query I show does not have a "selector" field. If you update the query to have a selector, the bug will not occur.
- eventReduce on/off does not matter

Sorry the test is a little goofy. I had a hard time getting the env to work correctly and also I can't get the assertion to work correctly? It may be waiting on some async work I'm not flushing/handling. But it throws an error and clearly shows there's a bug.


pubkey:
Thank you @willhoney7 
I think this was a very important finding and fixing this might have also fixed a wide range of time-based test failures.
https://github.com/pubkey/rxdb/pull/7535

PR #7501, opened 2025-10-27
feat: use shiki for ssr codeblocks

polvallverdu:
## This PR contains:
 - IMPROVED DOCS

I've switch from prismjs to shiki, using rehype to compile the codeblocks at build-time, not on the client.

Only detail is that the dracula theme changes from prismjs to shiki, here are some screenshots:

Currently, with prismjs 
Screenshot 2025-10-27 at 22 24 06

With shiki 
Screenshot 2025-10-27 at 22 23 37

I'm exploring rxdb for a personal sideproject, and I'd like to test some of the premium features to make it more performant, as per the [premium task list](https://github.com/pubkey/rxdb/blob/master/orga/premium-tasks.md).

## Todos
- [ ] Tests
- [ ] Changelog

### DIFF
--- a/docs-src/docusaurus.config.ts
+++ b/docs-src/docusaurus.config.ts
@@ -14,57 +14,13 @@ const config: Config = {
     favicon: '/img/favicon.png',
     // Add multiple sizes + Apple touch icon (+ optional SVG)
     headTags: [
-        {
-            tagName: 'link',
-            attributes: {
-                rel: 'icon',
-                type: 'image/svg+xml',
-                href: '/files/logo/logo.svg',
-            },
-        },
-        {
-            tagName: 'link',
-            attributes: {
-                rel: 'apple-touch-icon',
-                href: '/img/apple-touch-icon.png',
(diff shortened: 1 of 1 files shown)


polvallverdu:
I use prettier on save, and it formatted the docusaurus config file quite a lot. I couldn't find a formatter on the project to use.

pubkey:
Hi @polvallverdu 
This PR adds shiki but does not remove the prismjs from the main javascript bundle which was the main reason to switch from something else. Please check the build main.js file by running `npm run docs:build` it should not contain prismjs afterwards.

Changing the theme and style is no problem. I think it looks ok to me.

polvallverdu:
Yeah, I see... I think docusaurus comes with prismjs prebundled, even some examples that use shiki still pull prismjs. Let me try a few things and see what I can do.

pubkey:
I think it has to be excluded with via webpack config somehow.

polvallverdu:
#7590 seems to have got it, I think we can close this

PR #7530, opened 2025-11-07
Fix error in dev mode when writing object with no prototype

toofishes:
In serializing some arbitrary JSON data sourced from another library, I encountered an object with no prototype, similar to this:

```js
const x = Object.create(null)
x.level = 3
```

This caused `containsDateInstance` to error since it called `hasOwnProperty` directly on the object, rather than via `Object.prototype` as is done elsewhere. Adjust this call site so weird objects like these don't cause problems.

Here is a standalone test with the old and new methods. I didn't quite now how to make a unit test for this, since it would require exporting a function you may not wish to expose.

```ts
function containsDateInstanceOld(obj: any): boolean {
    if (typeof obj !== 'object' || obj === null) {
        return false;
    }
    for (let key in obj) {
        if (obj.hasOwnProperty(key)) {
            if (obj[key] instanceof Date) {
                return true;
            }
            if (typeof obj[key] === 'object' && containsDateInstance(obj[key])) {
                return true;
            }
        }
    }
    return false;
}

function containsDateInstance(obj: any): boolean {
    if (typeof obj !== 'object' || obj === null) {
        return false;
    }
    for (let key in obj) {
        if (Object.prototype.hasOwnProperty.call(obj, key)) {
            if (obj[key] instanceof Date) {
                return true;
            }
            if (typeof obj[key] === 'object' && containsDateInstance(obj[key])) {
                return true;
            }
        }
    }
    return false;
}

console.log("old:", containsDateInstanceOld({ a: new Date() })); // true
console.log("new:", containsDateInstance({ a: new Date() })); // true

const x = Object.create(null)
x.level = 3

try {
    console.log("old (error):", containsDateInstanceOld(x)); // throws error
} catch (e) {
    console.error("old (error):", e.message);
}
console.log("new (no error):", containsDateInstance(x)); // false
```

### DIFF
--- a/test/unit/rx-collection.test.ts
+++ b/test/unit/rx-collection.test.ts
@@ -248,6 +248,27 @@ describe('rx-collection.test.ts', () => {
                     await collections.nestedhuman.insert(schemaObjects.nestedHumanData());
                     db.close();
                 });
+                it('should insert nested human with null prototype subobject', async () => {
+                    const db = await createRxDatabase({
+                        name: randomToken(10),
+                        storage: config.storage.getStorage(),
+                    });
+                    const collections = await db.addCollections({
+                        nestedhuman: {
+                            schema: schemas.nestedHuman
+                        }
+                    });
+                    const doc = schemaObjects.nestedHumanData();
+                    doc.mainSkill = Object.create(null);
+                    doc.mainSkill.name = 'Skill';
+                    doc.mainSkill.level = 5;
+                    try {
+                        await collections.nestedhuman.insert(doc);
+                    } catch (err) {
+                        assert.fail('should not throw error: ' + (err as any).message);
+                    }
+                    db.close();
+                });
                 it('should insert more than once', async () => {
                     const db = await createRxDatabase({
                         name: randomToken(10),


pubkey:
Changes look good but we need at least one unit test for this to be merged.

toofishes:
> Changes look good but we need at least one unit test for this to be merged.

I agree, but I just wasn't sure how. From my PR description:

>> I didn't quite now how to make a unit test for this, since it would require exporting a function you may not wish to expose.

Can you give some guidance on the "how" here? I saw the tests in `test/unit/util.test.ts` as being the most promising place, but I wasn't sure how or if to expose this lower-level internal function to be able to be unit tested.

pubkey:
In the test you should use the outside api to insert ine of these object which didnt work for you. Same as you do in your app.

pubkey:
In the test you should use the outside api to insert ine of these object which didnt work for you. Same as you do in your app.

toofishes:
I've added a test to `rx-collection.test.ts` that fails with the old code, and passes with the new code.

PR #7566, opened 2025-11-27
feat: replication-supabase querybuilder

Sicria:
feat: replication-supabase querybuilder

## This PR contains:
 - A NEW FEATURE

Using `replicateSupabase` I want to be able to filter the results that are pulled so I can limit them to a certain `account_id` for example. I've split my rxdb based on an account_id and I'd like to provide that id along to the query like I would normally via `.eq('account_id', accountId)`. This way I don't need to filter all queries to that account and it also limits how much is fetched by users who belong to multiple accounts (eg not syncing an account until they visit it).


pubkey:
Thank you, merged.

Issue #7587, opened 2025-12-03
Max length primary keys and checkpoints


wuzzeb:
I am getting a max length error when writing the checkpoints to the meta instance, specifically this line in [checkpoint.ts](https://github.com/pubkey/rxdb/blob/98027c85a03fda9e4dce8c25ae2700ae108fa8b1/src/replication-protocol/checkpoint.ts#L128). The error is a validation error that the id exceeds the maximum length.  Stepping through the code, the `id` is set in a call to getComposedPrimaryKeyOfDocumentData a few lines above which is setting the id to `up|1` where up is the direction.

But looking at the definition of the meta instance, it is defined like so:

```
            id: {
                type: 'string',
                minLength: 1,
                // add +1 for the '|' and +1 for the 'isCheckpoint' flag
                maxLength: parentPrimaryKeyLength + 2
            },
```

But, in my own collection, I have a maximum key length of 1 (keys are only a single character),
so this is calculated to 3.  But `up|1` has length 4.

I see that `down|1` is also a key, so the max length for the meta instance should probably be `max(4, parentPrimaryKeyLength) + 2`.

pubkey:
Can you make a PR with a test case that reproduces your problem?
You can start with [this template](https://github.com/pubkey/rxdb/blob/master/test/unit/bug-report.test.ts).

wuzzeb:
I made a unit test.  When single-stepping through the unit test in the debugger, I made a few more observations.

* You can bypass the error by changing the maximum length of the `id` key in the schema to 4.  Then the error about the key `up|1` doesn't happen.

* The thrown error also does not appear if the primary key is named something other than `id`.  But I suspect this is still a problem because of what I saw when single-stepping in a debugger.
 
Try changing the primary key in the unit test  to something else, like primaryId or something.  Then, put a breakpoint at

https://github.com/pubkey/rxdb/blob/3ab3124eed2cf952c58ebb0b26955a3d3879cff2/src/replication-protocol/checkpoint.ts#L118

Notice that the call to metaInstance.bulkWrite returns an error, so `result` is an error.  The error is about `up|1` too long because the maximum key length is 1 + 2 = 3.  But then, `getWrittenDocumentsFromBulkWriteResponse` is called, and if you put a breakpoint at

https://github.com/pubkey/rxdb/blob/3ab3124eed2cf952c58ebb0b26955a3d3879cff2/src/rx-storage-helper.ts#L880

[...]

PR #7589, opened 2025-12-05
Unit test for checkpoints of short primary keys

wuzzeb:
Add a unit test for issue #7587

The primary key in the collection has maximum length 1, which causes the meta-instance to have maximum length 3.  Then when entering a key of `up|1` during replication, get an error.


pubkey:
replaced by https://github.com/pubkey/rxdb/pull/7591

PR #7591, opened 2025-12-08
FIX short primary key lengths does not match replication schema #7587

pubkey:


pubkey:
https://github.com/pubkey/rxdb/issues/7587

PR #7619, opened 2025-12-21
ADD restore ssr in angular example

trik:
## This PR contains:
 - IMPROVED TESTS

Restore server-side-rendering in angular examples and re-enable ssr tests. It was removed when the angular example has been updated to angular 17


trik:
@pubkey have you had the chance to look at my pr? i can see that CI is failing cause the ts-node package is missing

don't know if it can be useful, but i could work also on an update of the angular app. currently it's using v17 which is quite old and it does not use any of the features introduced by the new versions (standalone components, signals)

pubkey:
@trik Yes if you find the time, please update the angular version.
If the ts-node package is missing, just add it to the deps in the angular example.

trik:
@pubkey the angular app updated to v21 is ready

we have only two relevant changes:

1. this assertion has been removed
https://github.com/pubkey/rxdb/blob/5c4ac682ef2c3f8b1b797db75b4ee36c485ff4b8/examples/angular/src/app/components/heroes-list/heroes-list.component.ts#L62
the angular rendered has changed a lot in the latest versions, and it's not true anymore that the observables are run inside ngzone. if you try adding a console log inside the tap operator function and commenting the list template you could see that the observables is not running. if you want to add a similar test we should find another way

2. in the previous version of the application, the angular injector was passed as a parameter to the initDatabase function, i had to remove it because it was throwing "Injector has already been destroyed" exception on SSR
https://github.com/pubkey/rxdb/blob/5c4ac682ef2c3f8b1b797db75b4ee36c485ff4b8/examples/angular/src/app/app.module.ts#L58
the injector parameter is optional for the toSignal method, what was the purpose of manually passing as a parameter?

pubkey:
> what was the purpose of manually passing as a parameter

If I remember correctly, this was to make the RxDatabase starting up even before angular itself is loading, to this improves the time to interaction because when angular starts the database is already there.

trik:
> > what was the purpose of manually passing as a parameter
> 
> If I remember correctly, this was to make the RxDatabase starting up even before angular itself is loading, to this improves the time to interaction because when angular starts the database is already there.

yeah, i get it, the app initializer is still there, i only removed the manual Injector injection

i see that ci is failing, sorry i was focused on ssr and i did not run the other tests, the hero list component template has changed a little bit, that's why they are failing i guess

trik:
@pubkey the test execution should be good now, i tested also the other steps of the CI flow

trik:
@pubkey we still have the ts-node problem.. never got the same error locally, i will to reinstall deps as the CI workflow does and try to see if i can reproduce the problem

pubkey:
@trik I think `ts-node` should be added to the `package.json` of the angular example.

trik:
> @trik I think `ts-node` should be added to the `package.json` of the angular example.

i was able to reproduce the problem locally, apparently ts-node was the only package missing, now i can run the example-angular step using act, i guess it will run also on github

pubkey:
Thank you so much, merged.
Please contact me in discord for the free premium access.

PR #7688, opened 2026-01-15
ADD react plugin

trik:
## This PR contains:
A new React plugin that provides a set of React Hooks for RxDB

The library https://github.com/cvara/rxdb-hooks is outdated and not maintened anymore

## Todos
- [ ] Documentation


trik:
@pubkey the useRxQuery hook is simpler than the one in the original library
there's no option for pagination that can be achieved easily simply altering the mango query
the only other parameter is the live bool

probably we could return an helper method to manually stop the live query, that obviously does nothing with one-shot queries

it could be useful to add a useInitRxDatabase to simplify the db init process, what do you think?

pubkey:
@trik Thank you so much for your work.

I added the `RxDatabase.collections$` directly into the core (slightly renamed): https://github.com/pubkey/rxdb/pull/7705

> there's no option for pagination that can be achieved easily simply altering the mango query

That is ok, people can paginate by running new queries.

Can you add documentation for this somehow. I never used the old rxdb-hooks.

trik:
@pubkey removed the custom plugin and now i'm using the core collections$

added some documentation in code for all the hooks and the provider component

pubkey:
Thank you for your work @trik 
I justs merged this.
Notice that I will add some changes and docs. Will be released in the netx v17 beta.

PR #7695, opened 2026-01-16
Remove unnecessary setEndpointRealtime from Appwrite client examples

adityaoberai:
## This PR contains:
 - IMPROVED DOCS

The `setEndpointRealtime` method is only necessary in situations where a custom proxy is used, and Appwrite's expected default realtime endpoint may not work. Also, in this case, the endpoint used is incorrect.

This PR removes `setEndpointRealtime` from the Appwrite client examples in the Appwrite integrations docs.

### DIFF
--- a/docs-src/docs/replication-appwrite.md
+++ b/docs-src/docs/replication-appwrite.md
@@ -208,7 +208,6 @@ const collection = db.humans;
 ```ts
 const client = new Client();
 client.setEndpoint('https://cloud.appwrite.io/v1');
-client.setEndpointRealtime('https://cloud.appwrite.io/v1');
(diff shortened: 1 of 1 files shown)


pubkey:
Hi @adityaoberai 
This was required because when the normal endpoint is set to `localhost` on selfhosting (not using appwrite cloud), the realtime endpoint by default still pointed to the cloud. Did you change the default on the appwrite side?

adityaoberai:
@pubkey in the current latest version of the Web SDK, if the realtime endpoint is not set, it is formed based on the client endpoint.

https://github.com/appwrite/sdk-for-web/blob/d915592446e30654834ebcba4b87b360b5d4a3a9/src/services/realtime.ts#L132

So you wouldn't have a different default.

PR #7709, opened 2026-01-20
Add test to demonstrate that two pipelines hang

LordZardeck:
## This PR contains:
Test case to demonstrate hang and failure with two pipelines running

### DIFF
--- a/test/unit/bug-report.test.ts
+++ b/test/unit/bug-report.test.ts
@@ -9,18 +9,35 @@
  * - 'npm run test:browser' so it runs in the browser
  */
 import assert from 'assert';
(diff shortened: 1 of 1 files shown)


pubkey:
Hi @LordZardeck 
I think this behavior of RxDB is correct. By definition when a pipeline is running, all outside reads/writes to the collection are halted until the pipeline is done. Therefore if your second pipeline waits for the first to be idle and at the same time the first pipeline queries the second pipelines collection, they will block each other.

EDIT: I just added this info to the docs: https://github.com/pubkey/rxdb/commit/567d971d244ac01294740a39461170987e59a4b0

PR #7721, opened 2026-01-22
REMOVE outdated isomorphic-ws dependency

trik:
Since newer Node.js versions have a built-in WebSocket client (stable since v22.4.0), there's no need to rely on unmaintened isomorphic-ws library

### DIFF
--- a/package.json
+++ b/package.json
@@ -548,7 +548,6 @@
     "graphql": "16.12.0",
     "graphql-ws": "5.16.2",
     "is-my-json-valid": "2.20.6",
-    "isomorphic-ws": "5.0.0",
     "js-base64": "3.7.8",
     "jsonschema-key-compression": "1.7.0",
     "mingo": "7.1.1",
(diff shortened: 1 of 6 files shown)


trik:
https://nodejs.org/en/learn/getting-started/websocket

pubkey:
Thanks, merged.

trik:
@pubkey thanks for merging
i can see that one of the tests was failing on firefox cause of the `ensureIsWebsocket` check, which you removed and i already checked it was quite useless.. there was a reference to a similar check made by the `reconnecting-websocket` lib but i didn't find such a test in the library code

i couldn't run the tests locally mainly because  i'm working on ubuntu 25.10 which provides Firefox as Snap package and the `karma-firefox-launcher` plugin can't run the browser when running as a snap
https://github.com/karma-runner/karma-firefox-launcher/issues/183

All the karma ecosystem has been deprecated 3 year ago
https://github.com/karma-runner/karma?tab=readme-ov-file#karma-is-deprecated-and-is-not-accepting-new-features-or-general-bug-fixes
have you considered to migrate to another test runner? maybe Jest or Vitest

pubkey:
@trik I considered switching but it was never worth the time. We have so many different karma configs and sideeffects like hosting static files and starting up test-servers. So this would be a lot of work.

PR #7766, opened 2026-01-31
FIX supabase push handler throwing for objects

csenio:
## This PR contains:
 - A BUGFIX

supabase replication doesn't handle jsonb or array types. More specifically, the push handler throws

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Typings
- [ ] Changelog

### DIFF
--- a/src/plugins/replication-supabase/helper.ts
+++ b/src/plugins/replication-supabase/helper.ts
@@ -31,8 +31,12 @@ export function addDocEqualityToQuery<RxDocType>(
         const v = (doc as any)[key];
         const type = typeof v;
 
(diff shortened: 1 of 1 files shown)


pubkey:
Thank you for the contribution. Notice that each fix need at least one test case.

csenio:
Sry for the delay, wanted to test this in the wild first. The JSON / array equality check I attempted fails when those columns get too long since they're all url encoded. 

But this leads me to wonder why we even check the equality for all fields instead of just `id`+`updated_at`. Shouldn't those be enough to check the doc equality? 
 @pubkey

PR #7787, opened 2026-02-04
Fix memory leak in migratePromise() (#7786)

### LINKED ISSUE #7786: Memory leak: migratePromise() doesn't clean up internal replication state
## Summary

Calling `collection.migratePromise()` causes a memory leak because the internal replication state is never properly cleaned up after migration completes.

## Root Cause

In `rx-migration-state.ts`, the `migrateStorage()` method creates a `replicationState` via `replicateRxStorageInstance()`, but this local variable is **never assigned to `this.replicationState`**:

```typescript
const replicationState = replicateRxStorageInstance({...});  // Line 375
// ... migration happens ...
await this.cancel();  // Line 474 - but this.replicationState is undefined!
```

When `cancel()` runs, it checks `if (this.replicationState)` which is always `undefined`, so `cancelRxStorageReplication()` is never called.

## Consequence

The subscriptions created in `upstream.ts` are never unsubscribed:
- `forkInstance.changeStream().subscribe(...)` keeps a reference to the old storage
- `masterChangeStream$.subscribe(...)` also remains active

These subscriptions are only cleaned up when `state.events.canceled` emits `true`, which only happens when `cancelRxStorageReplication()` is called.

## Heap Snapshot Evidence

The retainer path shows:
```
sessionId → docData → previousDocumentData → events → task → Array → openTasks → Context → oldStorage → RxStorageInstanceDexie → changes$ → Subject
```

This confirms the `openTasks` array in `upstream.ts` and the `changes$` Subject are being retained.

## Environment

- RxDB version: 16.21.1+
- Storage: Dexie (IndexedDB)
- Plugins: RxDBMigrationSchemaPlugin, RxDBLeaderElectionPlugin

## Fix

Add one line after creating the replication state in `migrateStorage()`:

```typescript
const replicationState = replicateRxStorageInstance({...});
this.replicationState = replicationState;  // ADD THIS LINE
```

This allows `cancel()` to properly call `cancelRxStorageReplication()` which emits the cancel event and triggers subscription cleanup.

OskarD:
## Summary

Fixes #7786

The `migratePromise()` function caused a memory leak because the internal replication state was never assigned to `this.replicationState`, preventing `cancel()` from properly cleaning up subscriptions after migration.

## Changes

- Assign `replicationState` to `this.replicationState` in `migrateStorage()` so that `cancel()` can call `cancelRxStorageReplication()`
- Add test to verify the replication state is properly assigned and canceled after migration

## Root Cause

In `migrateStorage()`, the local variable `replicationState` was created but never assigned to `this.replicationState`. When `cancel()` ran at the end of migration, it checked `if (this.replicationState)` which was always `undefined`, so `cancelRxStorageReplication()` was never called.

This left the subscriptions in `upstream.ts` active, keeping references to the old storage instance.

## Test Plan

- [x] Added test that verifies `replicationState` is assigned during migration
- [x] Added test that verifies `replicationState.events.canceled` is `true` after migration
- [x] All 36 migration tests pass
- [x] Lint passes
- [x] TypeScript type check passes


pubkey:
@OskarD please rebase.

pubkey:
Thank you @OskarD for the PR, this helped a lot.
I made some changes and added a fix here https://github.com/pubkey/rxdb/pull/7887

Issue #7791, opened 2026-02-05
Migration memory leak: old collection meta doc not deleted + migratePromise() fire-and-forget causes orphaned subscriptions


OskarD:
## Summary

Two bugs in the migration-schema plugin combine to create a severe memory leak (~100 MB/min) that worsens on every app restart. Related to but distinct from #7786.

**RxDB version:** 16.21.1
**Environment:** Chromium-based desktop app (Overwolf), `autoMigrate: false`, calling `migratePromise()` manually

## Bug 1: Old collection meta doc not deleted after migration (`_rev` conflict)

After `startMigration()` successfully migrates all documents, it attempts to delete the old collection meta doc (e.g., `unified_hero_sessions-1`) via `writeSingle()`:

```js
// rx-migration-state.js, startMigration()
await writeSingle(this.database.internalStore, {
  previous: oldCollectionMeta,
  document: Object.assign({}, oldCollectionMeta, { _deleted: true })
}, 'rx-migration-remove-collection-meta');
```

This fails with a `_rev` conflict because `oldCollectionMeta` was fetched at the start of `startMigration()` and the internal store document has been modified since then (by `updateStatus()` calls or other internal writes). The conflict is caught and silently ignored when `isConflict.documentInDb._deleted` is truthy, but in this case the doc is **not** deleted — it just has a stale `_rev`.

**Result:** The old collection meta doc persists forever. On every subsequent restart, `migrationNeeded()` (which calls `mustMigrate()` → `getOldCollectionMeta()`) finds this doc and returns `true`, even though migration already completed.

## Bug 2: `migratePromise()` calls `startMigration()` without `await`

```js
// rx-migration-state.js
_proto.migratePromise = async function migratePromise(batchSize) {
    this.startMigration(batchSize);  // <-- fire-and-forget, no await
    var must = await this.mustMigrate;
    if (!must) { return { status: 'DONE', ... }; }
    var result = await Promise.race([
        firstValueFrom(this.$.pipe(filter(d => d.status === 'DONE'))),
        firstValueFrom(this.$.pipe(filter(d => d.status === 'ERROR')))
    ]);
    // ...
};
```

`startMigration()` is called without `await`. Then `migratePromise()` enters a `Promise.race` on `this.$` (an `observeSingle` + `shareReplay` pipeline on the internal store) waiting for DONE or ERROR status.

When combined with Bug 1, the following happens on every restart after the first migration:

[...]

PR #7792, opened 2026-02-05
Fix migration memory leak: old collection meta doc not deleted (#7791)

OskarD:
## This PR contains:

- A BUGFIX

Two bugs in the migration-schema plugin combine to create a memory leak (~100 MB/min) that worsens on every app restart (issue #7791).

**Bug 1:** After `startMigration()` migrates all documents, it tries to delete the old collection meta doc using a reference cached at the start of the function. If the doc's `_rev` has changed (due to concurrent writes or storage internals), the `writeSingle()` fails with a conflict. The conflict handler only ignores already-deleted docs, so the meta doc persists. On every restart, `mustMigrate()` finds this orphaned doc and returns `true`.

**Bug 2:** `migratePromise()` calls `startMigration()` without `await` (fire-and-forget, by design for multi-tab leader election). When combined with Bug 1, a stale DONE status doc from a previous migration run satisfies the `Promise.race` immediately, while the orphaned `startMigration()` creates leaked storage instances and subscriptions. The fire-and-forget also means any errors thrown from `startMigration()` become unhandled promise rejections.

## Changes

1. Replace single-attempt meta doc deletion with a conflict-retry loop that re-fetches the doc before each attempt (same pattern used by `updateStatus()` and `addConnectedStorageToCollection()`).
2. Add `.catch(() => {})` to the fire-and-forget `startMigration()` call to prevent unhandled promise rejections. Errors still propagate via the status observable.

## Todos
- [x] Tests
- [x] Typings

### DIFF
--- a/src/plugins/migration-schema/rx-migration-state.ts
+++ b/src/plugins/migration-schema/rx-migration-state.ts
@@ -231,27 +231,43 @@ export class RxMigrationState {
             return;
         }
 
-        // remove old collection meta doc
-        try {
-            await writeSingle(
-                this.database.internalStore,
-                {
-                    previous: oldCollectionMeta,
-                    document: Object.assign(
-                        {},
-                        oldCollectionMeta,
-                        {
-                            _deleted: true
-                        }
-                    )
-                },
-                'rx-migration-remove-collection-meta'
-            );
-        } catch (error) {
-            const isConflict = isBulkWriteConflictError<InternalStoreCollectionDocType>(error);
-            if (isConflict && !!isConflict.documentInDb._deleted) {
-            } else {
-                throw error;
+        /**
+         * Remove old collection meta doc with retry on conflict.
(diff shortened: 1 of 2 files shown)


OskarD:
**Note on .catch(() => {})** (line 585)

The original code called this.startMigration(batchSize) as fire-and-forget without await or .catch(). Errors before the inner try/catch (e.g. createStorageInstance(), leader election) were already unhandled promise rejections. The .catch(() => {}) makes this explicit and prevents runtime crashes in strict environments.

A better fix would wrap the entire startMigration() body in the existing try/catch so early failures also set status = ERROR and surface through the status observable. That way migratePromise() would resolve with ERROR instead of hanging. But that is a separate improvement beyond the scope of this PR.

pubkey:
Thank you so much. Great work. Merged.

PR #7804, opened 2026-02-09
Tests to illustrate downstream replication issues

mtctn:
## This PR contains:
 - specific tests to demonstrate a replication issue

Some runtime patterns could lead to situations where replication downstream discard changes received because local metaInstance and forkInstance are not strictly equal.
The current implementation assumes that local conflicts will be managed by upstream, which is not always the case in practice, leading to some documents being stuck in their current state, and no more updated by downstream.

This class of issues could perhaps be fixed by adding a step in downstream when a conflict is detected: somehow force the conflicting document to be pushed in upstream queue to ensure it will be resolved, instead of silently discarding the new value and assuming that upstream will deal with it.

Further described in the [following discord channel](https://discord.com/channels/969553741705539624/1237000453791678487/threads/1431254845565702266).


pubkey:
Thank you for providing the test case, this helps a lot.

> This class of issues could perhaps be fixed by adding a step in downstream when a conflict is detected: somehow force the conflicting document to be pushed in upstream queue to ensure it will be resolved, instead of silently discarding the new value and assuming that upstream will deal with it.

The problem with this is that on pull-only replications it would still skip the document.

mtctn:
Makes sense! And could we resolve the conflict directly in downstream instead?

pubkey:
> Makes sense! And could we resolve the conflict directly in downstream instead?

Great idea! Created a PR https://github.com/pubkey/rxdb/pull/7875

pubkey:
Fixed here, please check: https://github.com/pubkey/rxdb/pull/7875

PR #7828, opened 2026-02-12
CHANGE supabase tests to use .ts

pubkey:


pubkey:
@copilot please analyze and propose a fix or improvement for this issue/PR.

<details>

## Important Reference (RxDB docs for LLMs)

Use the official RxDB docs here to align with architecture, plugins, storage abstractions, replication patterns, typings, and coding conventions:
https://rxdb.info/llms.txt

---

## What to do (pick the right action based on what’s present)

### Bug report with **no test case**
- Create a **minimal failing test** that reproduces the bug.
- Ensure the test fails on current `main` (or before your changes) and passes after the fix.
- Keep the test small and focused (one behavior).

### Bug report with a **test case already present** but **no fix**
- Implement the **minimal fix** to make the existing test pass.
- Avoid unrelated refactors.
- Add additional assertions only if they clarify the intended behavior.

### **Regression** (something used to work)
- Add/confirm a regression test that captures the previous correct behavior.
- Fix with the smallest change that restores compatibility.

### **Dependency version bump required**
- Update the dependency versions as requested.
- Adjust code/types/tests to satisfy the new versions.
- If the bump requires breaking changes, document the reason and propose the smallest migration.

### **Types/TS errors**
- Fix types in a way that preserves runtime behavior.
- Prefer stricter/correct typings over `any` or broad casts.
- Add a type-level or runtime test if the repo pattern supports it.

### **Docs-only or README mismatch**
- Update docs/examples to match actual behavior.
- If docs reveal a bug, add a test and fix accordingly.

---

## Output requirements

Please provide:
- A short root-cause explanation
- A unified diff patch (code + tests when applicable)
- Notes on edge-cases / compatibility
- If relevant: any migration notes

---

## 🚨 Author mention rule

If your solution:
- modifies the original author’s implementation, OR
- adds additional logic on top of the author's work, OR
- adds/changes tests related to the author's contribution,

then mention the original issue/PR author at the end of your response:

@<author-username> please review the proposed changes.

</details>

PR #7838, opened 2026-02-14
fix graphql example

nickchomey:
The graphql example had lots of errors and wouldnt work. This seems to fix it.


pubkey:
Thank you  @nickchomey  merged.

PR #7875, opened 2026-02-23
FIX #7804

pubkey:
https://github.com/pubkey/rxdb/pull/7804

### DIFF
--- a/src/types/util.d.ts
+++ b/src/types/util.d.ts
@@ -10,7 +10,7 @@ export type PlainJsonValue =
     PlainSimpleJsonObject |
     PlainSimpleJsonObject[] |
     PlainJsonValue[] |
-    { [key: string]: PlainJsonValue }
+    { [key: string]: PlainJsonValue; }
     ;
 export type PlainSimpleJsonObject = {
     [k: string]: PlainJsonValue | PlainJsonValue[];


mtctn:
Thank you very much!

I'm curious to understand how the fix works.

As far a I understand, you mark forkStateFullDoc._meta with o._rev, wich is current forkStateFullDoc._rev increased by one.
bulkWrite will create a new revision to forkStateFullDoc, so it will end up being saved with condition below being true.

 ```
forkStateFullDoc._meta.o &&
(
    forkStateFullDoc._meta.o.hash === identifierHash &&
    forkStateFullDoc._meta.o._rev === getHeightOfRevision(forkStateFullDoc._rev)
)
```

This condition basically means to me "current document comes from downstream, and has not been updated since last downstream write".
So if true, then propagate downstream updates whatever the discrepancies we encountered, which is what you added (do not enter the `return PROMISE_RESOLVE_VOID` case).
But if false, somehow revision height changed since last downstream write, and do not propagate any other downstream changes until upstream resolves the sync.

Am I correct?

In this case, is there a risk to have some documents locally updated (so _rev increased), but for some reason upstream fails to get it in sync (crash during upstream process, or replication cancelled and started again in downstream only mode, or any other reason), leading to documents being in stale state again?

pubkey:
> This condition basically means to me "current document comes from downstream, and has not been updated since last downstream write".

Yes exactly.

> Am I correct?

Yes

> In this case, is there a risk to have some documents locally updated (so _rev increased), but for some reason upstream fails to get it in sync (crash during upstream process, or replication cancelled and started again in downstream only mode, or any other reason), leading to documents being in stale state again?

I do not think so. Upstream fails will be retried on the next push cycle. If you find any way how this coulds fail, please try to create a test case.

PR #7906, opened 2026-02-26
Misc edits

nickchomey:
Just some miscellaneous edits that I saw while working, as well as addressing a few TODOs (some date back to my work [a year ago to fix FoundationDB](https://github.com/pubkey/rxdb/pull/6876)).

### DIFF
--- a/src/custom-index.ts
+++ b/src/custom-index.ts
@@ -16,7 +16,7 @@ import type {
     JsonSchema,
     RxDocumentData,
     RxJsonSchema
-} from './types/index.ts';
+} from './types/index.d.ts';
 import {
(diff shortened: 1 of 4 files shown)


nickchomey:
@pubkey this PR introduced a bug when running `npm run build` 

`src/index.ts` changed from `export type * from './types/index.ts';`    to `export type * from './types/index.d.ts';`

I think i made this change because it was causing typescript stuff to not resolve properly in the IDE.

But then `scripts/fix-types.mjs` causes `dist/types/index.d.ts` to become `export type * from './types/index.d.d.ts';`

I'm not sure what the appropriate path forward should be here.

PR #7907, opened 2026-02-27
Attachments overhaul

nickchomey:
Note: I opened a previous PR, but something broke with my git and the PR got closed. This is a new one.

This PR changes RxDB's attachments to be handled and saved as Blobs, rather than Base64.

It also adds support for including multiple attachments in the insert and upsert functions, and also adds a putAttachments function for adding multiple attachments in one operation.

Attachments will be compressed if compression is enabled, but they will only be selectively compressed baased on the mime type provided when the attachment was created.

Encryption is also done on the binary data rather than base64.

The RxStorage implementations were modified as-needed to support binary data (eg Foundationdb now handles binary).

Localstorage and websocket remain base64. Websocket could, in theory, be changed to use binary data, but that seems like something that would be a larger breaking change for users of that plugin.

Various functions have been tidied up, removed, etc accordingly.

Tests were created for all functionality, and pass on the fast, node, bun and deno variations. MongoDB was not tested, as I couldn't figure out how to run it. Some bypasses were added for test:deno:dexie as https://github.com/denoland/deno/issues/12067#issuecomment-1975001079. This should be irrelevant, since dexie would never realistically be used on the server - Node and Bun work for testing.

This satisfies [this Premium Task](https://github.com/pubkey/rxdb/blob/966377775cb5dd38c28a59f07639507496ad46cc/orga/premium-tasks.md?plain=1#L19), and these backlog issues:
* https://github.com/pubkey/rxdb/blob/master/orga/BACKLOG.md#add-plugin-for-attachments-compression
* https://github.com/pubkey/rxdb/blob/master/orga/BACKLOG.md#upsert-atomic-with-attachments

Please let me know if you have any questions, concerns, etc...!


src/rx-collection-helper.ts
```diff
@@ -49,6 +50,11 @@ export function fillObjectDataBeforeInsert<RxDocType>(
...
         data._deleted = false;
     }
+    // Support 'attachments' as a user-facing alias for '_attachments'
+    if (Object.prototype.hasOwnProperty.call(data, 'attachments') && !Object.prototype.hasOwnProperty.call(data, '_attachments')) {
+        data._attachments = data.attachments;
+        delete data.attachments;
+    }
     if (!Object.prototype.hasOwnProperty.call(data, '_attachments')) {
         data._attachments = {};
     }
```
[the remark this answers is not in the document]
nickchomey: fixed

docs-src/docs/rx-attachment.md
```diff
@@ -186,14 +207,64 @@ const data = await attachment.getStringData(); // 'meow'
...
+
+Instead of inserting a document first and then calling `putAttachment()` separately, you can include attachments directly in the document data when using `insert()`, `bulkInsert()`, `upsert()`, `bulkUpsert()`, or `incrementalUpsert()`. Provide `attachments` as an array of `{ id, type, data }` objects — the same format used by `putAttachment()`.
+
+```javascript
+import { createBlob } from 'rxdb';
+
+// insert with inline attachments
+const doc = await myCollection.insert({
+    name: 'foo',
+    attachments: [
```
pubkey: This should not be `attachments` but `_attachments`. We currently have not reservation of the field `attachments` so people can have this field as a normal property in their documents already.

docs-src/docs/rx-attachment.md
```diff
@@ -218,15 +289,26 @@ const mySchema = {
...
+            // ... add your own patterns
+        ]
     }
 };
 
 /* ... create your collections as usual and store attachments in them. */
 
 ```
+
+The default compressible types include `text/*`, `application/json`, `application/xml`, `application/javascript`, `image/svg+xml`, `image/bmp`, `application/pdf`, font formats, and other text-based MIME types. Binary formats like `image/jpeg`, `image/png`, `video/*`, and `audio/*` are **not** in the default list and will be stored without re-compression.
```
pubkey: Because we use mdx, at this point in the docs we can js-import the actual default list and render it here. By doing so it would always be up to date.

src/plugins/storage-localstorage/rx-storage-instance-localstorage.ts
```diff
@@ -309,6 +319,7 @@ export class RxStorageInstanceLocalstorage<RxDocType> implements RxStorageInstan
                 '||' + attachment.attachmentId
             );
         });
+        await Promise.all(attachmentWrites);
```
pubkey: this is dangerous. Localstorage has no transactions and is non-async. So awaiting the attachment-writes after the document write is done can lead to other writes happen in between.

Instead we should run async work **before** `categorizeBulkWriteRows` is called to that after the conflict check, we can process the reads and writes non-async.

src/plugins/storage-remote-websocket/index.ts
```diff
@@ -95,11 +201,19 @@ export function getRxStorageRemoteWebsocket(
         async messageChannelCreator() {
             const messages$ = new Subject<MessageFromRemote>();
             const websocketClient = await createWebSocketClient(options as any);
-            websocketClient.message$.subscribe(msg => messages$.next(msg));
+            websocketClient.message$.subscribe(msg => {
+                deserializeBlobsFromWs(msg).then(deserialized => messages$.next(deserialized)).catch(err => {
+                    console.error('RxDB WebSocket client: failed to deserialize incoming message', err);
+                    messages$.next(buildErrorResponse(msg, err));
+                });
```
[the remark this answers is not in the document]
pubkey: Not needed. The receiving storage is supposed to do its own conflict check.

src/types/util.d.ts
```diff
@@ -75,8 +75,11 @@ export type ById<T> = {
 
 /**
  * Must be async to support async hashing like from the WebCrypto API.
+ * Accepts string for document revision hashing,
+ * ArrayBuffer for raw binary hashing,
+ * or Blob for attachment digest hashing.
  */
-export type HashFunction = (input: string) => Promise<string>;
+export type HashFunction = (input: string | ArrayBuffer | Blob) => Promise<string>;
```
pubkey, line 82: update this at the docs at `rx-database.md#hashfunction` and add a blob-to-string-conversion to the example `myOwnHashFunction`

src/rx-collection-helper.ts
```diff
@@ -49,6 +50,11 @@ export function fillObjectDataBeforeInsert<RxDocType>(
     if (!Object.prototype.hasOwnProperty.call(data, '_deleted')) {
         data._deleted = false;
     }
+    // Support 'attachments' as a user-facing alias for '_attachments'
+    if (Object.prototype.hasOwnProperty.call(data, 'attachments') && !Object.prototype.hasOwnProperty.call(data, '_attachments')) {
```
pubkey: Do not reserve `attachments`, just use `_attachments`

src/rx-collection.ts
```diff
@@ -447,6 +455,17 @@ export class RxCollectionBase<
             });
         }
 
+        // Normalize any inline attachment inputs (compute digest/length from Blob)
+        // Also converts array format to internal map format
+        await Promise.all(
```
pubkey: This runs in a hot path so microseconds performance here is important and await is expensive:
- Skip it if attachments is not enabled in the schema.
- Skip per document if attachments array is empty.

src/rx-collection.ts
```diff
@@ -608,9 +630,21 @@ export class RxCollectionBase<
...
                 });
             }
-            useJsonByDocId.set(primary, useJson);
-            insertData.push(useJson);
-        });
+            preparedDocs.push(useJson);
+        }
+
+        // Second pass: normalize inline attachments concurrently across all documents
+        await Promise.all(
```
pubkey: This runs in a hot path so microseconds performance here is important and await is expensive:
- Skip it if attachments is not enabled in the schema.
- Skip per document if attachments array is empty.

test/unit/attachments-compression.test.ts
```diff
@@ -42,6 +41,14 @@ modes.forEach(mode => {
             return;
         }
 
+        // Deno's structuredClone() silently destroys Blob data, returning {}. https://github.com/denoland/deno/issues/12067#issuecomment-1975001079
+        // fake-indexeddb (used by dexie in non-browser envs) relies on
+        // structuredClone, so Blob attachment roundtrips are broken in Deno+dexie.
+        // These tests pass fine on Node and Bun, which is sufficient coverage.
+        if (isDeno && config.storage.name === 'dexie') {
```
pubkey, line 49: - Shouldnt we use base64 then for deno to store attachments?
- This even disables the dexie in-browser tests which shouldnt be done.
- Do we even use structuredClone?

test/unit/attachments.test.ts
```diff
@@ -47,6 +47,15 @@ describeParallel('attachments.test.ts', () => {
     if (!config.storage.hasAttachments) {
         return;
     }
+
+    // Deno's structuredClone() silently destroys Blob data, returning {}. https://github.com/denoland/deno/issues/12067#issuecomment-1975001079
+    // fake-indexeddb (used by dexie in non-browser envs) relies on
+    // structuredClone, so Blob attachment roundtrips are broken in Deno+dexie.
+    // These tests pass fine on Node and Bun, which is sufficient coverage.
+    if (isDeno && config.storage.name === 'dexie') {
```
pubkey, line 59: - Shouldnt we use base64 then for deno to store attachments?
- This even disables the dexie in-browser tests which shouldnt be done.
- Do we even use structuredClone?

test/unit/attachments.test.ts
```diff
@@ -889,6 +898,364 @@ describeParallel('attachments.test.ts', () => {
             db.close();
         });
     });
+    describe('encryption + compression combined', () => {
+        if (!config.storage.hasAttachments) {
+            return;
+        }
+        it('should roundtrip compressible attachment through encryption and compression', async () => {
+            const {
```
pubkey: why do we need a dynamic import here?

test/unit/migration-schema.test.ts
```diff
@@ -1583,6 +1584,15 @@ describe('migration-schema.test.ts', function () {
             if (!config.storage.hasAttachments) {
                 return;
             }
+
+            // Deno's structuredClone() silently destroys Blob data, returning {}. https://github.com/denoland/deno/issues/12067#issuecomment-1975001079
+            // fake-indexeddb (used by dexie in non-browser envs) relies on
+            // structuredClone, so Blob attachment roundtrips are broken in Deno+dexie.
+            // These tests pass fine on Node and Bun, which is sufficient coverage.
+            if (isDeno && config.storage.name === 'dexie') {
```
pubkey, line 1592: - Shouldnt we use base64 then for deno to store attachments?
- This even disables the dexie in-browser tests which shouldnt be done.
- Do we even use structuredClone?

test/unit/attachments-compression.test.ts
nickchomey: In order to write base64 when using deno, that would have to be optionally implemented in the dexie rxstorage implementation. I chose to skip it because I figured that no one actually uses deno + dexie via fake-indexeddb. Is that assumption wrong? 

So, the intention here is to disable the test when it is deno and dexie - how does this disable for dexie in browser (or node or bun, for that matter)? I'll run the tests later to confirm manually though. 

I'll have to look on my computer later, but I think structuredclone is used internally when writing to fake-indexddb.

docs-src/docs/rx-attachment.md
nickchomey: Thanks, I'll look into this. I originally used _attachments then switched to attachments to match something I saw somewhere that I thought was exposed to the users already. It thought it was to match putAttachment, but you don't even use any such field there. 

I will have to review and will surely revert to _attachments and simplify the internals.

test/unit/attachments-compression.test.ts
pubkey: Oh sorry. Overlooked that that it is not a OR `||` but a AND `&&`. You can keep this skip then.

src/rx-collection.ts
nickchomey: added the if checks.

Is there a better non-promise way to do it?

src/rx-collection.ts
```diff
@@ -608,9 +630,21 @@ export class RxCollectionBase<
...
+            preparedDocs.map(async (useJson) => {
+                (useJson as any)._attachments = await normalizeInlineAttachments(
+                    this.database.hashFunction,
+                    (useJson as any)._attachments
+                );
+                const primary: string = (useJson as any)[this.schema.primaryPath];
+                useJsonByDocId.set(primary, useJson);
+                insertData.push(useJson);
+            })
+        );
```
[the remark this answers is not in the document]
nickchomey: added the if checks.

Is there a better non-promise way to do it?

src/rx-collection.ts
nickchomey: added the if checks.

Is there a better non-promise way to do it?

docs-src/docs/rx-attachment.md
nickchomey: I tried to do something for this, but couldnt figure it out - even with AI! Is it something you might be able to do?

nickchomey:
Looks like all the tests are passing now. Strange - there were random NATS and browser:remote errors earlier. Probably just CI issues... 

what do you want me to do about the docs? The mime type injection works locally with `npm run docs:build` and `npm run docs:serve`. Do i need to provide built versions in the PR or do they get rebuilt in CI?

image

pubkey:
Thank you so much @nickchomey Merged!

nickchomey:
Fantastic! Im looking forward to making use of this, as my users have low-end devices with little storage and ram, slow cpu, slow and expensive bandwidth.

PR #8075, opened 2026-03-16
Block unprotected write paths on CRDT-enabled documents

dearlordylord:
The CRDT plugin blocks `incrementalModify` and redirects `incrementalPatch`/`remove` through `updateCRDT`, but `modify()`, `patch()`, `incrementalRemove()`, and `update()` bypass the CRDT operation log by calling `_saveData()` directly. This causes document data and the CRDT operation log to diverge silently. On the next replication conflict, `rebuildFromCRDT()` reconstructs from operations only, discarding non-CRDT changes.

- [**`src/plugins/crdt/index.ts`**](https://github.com/dearlordylord/rxdb/blob/10c77d6c7/src/plugins/crdt/index.ts): Override `modify` (throws CRDT4, arbitrary functions cannot become CRDT ops), `patch`, `incrementalRemove`, and `update` (redirect through `updateCRDT`) on the RxDocument prototype when CRDT is active
- [**`src/plugins/dev-mode/error-messages.ts`**](https://github.com/dearlordylord/rxdb/blob/10c77d6c7/src/plugins/dev-mode/error-messages.ts#L1166-L1171): Add CRDT4 error code
- [**`test/unit/crdt.test.ts`**](https://github.com/dearlordylord/rxdb/blob/10c77d6c7/test/unit/crdt.test.ts#L260-L327): Add tests for all new overrides; load `RxDBUpdatePlugin` so CRDT tests work in isolation

tests are red/green checked

### DIFF
--- a/src/plugins/crdt/index.ts
+++ b/src/plugins/crdt/index.ts
@@ -352,6 +352,43 @@ export const RxDBcrdtPlugin: RxPlugin = {
                 });
             };
 
+            const oldincrementalRemove = proto.incrementalRemove;
+            proto.incrementalRemove = function (this: RxDocument) {
+                if (!this.collection.schema.jsonSchema.crdt) {
+                    return oldincrementalRemove.call(this);
+                }
+                return this.updateCRDT({
+                    ifMatch: {
+                        $set: {
+                            _deleted: true
+                        }
+                    }
+                });
+            };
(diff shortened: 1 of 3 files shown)


pubkey:
Thanks @dearlordylord , merged

PR #8275, opened 2026-04-04
Fix query-builder operators dropping implicit $eq from shorthand selectors

pubkey:
## This PR contains:
- A BUGFIX
- IMPROVED TESTS

When using the query-builder API with a selector shorthand value (e.g., `{ age: 5 }` instead of `{ age: { $eq: 5 } }`), chaining another operator on the same field would silently drop the implicit equality condition. For example:

```javascript
const builder = createQueryBuilder({ selector: { age: 5 } });
builder.where('age').gt(3);
// Expected: { age: { $eq: 5, $gt: 3 } }
// Actual (before fix): { age: { $gt: 3 } }  // $eq: 5 was lost
```

## Solution

Modified the `where()` method in `nosql-query-builder.ts` to detect when an existing condition is a shorthand value (not an object) and convert it to explicit operator form (`{ $eq: value }`) before adding the new operator. This preserves the original equality condition while allowing additional operators to coexist.

The fix handles:
- Numeric/string shorthand values: `{ age: 5 }` → `{ age: { $eq: 5, $gt: 3 } }`
- Null values: `{ name: null }` → `{ name: { $eq: null, $ne: 'bar' } }`
- Undefined values: Creates empty operator object as before

## Tests

Added two comprehensive unit tests:
1. `selector shorthand value should be preserved when chaining another operator on the same field` - Tests numeric shorthand with `gt()` operator
2. `selector shorthand null value should be preserved when chaining another operator` - Tests null shorthand with `ne()` operator

Both tests verify that the implicit `$eq` condition is preserved alongside the new operator.

https://claude.ai/code/session_01HGWfkrrxD3c9zDUUoVSDKa

### DIFF
--- a/test/unit/query-builder.test.ts
+++ b/test/unit/query-builder.test.ts
@@ -150,7 +150,7 @@ describeParallel('query-builder.test.js', () => {
         it('selector shorthand value should be preserved when chaining another operator on the same field', () => {
             // When a selector uses shorthand syntax (e.g., { age: 5 } instead of { age: { $eq: 5 } }),
             // chaining another operator on the same field should preserve the implicit equality condition.
-            const builder = createQueryBuilder<{ age: number }>({
+            const builder = createQueryBuilder<{ age: number; }>({
                 selector: {
                     age: 5
                 }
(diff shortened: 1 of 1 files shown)


pubkey:
@copilot check-types fails

pubkey:
@copilot fix lint

PR #8404, opened 2026-04-20
test: reproduce $elemMatch operator normalization regression

hashdash:
## Summary

This PR adds a failing bug-report test for a regression in `normalizeMangoQuery()` when `$elemMatch` already contains operators like `$eq`, `$regex`, and `$options`.

On `rxdb@17.1.0`, operator payloads inside `$elemMatch` are normalized as if they were selector shorthands. For example:

```ts
{
  roles: {
    $elemMatch: {
      $regex: '^applicant$',
      $options: 'i'
    }
  }
}
```

is rewritten into:

```ts
{
  roles: {
    $elemMatch: {
      $regex: { $eq: '^applicant$' },
      $options: { $eq: 'i' }
    }
  }
}
```

That breaks later query evaluation. In my app this surfaced as:

```txt
Invalid flags supplied to RegExp constructor '[object Object]'
```

## Regression window

- works on `16.21.1`
- fails on `17.1.0`

Likely related upstream change:
- `4766c7a` Normalize selector shorthands recursively in `$and`/`$or`/`$nor`/`$not` (`#8121`)

## Repro

This PR follows the repository issue template and puts the repro into `test/unit/bug-report.test.ts`.

The failing test covers both:
- `$elemMatch: { $regex: '^applicant$', $options: 'i' }`
- `$elemMatch: { $eq: 'Applicant' }`

Both should remain unchanged by `normalizeMangoQuery()`, but are currently wrapped in `$eq`.

## Validation

Ran:

```bash
./node_modules/.bin/eslint test/unit/bug-report.test.ts
npm run transpile
DEFAULT_STORAGE=dexie ./node_modules/.bin/mocha --expose-gc --config ./config/.mocharc.cjs ./test_tmp/unit/bug-report.test.js --grep 'operator payloads inside \$elemMatch'
```

Result:
- lint passes
- transpile succeeds
- targeted test fails with the expected assertion showing `$regex` and `$options` mutated into `{ $eq: ... }`

### DIFF
--- a/test/unit/bug-report.test.ts
+++ b/test/unit/bug-report.test.ts
@@ -1,141 +1,87 @@
-/**
- * this is a template for a test.
- * If you found a bug, edit this test to reproduce it
- * and than make a pull-request with that failing test.
- * The maintainer will later move your test to the correct position in the test-suite.
- *
- * To run this test do:
- * - 'npm run test:node' so it runs in nodejs
- * - 'npm run test:browser' so it runs in the browser
- */
 import assert from 'assert';
-import AsyncTestUtil from 'async-test-util';
-import config from './config.ts';
-
 import {
-    createRxDatabase,
-    randomToken
+    fillWithDefaultSettings,
+    normalizeMangoQuery
 } from '../../plugins/core/index.mjs';
-import {
-    isNode
-} from '../../plugins/test-utils/index.mjs';
+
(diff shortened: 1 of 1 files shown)


pubkey:
Fixed here: https://github.com/pubkey/rxdb/pull/8406

Issue #7986, opened 2026-03-08
replicateSupabase push handler generates excessively long URLs by adding all document fields to WHERE clause


Zoltan3057:
### Description
When using replicateSupabase to synchronize documents with large text/JSON fields (like HTML content or rich text JSON), the push handler fails with 400 Bad Request errors. The issue occurs because the addDocEqualityToQuery function adds all document fields as equality conditions in the PATCH request URL, causing URL length limits to be exceeded.

### Environment
RxDB: 17.0.0-beta.10
Plugin: rxdb/plugins/replication-supabase
Supabase: @supabase/supabase-js latest
Platform: Browser (Chrome/Firefox)

### Steps to Reproduce
Create a Supabase table with text/JSON fields that can contain large content (e.g., HTML, rich text)
Set up RxDB collection with matching schema
Configure replicateSupabase with both pull and push enabled
Update a document that has large content fields
Observe the PATCH request in network tab - all fields are included as query params
Replication fails with 400 Bad Request when URL exceeds server limits

### Root Cause Analysis
The issue is in addDocEqualityToQuery function (packages/replication-supabase/src/supabase-replication.ts). This function is called during the push handler for updates and adds every field from the assumedMasterState document as a WHERE condition to the PATCH request.

```

// Simplified view of what happens
let query = supabase
    .from('table_name')
    .update(updateData);
// This adds ALL fields from assumedMasterState as query parameters
query = addDocEqualityToQuery(assumedMasterState, query);


```

### Expected Behavior
The push handler should generate PATCH requests that:
Use the request body for data, not URL query parameters
Only include necessary fields in WHERE clause for optimistic concurrency (primary key + modified timestamp)
Respect URL length limits (typically 2048-8192 characters)

pubkey:
Seems good, PR is welcomed @Zoltan3057

PR #8432, opened 2026-04-27
fix: handle excessively long URLs by performing local equality checks. ref #7986

len360:
## This PR contains:

- A bug fix
- Improved tests

Without this fix, large payloads were being included in the URL, causing bad request errors. This issue is related to https://github.com/supabase/postgrest-js/issues/393.

## Pull Request Description

This PR contains a bug fix corresponding to issue #7986. It removes the inclusion of all parameters in the URL in order to perform a local comparison of objects instead of making long requests that cause bad request errors.

## Logic implemented

On update events:
1. Fetch the selected document from the server
2. Compare the assumed master state with the server state locally
3. If a difference is detected, a conflict is detected and the current server state is returned
4. Otherwise, the server state is updated

## Unit tests

A unit test has been added to verify that updating a property with a large number of characters does not cause any errors, ensuring that this pull request properly resolves the issue.


len360:
Hi @pubkey, I'm embarrassed because the issue detected in `CI / test-others-replications (pull_request)` can't be reproduced on my computer and seems to appear in multiple GitHub Actions (#44804).

Another thing is that the issue in `Verify Test Reproduction / test-without-fix (pull_request)` mentions a missing issue. This PR is related to issue #7986. Did I miss something to indicate the link between the PR and the ticket, or does it need to be reopened?

src/plugins/replication-supabase/index.ts
```diff
@@ -197,29 +197,30 @@ export function replicateSupabase<RxDocType>(
...
-
-                query = addDocEqualityToQuery(
-                    collection.schema.jsonSchema,
-                    deletedField,
-                    modifiedField,
-                    assumedMasterState,
-                    query
-                );
+                // fetch the current document state from the server
+                const docOnServer: WithDeleted<RxDocType> = await fetchById(id);
```
pubkey, line 201: By fetching here first it is not inside of one transaction anymore. What happens if someone else changes the document on the server after we fetch docOnServer and before sending the new one to the server?

len360: In fact, this implementation seems problematic. If someone modifies the data in the database while another user is fetching it, the last user to write will overwrite the previous changes without any conflict handling.

Another way to address the issue of overly long URLs would be to add a hash property to the object. This "hash" could be included in the URL alongside the identifier, allowing the update to be performed in a single transaction while also helping detect potential conflicts.

What do you think about this approach?

pubkey: That would require other entities that change anything in supabase to use exactly the same hashing and also not forget to update the hash on each changes. Or can this be set in a supabase config?

pubkey:
Closing this, the proposed solution is not safe and could cause wrong behavior because its non-atomic.

PR #8462, opened 2026-05-01
Add live query update throttle option

tsekiguchi:
## Summary
- Add an opt-in `liveQueryUpdateThrottleTime` option that groups change-triggered live query updates before `_ensureEqual()` runs.
- Keep the first live query result immediate, so subscriptions still behave like a BehaviorSubject on startup.
- Add reactive query tests that prove write bursts are grouped and that the latest matching result is still emitted after grouped writes.

## Why this is needed
In our application we have live queries over large folders, including a case with about 155k matching documents. During initial sync and follow-up metadata updates, many writes can touch documents that stay in the same query result.

We first tried to throttle downstream with RxJS after `find(...).$`, but that was too late. RxDB had already received each change bulk and called `_ensureEqual()` before subscribers saw the value. For large result sets this means RxDB may repeatedly run EventReduce or a full storage query, map many documents through the cache, and build another large live query result before application code can throttle it.

The visible effect in the app was renderer stalls during sync. One run showed multi-second raw live-query gaps before our throttling code ran, and repeated emissions of the full 155k result set blocked the renderer. Detaching that large live query after the first final snapshot avoided the repeated work, but that is an app-level workaround and loses the normal live-query behavior.

This option gives applications a small upstream control point at the place where it matters: before `_ensureEqual()` does the expensive work. It defaults to disabled, so existing RxDB behavior is unchanged.

## Test plan
- `npm run lint`
- `npm run build`
- `npm run check-types`
- `npm run test:fast:memory`
- Focused check: `npm run transpile && npm run build:plugins && npx cross-env DEFAULT_STORAGE=memory NODE_ENV=fast mocha --config ./config/.mocharc.cjs ./test_tmp/unit/reactive-query.test.js --grep "groups live query write bursts|emits the latest matching result after grouped writes"`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/rx-query.ts
+++ b/src/rx-query.ts
@@ -1,9 +1,11 @@
 import {
     BehaviorSubject,
     Observable,
-    merge
+    merge,
+    of
 } from 'rxjs';
 import {
+    auditTime,
     mergeMap,
     filter,
     map,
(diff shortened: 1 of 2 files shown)


pubkey:
Needs updates in docs and we have to add this to the typings.

pubkey:
I think the throttling should self-increase after the first few times. So after the first change we do not wait. Then we wait half the time and the we wait the full throttle time. Or maybe the throttling should be more defined like MAX_X_EMITS_PER_SECOND.

pubkey:
replaced by https://github.com/pubkey/rxdb/pull/8466

PR #8475, opened 2026-05-05
Block writes during schema migration to prevent conflicts

pubkey:
## This PR contains:
- A BUGFIX
- IMPROVED TESTS
- IMPROVED DOCS

When a schema migration is running on an RxCollection, external writes (insert, upsert, bulkInsert, etc.) could race with the migration's replication process. This could cause confusing `RC_PUSH` errors or other conflicts instead of failing fast with a clear error message.

Additionally, writes were allowed even when a migration was pending but not yet started, which could interfere with the migration process.

## Solution

This PR implements a `migrationInProgress` flag on RxCollection that:

1. **Blocks all writes during migration**: The flag is set when a migration starts and cleared when it completes, fails, or is cancelled. Any write attempt while the flag is true throws a clear `COL25` error.

2. **Blocks writes when migration is pending**: The flag is optimistically set when a collection is created (for schema version > 0), then cleared if no migration is actually needed. This prevents writes from interfering with pending migrations.

3. **Provides clear error messaging**: Instead of racing errors, users get an immediate `COL25` error with guidance to wait for the migration to complete.

## Changes made:

- Added `migrationInProgress` property to `RxCollectionBase`
- Created `isWriteAllowed()` helper function that checks both closed state and migration state
- Replaced `ensureRxCollectionIsNotClosed()` calls with `isWriteAllowed()` in all write methods (insert, bulkInsert, upsert, bulkUpsert, incrementalUpsert, bulkRemove)
- Added migration plugin hook to set `migrationInProgress` flag when collection is created
- Updated `RxMigrationState.startMigration()` to manage the flag throughout the migration lifecycle
- Added comprehensive tests covering: blocking writes during migration, blocking writes when migration is pending, resetting flag on error, and racing scenarios
- Added `COL25` error definition and documentation warning

## Test Plan

Added 4 new unit tests in `migration-schema.test.ts`:
- Verifies writes are blocked while migration runs and allowed after completion
- Verifies bulkInsert that races with migration start is blocked
- Verifies writes are blocked when migration is needed but not yet started
- Verifies `migrationInProgress` flag is reset on migration error

All tests pass and cover the key scenarios where writes could interfere with migrations.

https://claude.ai/code/session_01PFvmQnnz2GCozXqMfCjYvy


pubkey:
@copilot run the dexie tests and fix them.

PR #8496, opened 2026-05-11
chore(deps): bump dexie 4.0.10 -> 4.4.2

sahanxdissanayake:
## Summary

Bumps `dexie` from 4.0.10 (released 2024-11-15) to 4.4.2 (released 2026-03-31) to pick up dexie 4.2.1's fix for [#2186](https://github.com/dexie/Dexie.js/issues/2186) ([PR #2187](https://github.com/dexie/Dexie.js/pull/2187)):

> *"Resolve issue #2186 by letting an external closure of the DB make it behave identical to explicit closing the database, except that we still respect autoOpen option."*

## Why this matters for RxDB downstream

In an RxDB 17.1.0 application using `getRxStorageDexie()` on iOS Safari, the dexie #2186 latent state-inconsistency is producing three error variants at production volume:

| Error | Source |
| --- | --- |
| `AbortError: Transaction aborted` | dexie wrapper |
| `BulkError: docs.bulkPut(): 1 of 1 operations failed. Errors: UnknownError: Attempt to delete range from database without an in-progress transaction` | WebKit IDB |
| `UnknownError: Attempt to get a record from database without an in-progress transaction` | WebKit IDB |

Mechanism:
1. iOS jetsam kills the `com.apple.WebKit.Storage` XPC process under memory pressure / app backgrounding.
2. In pre-4.2.1 dexie, `Dexie.on.close` fires but `Dexie.isOpen()` still returns `true` — internal state is inconsistent.
3. RxDB's next `bulkPut` / `get` opens a transaction against the (now dead) connection.
4. The next op on that transaction throws one of the variants above.

After bumping to dexie 4.2.1+, external closure transitions Dexie to a real closed state, surfacing the clean `DatabaseClosedError` which applications can filter / recover from.

The class of error has been observed to produce silent local-write loss in an offline-first admin tool — exactly the failure mode this codebase aims to prevent. The RxDB 17 internal bulkWrite surface increase appears to have amplified how often the latent dexie bug fires (first-seen jumped 9 days after a downstream RxDB 16 → 17 bump).

## Compatibility verified

Locally against this tag (`pubkey/rxdb@17.1.0`):

```bash
git clone --depth 1 --branch 17.1.0 https://github.com/pubkey/rxdb.git
# bump dexie to 4.4.2 in package.json
yarn install --ignore-scripts
yarn run build
yarn run build:test
cross-env DEFAULT_STORAGE=dexie NODE_ENV=fast \
  ./node_modules/.bin/mocha --config ./config/.mocharc.cjs ./test_tmp/unit.test.js
```

Result: **1000 / 1000 dexie tests pass**. The single failure was `replication-webrtc.test.ts > import WebRTC polyfills on Node.js` — unrelated to dexie (it's a `node_datachannel.node` native binding install issue on darwin-arm64 when `--ignore-scripts` is used to skip the prebuild).

## Note for review

This bump also requires `rxdb-old`'s nested dexie to be updated. `rxdb-old` (a test dep used for backward-compat tests) pins its own `dexie@4.0.10` in its nested `node_modules`. At Dexie construction time the dedup check throws `Two different versions of Dexie loaded in the same app: 4.0.10 and 4.4.2`. To run the suite locally I symlinked `node_modules/rxdb-old/node_modules/dexie` to the top-level dexie. The correct upstream fix is to bump `rxdb-old` to a version whose own dexie pin is 4.4.2 (or similarly compatible).

## Context

Prior renovate bump PRs (#7387 4.2.0, #7443 4.2.1, #7764 4.3.0) were each closed without explanatory comment. PR #7948 then added `dexie` to `renovate.json:ignoreDeps`. If there is a specific known incompatibility, a one-line note would be very helpful — I can dig further and adjust this PR. Otherwise this looks like the right path to resolving a real iOS-Safari issue for downstream users without forcing them to use a package-manager override.

## Refs

- [dexie #2186](https://github.com/dexie/Dexie.js/issues/2186) — Once Dexie.on.close is triggered, Dexie.isOpen() is still true
- [dexie PR #2187](https://github.com/dexie/Dexie.js/pull/2187) — the fix
- [dexie 4.2.1 release notes](https://github.com/dexie/Dexie.js/releases/tag/v4.2.1)


package.json
```diff
@@ -710,7 +710,11 @@
   },
   "overrides": {
     "eslint": "10.2.1",
-    "webpack-dev-server": "5.2.3"
+    "webpack-dev-server": "5.2.3",
+    "dexie": "4.4.2"
```
pubkey, line 713: why do we need these overwrites?

sahanxdissanayake: Both blocks are needed because of the `rxdb-old` devDep — it's an npm alias for the old RxDB release (`"rxdb-old": "npm:rxdb@16.21.1"`) that brings its own nested `dexie@4.0.10` in `node_modules/rxdb-old/node_modules/dexie`. With the top-level dexie bumped to 4.4.2 and the nested copy left at 4.0.10, dexie's own module-load-time dedup check throws:

```
Error: Two different versions of Dexie loaded in the same app: 4.0.10 and 4.4.2
    at node_modules/rxdb-old/node_modules/dexie/import-wrapper.mjs:8:11
```

Every storage-using test then fails at import time before any test code runs — this was the actual red CI on my first push (`storage-dexie`, `storage-memory`, `storage-localstorage`, `storage-sqlite-trial`, `storage-mongodb`, `storage-remote` all failed with the same message).

Two blocks because CI and local use different package managers:
- **`overrides`** — for `npm install`, which is what `.github/workflows/main.yml` runs on every job.
- **`resolutions`** — for `yarn install`, matching the `packageManager: yarn@1.22.22+...` field for local dev.

Without `overrides`: CI breaks. Without `resolutions`: anyone cloning the repo and running `yarn install` (the documented contributor flow) breaks. With both, the nested dexie under `rxdb-old` gets coerced to 4.4.2 in both environments and the dedup check is satisfied.

If you prefer a tighter form, both can be path-scoped to only affect `rxdb-old`:

```jsonc
"overrides": {
  "eslint": "10.2.1",
  "webpack-dev-server": "5.2.3",
  "rxdb-old": {
    "dexie": "4.4.2"
  }
},
"resolutions": {
  "rxdb-old/**/dexie": "4.4.2"
}
```

That makes the intent self-documenting — the top-level `dependencies.dexie` change carries the actual upgrade; the override only exists to keep `rxdb-old`'s nested copy in lock-step. Happy to push that as a follow-up commit if you'd like.

The truly clean fix (bump `rxdb-old` to a newer rxdb release with a newer dexie) doesn't work yet because every released `17.x` (17.0, 17.1, 17.2) still pins `dexie 4.0.10`. That has to land first.

pubkey: makes sense, keep the overwrite.

pubkey:
Please remove that test case from the PR. Updating dexie is good enough, we do not want to test dexie internals.

PR #8522, opened 2026-05-16
schema using 'patternProperties' with square bracket character classes gets interpreted as array access

KampfCaspar:
## This PR contains:

BUG TEST CASE

## Describe the problem

I want to include objects as fields in an RxJsonSchema. If I specify that deep objects properties by *patternProperties* I want
to use *character classes* in the regex. rxdb seems to interpret the square brackets (originating from the regex character classes) as indicating array indizes and bails out.

## Describe solution

Regexs in *patternProperties* should NOT be interpreted as array indizes - at least not directly. *If* such indizes serve a purpose, only interpret square brackets that are *not* part of the regex as such (e.g. escaped by backslash for a 'literal' square bracket).

### DIFF
--- a/src/plugins/dev-mode/check-schema.ts
+++ b/src/plugins/dev-mode/check-schema.ts
@@ -496,6 +496,7 @@ export function checkSchema(jsonSchema: RxJsonSchema<any>) {
             return split.join('.');
         })
         .filter(key => key !== '')
+        .filter(key => key.indexOf('.patternProperties.') == -1 ) // regex in patternProperties not interpreted
         .filter((elem, pos, arr) => arr.indexOf(elem) === pos) // unique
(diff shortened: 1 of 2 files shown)


KampfCaspar:
It seems the issue resides in plugins/dev-mode/check-schema, while patternProperties seem to be excluded in validateFieldsDeep (line 78), it is *not* handled in checkSchema (line 501).

The block seems to check for an older configuration variant with 'index: true' on individual fields. I don't know if that ever worked for patternProperties - the 'flattened' object path and regex syntax just don't mix.

I therefore suggest excluding patternProperties from this old-config check.

pubkey:
Fixed at https://github.com/pubkey/rxdb/pull/8577

PR #8527, opened 2026-05-17
add test case for a beforeInsert hook on a composite primary key

KampfCaspar:
## This PR contains:
BUG TEST CASE

## Describe the problem you have

I want to dynamically but centrally assign primary keys for new documents in a pre-insert hook. This works fine for *scalar primary keys* but fails for *composite primary keys*

## Describe the solution

Best: fields for a composite primary key can bet set in a pre-Hook.
Other: primary key fields (irrespective of primary key type) are taboo.

### DIFF
--- a/test/unit/bug-report.test.ts
+++ b/test/unit/bug-report.test.ts
@@ -9,54 +9,51 @@
  * - 'npm run test:browser' so it runs in the browser
  */
 import assert from 'assert';
-import AsyncTestUtil from 'async-test-util';
 import config from './config.ts';
 
 import {
     createRxDatabase,
     randomToken
 } from '../../plugins/core/index.mjs';
-import {
-    isNode
-} from '../../plugins/test-utils/index.mjs';
 describe('bug-report.test.js', () => {
     it('should fail because it reproduces the bug', async function () {
 
-        /**
-         * If your test should only run in nodejs or only run in the browser,
-         * you should comment in the return operator and adapt the if statement.
-         */
-        if (
-            !isNode // runs only in node
-            // isNode // runs only in the browser
-        ) {
-            // return;
-        }
-
-        if (!config.storage.hasMultiInstance) {
-            return;
-        }
-
-        // create a schema
-        const mySchema = {
+        const schemaScalar = {
             version: 0,
-            primaryKey: 'passportId',
+            primaryKey: 'id',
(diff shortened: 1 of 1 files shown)


test/unit/bug-report.test.ts
```diff
@@ -80,62 +77,27 @@ describe('bug-report.test.js', () => {
...
-            mycollection: {
-                schema: mySchema
+            collectionScalar: {
+                schema: schemaScalar
+            },
+            collectionComposite: {
+                schema: schemaComposite
             }
         });
+        collections.collectionScalar.preInsert( (data) => {
```
pubkey, line 87: why do you use two different hooks? I think you need a single one that asigns id, name and number at the same time.

KampfCaspar: If I understood the documentation correctly, I should only set the composite fields of a composite primary key. The 'primary key field' - while I seem to have to define it myself - should be set by rxdb itself (using the composite fields).

Therefore in the *scalar* example, I *directly* set the id field - whereas in the *composite* example, I set name and number (and rxdb should combine them into id itself).

The two pre-insert hooks fill different fields.

KampfCaspar:
I researched the case further and found a further complication:

It seems in fillObjectDataBeforeInsert (line 40) tries to fill in the primary key field (before the pre-insert hook) but ONLY on composite primary keys - ordinary scalar (string) primary key fields are left alone. This results in a scalar primary key being fillable by the pre-insert hook, whereas a composite primary key bails out before the hook is called.

now insert/bulkInsert and bulkUpsert (which calls bulkInsert) could be easily be adapted, but bulkUpsert calls fillObjectDataBeforeInsert effectively twice and checks the existence of (any) primary key itself...  Might it be prudent to split the 'document preparation' from the actual 'document insertion'?

incerementalUpsert, however, uses the primary key for further concurrency checks. Therefore it definitely needs the 'combined' composite primary key even earlier - definitely *before* pre-insert is run.

If setting primary key fields in a pre-hook should be supported, it seems a separate 'document preparation hook' to be called even before the pre-insert hook might be prudent?

pubkey:
Fixed in https://github.com/pubkey/rxdb/pull/8550

Please test.

KampfCaspar:
works fine. Thx.

Issue #8609, opened 2026-06-03
upsertLocal() 409 conflict


epodgaysky:
`upsertLocal()` throwing a 409 conflict when a local document exists in storage but was evicted from the local-documents `docCache`

epodgaysky:
ping

pubkey:
fixed https://github.com/pubkey/rxdb/pull/8674

PR #8610, opened 2026-06-03
fixed upsertLocal 409 conflict after the document evicted from cache

epodgaysky:
Fix for an issue https://github.com/pubkey/rxdb/issues/8609


pubkey:
https://github.com/pubkey/rxdb/issues/8609#issuecomment-4768060410

PR #8688, opened 2026-06-23
feat(replication): add replicationState.awaitDocumentPushed()

### LINKED ISSUE #8632: Feature request: awaitable per-write push acknowledgement (opt-in write-through for replication)
An integrated, opt-in way to `await` the **server's acceptance of one specific local write** through the existing replication machinery — a "write-through" mode for individual writes, while everything else stays local-first.

```ts
// Sketch — exact shape up to you, see options below
const doc = await collection.upsert(data);
const ack = await replicationState.awaitDocumentPushed(doc.primary, { timeout: 5000 });
// ack: { status: 'accepted' }
//    | { status: 'conflict', masterState }   — server rejected/resolved differently
// throws on timeout / replication cancelled / offline
```

Local-first is the right default and we love it. But almost every app has a handful of writes where the user must synchronously know "did the server take it?" — admin/settings forms with server-side validation, anything money-adjacent, uniqueness-constrained fields. Today those writes have to leave the replication world entirely, which creates real bugs (war story below).

## Why the existing primitives don't cover it

- **`awaitInSync()`** is collection-wide, not per-document; your own docs (rightly) warn against blocking the app on it; with `multiInstance: true` + `waitForLeadership: true` it never resolves in a non-leader tab; and offline it simply never settles — so it can't back a "Save" button.
- **`sent$` / `received$` / `error$`** emit documents and errors, but there is no way to correlate *my* write with *its* push outcome. Worse, a server-side **rejection** is expressed as a conflict row in the push handler's return value — RxDB then resolves it via the `conflictHandler` and the local state quietly converges to the master. Correct for sync; invisible to the caller who needs to show "VAT id invalid" on the form.
- **Hand-rolling a parallel REST endpoint** (what we do now) means two write paths for the same entity, duplicated validation, and a footgun we actually shipped: after the REST save we patched ONE field of the saved entity into the local RxDB doc (so local readers update immediately) — that local write marked the doc dirty, replication pushed the **whole stale document**, and last-write-wins reverted the values the REST call had just written. Entirely our bug, but it's the kind of bug this feature would make unnecessary: with an awaitable push we'd have written through replication in the first place — one write path, one roundtrip.

## Proposed semantics

1. The local write happens normally (optimistic, local-first untouched).
2. The caller can await the outcome of that document's next push:
   - **accepted** — the push handler returned no conflict for the row → resolve.
   - **conflict/rejected** — the push handler returned the row in its conflicts array → the normal `conflictHandler` flow runs (so state converges exactly as today), and the awaited promise resolves with `{ status: 'conflict', masterState }` (or rejects, bikeshed) so the UI can surface the server's reason.
   - **offline / cancelled / timeout** — reject fast with a typed error; the write stays local and syncs later (caller decides whether to revert).
3. Optionally a way to **flush immediately** (`replicationState.pushNow()` or an option on the write) so the ack doesn't wait for batching/debounce.
4. Must work from non-leader tabs under `multiInstance` (forward to the leader or temporarily push from the calling tab).

Per-row outcome data already exists internally — the push handler's return value is row-granular — so this is mostly about exposing a correlation point per document, not changing the protocol.

## API shape options

```ts
// A) On the replication state (least invasive)
await replicationState.awaitDocumentPushed(docId, { timeout?, flush?: boolean });

// B) Per-write opt-in on the collection
await collection.upsert(data, { awaitPush: { replicationState?, timeout? } });

// C) A combined helper with optional revert-on-reject
await replicationState.writeThrough(
  () => collection.upsert(data),
  { timeout: 5000, revertOnConflict: true },
);
```

A) feels most aligned with the existing `RxReplicationState` API (`awaitInSync` family). C)'s `revertOnConflict` is sugar over the standard conflict resolution and could come later.

## Alternatives considered

- `upsert()` + `awaitInSync()` + reading the doc back and diffing: racy (another pull can interleave), collection-wide, blocks on unrelated pending writes, and cannot distinguish "accepted" from "conflict-resolved back to master".
- Subscribing to `sent$`/`error$` with manual correlation: `sent$` fires on send, not on accept; rejections surface in neither stream as a per-row outcome.
- Parallel REST endpoint: works, but splits the write path and reintroduces exactly the consistency bugs replication exists to prevent (see war story).
- Prior art: Firestore's `waitForPendingWrites()` has the same collection-wide limitation; Replicache's mutators give per-mutation server acks — that's the ergonomics we're after, inside RxDB's protocol.

## Environment

RxDB 16.x, `replicateRxCollection` with custom HTTP push/pull handlers (multi-tenant offline-first B2B app; ~40 replicated collections).

pubkey:
## Summary

Implements the feature requested in #8632: a per-document push acknowledgement on the replication state.

`replicationState.awaitDocumentPushed(rxDocument)` returns a promise that resolves once the given `RxDocument` instance was successfully pushed to the server.

```ts
const doc = await myCollection.insert({ id: 'foobar', value: 10 });
await myReplicationState.awaitDocumentPushed(doc);
// here we know that this document state reached the master
```

## How it works

Each `RxDocument` carries its write time in `_meta.lwt`. The upstream (push) checkpoint has the default RxStorage checkpoint shape `{ id, lwt }` and moves forward with every successful push, in the sort order `[_meta.lwt ASC, primaryKey ASC]`. A document state is considered pushed once it is no longer "after" the checkpoint, which mirrors the comparison used by `getChangedDocumentsSinceQuery()`.

The implementation:
- does an immediate check against the current upstream checkpoint
- otherwise re-checks on each `sent$` emission (when a document is sent) and when the replication becomes idle (`active$` → `false`), at which point the checkpoint for that push cycle has been written
- if the document was overwritten by a newer local write before it could be pushed, the promise still resolves once any later state (or any document with a higher write time) was pushed, since that proves the given state reached the server

No timeout is built in by design. The docs show how to combine it with `Promise.race()`.

## Behavior notes

- Calling it on a pull-only replication throws `RC_PUSH_AWAIT`.
- If the replication is canceled before the document was pushed, the promise stays pending, matching the behavior of `awaitInitialReplication()`.

## Changes

- `src/plugins/replication/index.ts`: new `awaitDocumentPushed()` method and `isDocumentStateOlderThenCheckpoint()` helper
- `src/plugins/dev-mode/error-messages.ts`: new error code `RC_PUSH_AWAIT`
- `test/unit/replication.test.ts`: tests for resolve-after-push, resolve-immediately-if-already-pushed, not-resolve-before-push, document-inserted-before-replication, and the pull-only throw
- `docs-src/docs/replication.md`: new `awaitDocumentPushed()` section with a `Promise.race()` timeout example
- changelog entry

Closes #8632

🤖 Generated with <a href="https://claude.com/claude-code">Claude Code</a>

---
_Generated by <a href="https://claude.ai/code/session_01W7jgHSwWQBwmDBdagfUYdi">Claude Code</a>_

### DIFF
--- a/docs-src/docs/replication.md
+++ b/docs-src/docs/replication.md
@@ -552,16 +552,10 @@ If the document was overwritten by a newer local write before it could be pushed
 `awaitDocumentPushed()` does not set a timeout on purpose. If you need one, combine it with `Promise.race()`:
 
 ```ts
-function timeout(ms: number) {
-    return new Promise((_res, reject) => {
-        setTimeout(() => reject(new Error('push timeout')), ms);
-    });
-}
-
 const doc = await myCollection.insert({ id: 'foobar', value: 10 });
 await Promise.race([
     myReplicationState.awaitDocumentPushed(doc),
-    timeout(5000)
+    new Promise((_, reject) => setTimeout(() => reject(new Error('timeout')), 5000))
 ]);
 ```
 
--- a/docs-src/docs/replication.md
+++ b/docs-src/docs/replication.md
@@ -545,6 +545,8 @@ await myReplicationState.awaitDocumentPushed(doc);
 // here we know that the document state was pushed to the server
 ```
 
+A `RxDocument` represents the state of a document at a given point in time, not the document in general. An older or a newer `RxDocument` instance of the same document is not the exact same `RxDocument`, because each instance carries the field values and the internal write time of that specific state. `awaitDocumentPushed()` therefore resolves based on the exact state of the instance you pass in.
+
 It works by comparing the documents internal write time (`_meta.lwt`) with the push checkpoint. The push checkpoint moves forward with each successful push, so once it covers the documents write time, RxDB knows that this version of the document has been replicated to the master.
 
 If the document was overwritten by a newer local write before it could be pushed, the promise resolves as soon as any later state of that document (or any document with a higher write time) has been pushed, because that also proves the given state reached the server.
--- a/docs-src/docs/replication.md
+++ b/docs-src/docs/replication.md
@@ -531,6 +531,40 @@ await hideLoadingSpinner();
 :::
 
 
+### awaitDocumentPushed()
+
+Returns a `Promise` that resolves when a specific `RxDocument` instance was successfully pushed to the server.
+
+While `awaitInSync()` waits for the whole collection to be in sync, `awaitDocumentPushed()` only waits for a single document. This is useful when you have a sensitive write (like a financial transaction or a value with a uniqueness constraint) and you want to confirm that exactly this write reached the backend, without blocking on every other unrelated change in the collection.
+
+You pass the `RxDocument` instance you got back from a write operation:
+
+```ts
+const doc = await myCollection.insert({ id: 'foobar', value: 10 });
+await myReplicationState.awaitDocumentPushed(doc);
+// here we know that the document state was pushed to the server
+```
+
+It works by comparing the documents internal write time (`_meta.lwt`) with the push checkpoint. The push checkpoint moves forward with each successful push, so once it covers the documents write time, RxDB knows that this version of the document has been replicated to the master.
+
+If the document was overwritten by a newer local write before it could be pushed, the promise resolves as soon as any later state of that document (or any document with a higher write time) has been pushed, because that also proves the given state reached the server.
+
+`awaitDocumentPushed()` does not set a timeout on purpose. If you need one, combine it with `Promise.race()`:
+
+```ts
+function timeout(ms: number) {
+    return new Promise((_res, rej) => setTimeout(() => rej(new Error('push timeout')), ms));
+}
+
+const doc = await myCollection.insert({ id: 'foobar', value: 10 });
+await Promise.race([
+    myReplicationState.awaitDocumentPushed(doc),
+    timeout(5000)
+]);
+```
+
+Calling `awaitDocumentPushed()` on a replication that has no `push` handler (pull-only) throws an error, because such documents are never sent to the server.
+
 ### reSync()
 
 Triggers a `RESYNC` cycle where the replication goes into [checkpoint iteration](#checkpoint-iteration) until the client is in sync with the backend. Used in unit tests or when no proper `pull.stream$` can be implemented so that the client only knows that something has been changed but not what.
--- a/docs-src/docs/replication.md
+++ b/docs-src/docs/replication.md
@@ -561,8 +561,6 @@ await Promise.race([
 ]);
 ```
 
-Calling `awaitDocumentPushed()` on a replication that has no `push` handler (pull-only) throws an error, because such documents are never sent to the server.
-
 ### reSync()
 
 Triggers a `RESYNC` cycle where the replication goes into [checkpoint iteration](#checkpoint-iteration) until the client is in sync with the backend. Used in unit tests or when no proper `pull.stream$` can be implemented so that the client only knows that something has been changed but not what.
--- a/docs-src/docs/replication.md
+++ b/docs-src/docs/replication.md
@@ -553,7 +553,9 @@ If the document was overwritten by a newer local write before it could be pushed
 
 ```ts
 function timeout(ms: number) {
-    return new Promise((_res, rej) => setTimeout(() => rej(new Error('push timeout')), ms));
+    return new Promise((_res, reject) => {
+        setTimeout(() => reject(new Error('push timeout')), ms);
+    });
 }
 
 const doc = await myCollection.insert({ id: 'foobar', value: 10 });
--- a/src/plugins/replication/index.ts
+++ b/src/plugins/replication/index.ts
@@ -522,18 +522,17 @@ export class RxReplicationState<RxDocType, CheckpointType> {
 
     /**
      * Returns a promise that resolves when the given RxDocument instance
-     * (more precisely: the exact document state with its current _meta.lwt)
      * was successfully pushed to the server.
      *
-     * It works by comparing the documents _meta.lwt with the upstream (push)
-     * checkpoint. The upstream checkpoint moves forward with each successful
-     * push, so once the checkpoint covers the documents write time, we know
-     * that this version of the document has been replicated to the master.
+     * It resolves either when the document is emitted on sent$ (the live
+     * push case) or, for documents that were already pushed before this was
+     * called, when the upstream (push) checkpoint already covers the
+     * documents write time (_meta.lwt).
      *
      * If the document was overwritten by a newer local write before it could
      * be pushed, the promise resolves as soon as any later state of the
-     * document (or any other document with a higher write time) has been
-     * pushed, because that also proves the given state reached the server.
+     * document has been pushed, because that also proves the given state
+     * reached the server.
      */
     async awaitDocumentPushed(doc: RxDocument<RxDocType>): Promise<void> {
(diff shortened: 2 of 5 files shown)


docs-src/docs/replication.md
```diff
@@ -531,6 +531,40 @@ await hideLoadingSpinner();
...
+```
+
+It works by comparing the documents internal write time (`_meta.lwt`) with the push checkpoint. The push checkpoint moves forward with each successful push, so once it covers the documents write time, RxDB knows that this version of the document has been replicated to the master.
+
+If the document was overwritten by a newer local write before it could be pushed, the promise resolves as soon as any later state of that document (or any document with a higher write time) has been pushed, because that also proves the given state reached the server.
+
+`awaitDocumentPushed()` does not set a timeout on purpose. If you need one, combine it with `Promise.race()`:
+
+```ts
+function timeout(ms: number) {
```
pubkey: @copilot inline this to make the example shorter

docs-src/docs/replication.md
```diff
@@ -531,6 +531,40 @@ await hideLoadingSpinner();
...
+
+You pass the `RxDocument` instance you got back from a write operation:
+
+```ts
+const doc = await myCollection.insert({ id: 'foobar', value: 10 });
+await myReplicationState.awaitDocumentPushed(doc);
+// here we know that the document state was pushed to the server
+```
+
+It works by comparing the documents internal write time (`_meta.lwt`) with the push checkpoint. The push checkpoint moves forward with each successful push, so once it covers the documents write time, RxDB knows that this version of the document has been replicated to the master.
```
pubkey, line 550: explain that another older or newer RxDocument instance of that document is not the exact same RxDcoument. a RxDocument represents the state of the document in a given point in time @copilot

src/plugins/replication/index.ts
```diff
@@ -518,6 +520,91 @@ export class RxReplicationState<RxDocType, CheckpointType> {
...
+     * checkpoint. The upstream checkpoint moves forward with each successful
+     * push, so once the checkpoint covers the documents write time, we know
+     * that this version of the document has been replicated to the master.
+     *
+     * If the document was overwritten by a newer local write before it could
+     * be pushed, the promise resolves as soon as any later state of the
+     * document (or any other document with a higher write time) has been
+     * pushed, because that also proves the given state reached the server.
+     *
+     * Notice that this does NOT set a timeout. If you need one, combine it
```
pubkey: @copilot here we do not need the hint about timeout

src/plugins/replication/index.ts
```diff
@@ -518,6 +520,91 @@ export class RxReplicationState<RxDocType, CheckpointType> {
...
+        if (!this.push) {
+            throw newRxError('RC_PUSH_AWAIT', {
+                id: doc.primary,
+                args: { replicationIdentifier: this.replicationIdentifier }
+            });
+        }
+        await this.startPromise;
+        const internalReplicationState = ensureNotFalsy(this.internalReplicationState);
+        const primaryPath = this.collection.schema.primaryPath;
+        const docId: string = (doc._data as any)[primaryPath];
```
pubkey: @copilot dont we have myRxDoucment.primary?

src/plugins/replication/index.ts
```diff
@@ -518,6 +520,91 @@ export class RxReplicationState<RxDocType, CheckpointType> {
...
+                docLwt
+            );
+        };
+
+        if (await isPushed()) {
+            return;
+        }
+
+        return new Promise<void>((resolve, reject) => {
+            const sub = internalReplicationState.events.active.up.subscribe((active: boolean) => {
```
pubkey: @copilot shouldnt we use myRxReplicationState.sent$ here?

src/plugins/replication/index.ts
```diff
@@ -617,6 +704,32 @@ export class RxReplicationState<RxDocType, CheckpointType> {
...
+ * _meta.lwt write time) is already covered by the upstream (push) checkpoint.
+ *
+ * The upstream checkpoint has the default RxStorage checkpoint shape
+ * { id, lwt } and points to the last document that was processed by the
+ * push in the sort order [_meta.lwt ASC, primaryKey ASC].
+ *
+ * A document state is considered pushed when it is NOT after the checkpoint,
+ * which mirrors the comparison used by getChangedDocumentsSinceQuery().
+ */
+export function isDocumentStatePushedToMaster(
```
pubkey: @copilot call thsi function isDocumentStateOlderThenCheckpoint

test/unit/replication.test.ts
```diff
@@ -805,6 +805,131 @@ describe('replication.test.ts', () => {
                 remoteCollection.database.close();
             });
         });
+        describe('.awaitDocumentPushed()', () => {
+            it('should resolve after the document was pushed to the master', async () => {
```
pubkey, line 809: @copilot add test where the insert happens before the replication state is created

src/plugins/replication/index.ts
pubkey: Done in 03ff70e, now triggered by `sent$`.

One note: `sent$` is used as the trigger, but I kept the `isDocumentStateOlderThenCheckpoint()` comparison as the actual resolve condition instead of resolving directly on the emitted document. Two reasons:
- `sent$` emits `WithDeleted<RxDocType>` without `_meta`, so it does not carry the `_meta.lwt` needed to match the exact instance version.
- `sent$` fires before the push checkpoint is persisted, and matching only by primary key could resolve on an older in-flight push of the same document. Comparing against the checkpoint (after `await streamQueue.up`) keeps the point-in-time precision.

The same commit also renames the helper to `isDocumentStateOlderThenCheckpoint` and adds the test where the insert happens before the replication state is created.

---

docs-src/docs/replication.md
```diff
@@ -531,6 +531,38 @@ await hideLoadingSpinner();
...
+
+```ts
+const doc = await myCollection.insert({ id: 'foobar', value: 10 });
+await Promise.race([
+    myReplicationState.awaitDocumentPushed(doc),
+    new Promise((_, reject) => setTimeout(() => reject(new Error('timeout')), 5000))
+]);
+```
+
+Calling `awaitDocumentPushed()` on a replication that has no `push` handler (pull-only) throws an error, because such documents are never sent to the server.
```
pubkey: we do not need that hint here in the docs

src/plugins/replication/index.ts
```diff
@@ -518,6 +520,90 @@ export class RxReplicationState<RxDocType, CheckpointType> {
...
+        const docLwt: number = doc._data._meta.lwt;
+
+        const isPushed = async (): Promise<boolean> => {
+            /**
+             * Wait for the currently running upstream cycle and its checkpoint
+             * write to finish so that lastCheckpointDoc.up reflects the latest
+             * pushed state. The checkpoint is written at the end of a push
+             * cycle, so we must not read it while a cycle is still in progress.
+             */
+            await internalReplicationState.streamQueue.up;
```
pubkey: i do not think we should await this here, before isPushed is called another cycled could have started already that does not affect our doucment anyways.

pubkey: Removed in b5dc0f0. You are right.

I dropped the `streamQueue.up` await entirely and now resolve directly on the `sent$` emission whose primary key matches the document. That emission already proves the document reached the master, so there is no need to wait for any cycle or re-read the checkpoint. The `isDocumentStateOlderThenCheckpoint()` comparison is now only used for the initial check (documents that were already pushed before `awaitDocumentPushed()` was called).

---

PR #8797, opened 2026-07-15
FIX dexie storage: close Dexie connections and evict state cache on last close (#8793)

### LINKED ISSUE #8793: Dexie RxStorage: reference counting in dexie-helper never closes Dexie connections — reopening a database name after deletion fails (DatabaseClosedError + DB8)
## Affected versions

17.3.0 and 17.4.0 (current), Dexie-based RxStorage. See the analysis below for why
this is version-independent.

## Summary

While debugging a `DatabaseClosedError` after a close → delete → reopen cycle in our
app, I ended up in `src/plugins/storage-dexie/dexie-helper.ts` and - unless I'm
misreading the code - the reference counting around `DEXIE_STATE_DB_BY_NAME` /
`REF_COUNT_PER_DEXIE_DB` seems to be broken in two independent ways:

1. **The counter is never incremented anywhere.** The only writes to
   `REF_COUNT_PER_DEXIE_DB` in the entire package are `set(x, 0)` at creation and
   `set(x, count - 1)` in `closeDexieDb()`. Starting from 0, the counter can only
   ever be 0, -1, -2, ... so the condition `newCount === 0` (which guards
   `state.dexieDb.close()` and the cache eviction) can never become true. The close
   branch is unreachable.

2. **The initial counter is stored under the wrong key.** Inside the
   `getFromMapOrCreate` factory callback, `REF_COUNT_PER_DEXIE_DB.set(state, 0)`
   runs before `const state = ...` has been assigned (the callback executes
   synchronously during that assignment), so the key is effectively `undefined`.
   `closeDexieDb()` later looks up the actual state promise, gets `undefined`, and
   computes `undefined - 1 = NaN`. The `(prevCount as any) - 1` cast in the source
   silences the TypeScript error that would have flagged this. This part appears to
   have changed with the "REFACTOR Map cache code" commit (#4604), which moved those
   lines into the factory callback.

## Consequences

- `RxStorageInstanceDexie.close()` never closes the underlying Dexie/IndexedDB
  connections (connection leak), and the `DEXIE_STATE_DB_BY_NAME` entry is never
  removed.
- If the application afterwards deletes the IndexedDB databases (e.g.
  `indexedDB.deleteDatabase(...)` on logout, to remove user data from a shared
  device), the delete request is blocked by the leaked connections; Dexie
  force-closes them ("Another connection wants to delete database ..."). The module
  cache now holds force-closed Dexie instances.
- Re-creating an RxDatabase with the same name in the same JS context then receives
  the closed Dexie instance from the cache: the first storage operation throws
  Dexie's `DatabaseClosedError` ("Database has been closed"), and a retry fails with
  RxDB error DB8 because the half-created database is still registered.

## Reproduction (browser, dexie storage, no plugins required)

```js
import { createRxDatabase, removeRxDatabase } from 'rxdb';
import { getRxStorageDexie } from 'rxdb/plugins/storage-dexie';

const schema = {
    version: 0, primaryKey: 'id', type: 'object',
    properties: { id: { type: 'string', maxLength: 100 } }, required: ['id']
};
const open = async (name) => {
    const db = await createRxDatabase({ name, storage: getRxStorageDexie(), multiInstance: true });
    await db.addCollections({ docs: { schema } });
    return db;
};

const name = 'repro' + Date.now();
let db = await open(name);
await db.docs.insert({ id: 'a' });
await db.close();                                   // leaks the Dexie connections
await removeRxDatabase(name, getRxStorageDexie());
for (const d of await indexedDB.databases()) {      // app-level cleanup of the per-collection DBs
    if (d.name?.includes(name)) {
        await new Promise(res => {
            const req = indexedDB.deleteDatabase(d.name);
            req.onsuccess = req.onerror = () => res();
            // logs "Another connection wants to delete database ..." for every sub-DB,
            // which shows that close() did not release the connections
        });
    }
}
db = await open(name);                              // throws DatabaseClosedError
```

## Runtime trace of the counter

To verify the analysis at runtime, I instrumented `Map.prototype.set`/`delete` in
the reproduction and logged every access whose call stack originates in the dexie
helper (function names containing `DexieDb`).

<details>
<summary>Trace with 17.3.0 as published (reproduction fails with DatabaseClosedError)</summary>

```text
set(key="rxdb-dexie-repro…--0--_rxdb_internal", value=undefined)  in getDexieDbWithTables
set(key=undefined, value=0)                                       in getDexieDbWithTables
set(key="rxdb-dexie-repro…--0--_rxdb_internal", value=<Promise>)  in getDexieDbWithTables
set(key="rxdb-dexie-repro…--0--docs", value=undefined)            in getDexieDbWithTables
set(key=undefined, value=0)                                       in getDexieDbWithTables
set(key="rxdb-dexie-repro…--0--docs", value=<Promise>)            in getDexieDbWithTables
set(key=<Promise>, value=NaN)                                     in closeDexieDb
set(key=<Promise>, value=NaN)                                     in closeDexieDb
set(key=<Promise>, value=NaN)                                     in closeDexieDb
set(key=<Promise>, value=NaN)                                     in closeDexieDb
set(key=<Promise>, value=NaN)                                     in closeDexieDb
```

The initial count is stored under the key `undefined` (defect 2). Every
`closeDexieDb()` call writes `NaN` (`undefined - 1`), meaning the else-branch ran
and `dexieDb.close()` was never reached. No increment and no `delete` appear
anywhere in the trace: connections and cache entries are never released.

</details>

<details>
<summary>Same trace with the suggested fix applied (reproduction passes)</summary>

```text
set(key=<Promise>, value=0)                                       in getDexieDbWithTables
set(key="rxdb-dexie-repro…--0--_rxdb_internal", value=<Promise>)  in getDexieDbWithTables
set(key=<Promise>, value=1)                                       in getDexieDbWithTables
set(key=<Promise>, value=0)                                       in getDexieDbWithTables
set(key="rxdb-dexie-repro…--0--docs", value=<Promise>)            in getDexieDbWithTables
set(key=<Promise>, value=1)                                       in getDexieDbWithTables
delete(key=<Promise>)                                             in closeDexieDb
delete(key="rxdb-dexie-repro…--0--docs")                          in closeDexieDb
delete(key=<Promise>)                                             in closeDexieDb
delete(key="rxdb-dexie-repro…--0--_rxdb_internal")                in closeDexieDb
… (same balanced cycle repeats for removeRxDatabase and for the reopen)
```

Counts go 0 → 1 per storage instance, and on close both the counter and the
name-cache entry are removed. The "Another connection wants to delete" warnings
disappear as well, since the connections are now released before the delete.

</details>

## Bug checklist

- Different browsers: reproduced in both Firefox and Chromium.
- Different storages: not applicable - the defect is located in
  `dexie-helper.ts`, which only the Dexie storage executes.
- `eventReduce: false`: still fails with the same error.
- Key-compression: not used (default off).
- `multiInstance: false`: still fails with the same error.

## Suggested fix

- Key the initial count on `value` (not the still-unassigned `state`).
- Increment the count on every `getDexieDbWithTables()` call. It is called exactly
  once per storage instance (`createDexieStorageInstance`), and `closeDexieDb` runs
  exactly once per instance close, so the counting stays balanced.
- When the last reference is released, also delete the `DEXIE_STATE_DB_BY_NAME`
  entry — otherwise the next open of the same name still receives the closed
  instance.

I have this change running as a patch-package patch in our app, and with it the
scenario above works. I can follow up with a PR containing the fix and a regression
test based on `test/unit/bug-report.test.ts`.

Cr4ftingMine:
Fixes #8793, as requested there, with a regression test based on `test/unit/bug-report.test.ts`.

The PR is structured in two commits so the reproduction stays verifiable:
1. The first commit contains only the test. On that commit the test fails with Dexie's `DatabaseClosedError` ("Database has been closed").
2. The second commit contains the fix (plus changelog entry). From here on the test passes.

## Test case

Reproduces the scenario from the issue: create a database with the dexie storage, close it, delete the underlying per-collection IndexedDB databases (like an application does on logout to remove user data from a shared device), then re-create the same database name in the same JS context. The test runs for the dexie storage only and returns early for other storages, since the defect is located in `dexie-helper.ts`.

## Fix

Repairs the reference counting in `dexie-helper.ts`:

- key the initial count on `value` instead of the still-unassigned `state` (the factory callback runs synchronously inside the assignment, so the count was stored under `undefined` and `closeDexieDb()` computed `undefined - 1 = NaN`),
- count one reference per storage instance in `getDexieDbWithTables()` (there was no increment anywhere, so the close branch was unreachable),
- when the last reference is released, also evict the `DEXIE_STATE_DB_BY_NAME` entry, otherwise the next open of the same name still receives the closed Dexie instance.

## Validation

- new test fails on master with "Database has been closed" and passes with the fix
- `npm run test:fast:dexie`: 1404 passing
- `npm run check-types`: clean
- changelog entry added under `orga/changelog/`


pubkey:
fixed https://github.com/pubkey/rxdb/pull/8907

PR #8842, opened 2026-07-23
docs-src: fire visit_x_urls at 3 urls, add premium_form_focus_10_sec event

pubkey:
## This PR contains:
- SOMETHING ELSE (docs-site tracking-event changes, no library code touched)

Two gaps in the rxdb.info tracking events:

1. `visit_x_urls` only fired on the 5th distinct URL a visitor opens. Most sessions never get that far, so the event was too rare to be useful. It now fires on the 3rd distinct URL (`URL_EVENT_COUNT` 5 -> 3 in `trackUrlChanges()`; the accompanying `visit_<n>_urls` event becomes `visit_3_urls` automatically).

2. There was no signal between "opened a premium package form" and "submitted it". The forms are cross-origin Pipedrive iframes, so typing cannot be observed from the parent page. This PR adds an optional `focusEventType` prop to `IframeFormModal`: when the visitor keeps focus inside the form iframe for 10 seconds (detected via `window` blur with the iframe as `document.activeElement`, cancelled when the window regains focus or the tab is hidden), the event fires once per user. The three premium package dialogs pass `premium_form_focus_10_sec`, so an accidental single click into the form does not count, sustained focus does.

## Todos
- [ ] Tests: not applicable, docs-site tracking only (no `test/` coverage exists for `docs-src` tracking events)
- [ ] Documentation: not applicable, internal tracking behavior
- [ ] Typings: included (`focusEventType?: string` prop)
- [ ] Changelog: not applicable, no library change

https://claude.ai/code/session_01J3tLAPjHjiPNA3dEUgooRd

---
_Generated by [Claude Code](https://claude.ai/code/session_01J3tLAPjHjiPNA3dEUgooRd)_

### DIFF
--- a/docs-src/src/components/modal.tsx
+++ b/docs-src/src/components/modal.tsx
@@ -1,5 +1,5 @@
 // Modal.tsx
-import React, { forwardRef } from 'react';
+import React, { forwardRef, useEffect, useRef } from 'react';
 import type { ModalProps as AntdModalProps } from 'antd';
 import { Modal as AntdModal } from 'antd';
 import { IconClose } from './icons/close';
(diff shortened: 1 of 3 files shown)


pubkey:
CI note: `test-others-replications` failed on `replication-google-drive.test.ts` ("should realtime sync on both sides", a timing assertion where the inserted doc had not reached the second collection yet). This PR only touches docs-src tracking code, and the same job failed on master yesterday (run 47404, 2026-07-22) while the latest master run on this PR's base commit (`f176eda`) is green, so this looks like the known flakiness of that suite rather than something introduced here. If it stays red once the full run completes, a re-run of the failed job should clear it (I can't re-run individual jobs from this session).

---

PR #8860, opened 2026-07-28
Migrate from mocha.parallel to the mocha native --parallel mode

pubkey:
## This PR contains:

- Improved tests (test infrastructure migration, no source code behavior change)

The `mocha.parallel` devDependency is unmaintained since around 2016 while Mocha 11 ships a native `--parallel` mode. This PR removes the dependency and migrates the `NODE_ENV=fast` test runs to the native mode.

What changed:

- The `test:fast:*` node scripts now run the single transpiled files under `test_tmp/unit/` with `mocha --parallel` instead of the serial `unit.test.js` bundle. Native parallel mode parallelizes per file in worker processes, so this replaces the per-suite concurrency that `mocha.parallel` provided.
- The new `test/helper/parallel-setup.ts` is passed to mocha via `--require`:
  - `mochaGlobalSetup` runs once in the main process: it clears the BroadcastChannel folder and starts the fixed-port test servers (previously done by `init.test.ts`), so worker processes do not collide on ports.
  - The root `beforeAll` hook runs per test file in the worker and calls `config.storage.init()` (needed for the foundationdb and mongodb fast runs).
  - It registers the union of plugins that the test files register at module scope. When everything ran as one bundle, all module-scope `addRxPlugin()` calls had already run before the first test started, so files could rely on plugins registered by other files. Registering the same union per worker keeps the behavior equal.
- `init.test.js` is excluded from the parallel run via `--ignore` because its side effects moved to the global fixture. It still runs in the serial bundle used by `test:node:*`, karma, Deno, and Bun.
- `describeParallel` is removed and all suites use plain `describe`. It only differed from `describe` in fast mode, where parallelism now happens at the file level. Test files that imported only `describeParallel` from `config.ts` keep a side-effect import because `config.ts` calls `setConfig()` at module load.
- Removed the `mocha.parallel` devDependency and its `src/types/modules/mocha.parallel.d.ts` module shim.

Verification (all exit 0):

- `test:fast:memory`: 1397 passing, run three times. The serial bundle runs 1401 tests, which is exactly these 1397 plus the 4 excluded `init.test.ts` tests, so no tests were lost. Mocha wall time dropped from 27s to 10s on a 4-core machine.
- `test:fast:dexie`: 1452 passing.
- `test:fast:memory-random-delay`: 1337 passing.
- The serial bundle (`test:node:*` style invocation) still passes with 1401 tests.
- `npm run lint` and `npm run check-types` are clean.

The foundationdb and mongodb fast runs need their services and could not be run locally; they use the same code path plus the per-worker `storage.init()` root hook.

https://claude.ai/code/session_01SskDskRL4FW5p1VsAroB3P

---
_Generated by [Claude Code](https://claude.ai/code/session_01SskDskRL4FW5p1VsAroB3P)_


pubkey:
This warning is expected for this PR. The test-without-fix check verifies that changed tests reproduce a bug, but this PR is a test infrastructure migration (mocha.parallel to the native mocha `--parallel` mode) and contains no source fix. The test files only changed from `describeParallel` to `describe`, so they pass with and without the diff by design.

---

pubkey:
CI note: the `example-tauri` job failed in the webdriver e2e test (`Heroes should insert a hero` finds 0 elements right after the insert click). This PR only changes the unit test infrastructure and a devDependency, nothing the tauri example consumes, and the check is green on recent master pushes. This looks like a timing flake in the example e2e test. I will re-run the failed job when the workflow run completes.

---

PR #8872, opened 2026-07-30
docs: verify and complete project-activity numbers in the alternatives articles

pubkey:
## This PR contains:

Improved docs, plus one small docs component. No changes to `src/`.

The alternatives articles compare RxDB with other projects, but the checkable facts about each project (GitHub stars, npm downloads, how old the last commit is) were only present on some pages, and a few of the maintenance claims were wrong. Readers who verify a number and find it stale or contradicted stop trusting the rest of the page.

All 28 articles were checked against the GitHub API, the npm downloads API and each repository's commit history on July 30, 2026. The existing numbers matched today's values and the 14 last-commit months already claimed were each re-verified as correct, so most pages needed nothing. What changed:

- **couchbase**: had no numbers at all. Added stars (couchbase-lite-core 265, couchbase-lite-ios 1,663 vs RxDB 23,296) and downloads for the React Native binding (`cbl-reactnative` 99 vs `rxdb` 270,494), with an explicit note that Couchbase Lite is installed mostly through its native iOS, Android and .NET SDKs, so npm says nothing about its total install base.
- **minimongo**: added the missing maintenance fact. The newest commit on `master` is from October 2025.
- **nedb**: the page contradicted itself, saying "last commit 2016, archived" in the comparison table and "last commit May 2025" in the body. The repository is not archived, the library code has not changed since May 2016, the README carries a "not maintained anymore" notice from 2021, and the May 2025 commits only edit the README. All three places now say that. Also mentions the maintained `@seald-io/nedb` fork (624,845 downloads in the last 30 days) instead of claiming no fork became widely used.
- **cloudant**: `nodejs-cloudant` was archived in March 2022 and is read-only. Added that, plus the maintained replacement `@ibm-cloud/cloudant`, which is a REST SDK for the hosted service and does not cover the offline case.
- **replicache**: the repository is archived and read-only, development moved to the Zero monorepo.
- **pouchdb**: the repository moved to the Apache org, so the link now points at `apache/pouchdb`.
- **npm figures** across the set refreshed to the 30-day window ending July 29, 2026, so every number is reproducible on npmtrends today.

## The TimeSince component

`docs-src/src/components/time-since.tsx` renders how long ago a date was, and is registered globally so pages can use it without an import:

```
The last commit was in March 2022, TimeSince date="2022-03-28"    renders  4 years ago
```

It picks days below a month, months below two years and years above that, and takes `unit="months"` or `unit="years"` to force one. Months are counted as whole calendar months, so it does not drift the way dividing days by an average month length does. The value is calculated at render time, so the static HTML carries the distance as of the last docs build and the browser corrects it on hydration.

It is used on all 14 last-commit claims in the alternatives articles. The absolute date stays in the sentence, because that is the part a reader can verify. It also replaces the hardcoded "in ten years" in `nedb-alternative.md`, which was the one relative claim on these pages that would have gone wrong on its own.

## Todos

- [x] Documentation
- [ ] Tests (docs prose and one docs-only component; the component was rendered through react-dom/server to check the unit thresholds, the calendar-month boundaries, the forced-unit fallbacks and the two error paths, and the `docs:build` step compiles it in CI)
- [ ] Typings (not applicable)
- [ ] Changelog (not applicable, neither a testcase nor a FIX)


pubkey:
CI status on this branch, in case it looks noisier than it is. Two different jobs went red across three runs, neither one caused by this diff, which touches only `docs-src/`, `scripts/` and two script entries in `package.json`.

**Run 2 (commit 2410386), `example-react` cancelled.** The job hung inside `npm install` for 30 minutes and was killed by its `timeout-minutes: 30`. The log ends with `Terminate orphan process: pid (1729) (npm install)`. Same job passed on the commit before it.

**Run 3 (commit e310b1c), `test-others-replications` failed.** `test/replication-google-drive.test.ts:1184`, the "insert one" step, where `awaitCollectionsHaveEqualState` timed out waiting for the `insert` document to reach collection B. e310b1c is an empty commit, so its tree is byte-identical to 2410386, where this job passed. Same input, different result, which makes it a timing-dependent test rather than a regression.

The jobs that actually cover this change are green on the current head: `test-builds` (which runs `npm run docs:build` and compiles the new components and the converted mdx), `test-code-style`, and the em-dash check inside it.

I cannot re-run individual jobs through this integration, and re-triggering the whole workflow with empty commits is not a good pattern, so I am leaving it here rather than pushing another one. A re-run of `test-others-replications` should be enough.

---

PR #8877, opened 2026-07-30
Update graphql-ws from 5.16.2 to 6.2.0

pubkey:
## This PR contains:

- A BREAKING CHANGE (dependency major bump, raised Node.js requirement)
- IMPROVED DOCS

RxDB pinned `graphql-ws` at `5.16.2`. v5 works, but v6 has been the maintained line since 2025 and v5 no longer gets fixes. Staying on v5 also forces users who run their own `graphql-ws` server to keep two major versions of the same package in their tree.

This PR moves the [GraphQL replication](https://rxdb.info/replication-graphql.html) to `graphql-ws@6.2.0`.

### Changes

- **`package.json`**: `graphql-ws` `5.16.2` → `6.2.0`. Same bump in `examples/graphql/package.json`.
- **`src/plugins/replication-graphql/graphql-websocket.ts`**: import `createClient` from `graphql-ws/client` instead of the package root. v6 splits out a truly zero-dependency client, so the bundle no longer pulls in the `graphql-ws` server code and its `graphql` peer dependency. The package root also resolves through the `import` condition before `browser` in the v6 exports map, so the explicit subpath is the reliable way to get the client-only entry.
- **`src/types/plugins/replication-graphql.d.ts`**: `ClientOptions` now imported from `graphql-ws/client`. `RxGraphQLPullWSOptions` is unchanged: `url`, `shouldRetry` and `webSocketImpl` all still exist on the v6 `ClientOptions`.
- **`test/helper/graphql-server.ts`** and **`examples/graphql/server/index.js`**: `graphql-ws/lib/use/ws` → `graphql-ws/use/ws`, since v6 dropped the `/lib/` part from all subpath imports.
- **`package.json` engines**: `node >=18` → `node >=20`. graphql-ws v6 declares `engines.node >= 20`, so keeping `>=18` would be wrong and would produce `EBADENGINE` warnings on install. Node.js 18 has been end of life since April 2025 and `.nvmrc` here is already at 24.
- **`docs-src/docs/replication-graphql.md`**: the linked `ClientOptions` reference page returned 404. The typedoc output moved to `https://the-guild.dev/graphql/ws/docs/client/interfaces/ClientOptions`.
- **`orga/changelog/graphql-ws-v6.md`**: changelog entry covering the upgrade and the Node.js bump.

### What did not need changing

The v6 breaking changes that affect server hook signatures (`schema`, `context`, `onSubscribe`, `onOperation`, `onError`, `onNext`, `onComplete` now receive `id` and `payload` instead of the full message) do not apply here. The test server and the example server only use `schema` as a value plus `execute`, `subscribe`, `roots` and `onConnect`, all of which kept their v5 signature. The client `subscribe()` sink and `dispose()` are also unchanged.

### Note for users

Anyone running their own `graphql-ws` server has to remove the `/lib/` part from their imports and adapt the server hooks listed above. That is documented in the changelog entry and linked to the [upstream migration notes](https://github.com/enisdenjo/graphql-ws/blob/master/CHANGELOG.md#600).

## Todos

- [x] Tests - the existing GraphQL replication suite covers the websocket path (`live:true with pull.stream$`, `should respect pull.includeWsHeaders`, `should respect pull.wsOptions`, `should be able to use the ws-subscriptions`). 69 passing, 0 failing.
- [x] Documentation - fixed the dead `ClientOptions` link.
- [x] Typings - `npm run check-types` passes.
- [x] Changelog - `orga/changelog/graphql-ws-v6.md`.

### Verification

```
npm run build          # ok
npm run check-types    # ok
npm run lint           # ok
npm run test:fast:memory   # 1397 passing, 0 failing
```

Also verified that both `graphql-ws/client` and `graphql-ws/use/ws` resolve under ESM and CJS, and that the emitted `dist/esm`, `dist/cjs` and `dist/types` output references `graphql-ws/client`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01YLU9vxehHiTbGzQcXdkp7v)_


pubkey:
CI status: 23 of 24 jobs green. The one failure is `storage-sqlite-trial`, and it is not caused by this PR.

**What failed**

`reactive-query.test.ts` → `groups live query write bursts before _ensureEqual when enabled`:

```
AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
757 !== 756
at test/unit/reactive-query.test.ts:189:24
```

**Why it is unrelated to this diff**

This PR only changes the `graphql-ws` version, the import specifiers in the GraphQL replication plugin and the two GraphQL servers, the `engines.node` field, a docs link, and a changelog file. It does not touch `rx-query.ts`, the query cache, the live-query throttling, or any storage.

The same test file passed in this run on `storage-memory`, `storage-dexie`, `storage-localstorage`, `storage-mongodb`, `storage-foundationdb`, and `windows`. Only the slowest storage tripped.

**The race**

The test sets `liveQueryUpdateThrottleTime: 100`, which becomes `auditTime(100)` on the change-event stream in `rx-query.ts`. The audit window opens when the *first* insert's change event arrives, not when `Promise.all` resolves:

```ts
await Promise.all(
    new Array(5).fill(0).map((_v, index) => c.insert(...))
);
await promiseWait(5);
assert.strictEqual(query._lastEnsureEqual, ensureEqualAfterInitialRun);
```

When the five inserts plus the 5ms wait take longer than 100ms of wall clock, the audit window expires while the later inserts are still running, `_ensureEqual` legitimately fires once, and the counter advances by exactly one. That matches the observed 756 → 757. On a loaded runner with the SQLite storage that is easy to hit; on the in-memory storages the inserts finish well inside the window.

Run in isolation under `DEFAULT_STORAGE=sqlite-trial` the test passed 8 out of 8 times locally, which is consistent with a load-dependent race rather than a regression.

**Why I am not fixing it here**

The fix belongs in the test, not in this PR: the assertion needs a throttle window that cannot be outrun by the writes (for example a much larger `liveQueryUpdateThrottleTime`, which costs no extra runtime since the test only waits 5ms afterwards). Bundling an unrelated test-stabilization change into a dependency upgrade would muddy both. Happy to do it here or in a separate PR, whichever you prefer.

---

PR #8886, opened 2026-07-31
docs: shorten the sitewide title suffix to "| RxDB"

pubkey:
## This PR contains:

- IMPROVED DOCS (SEO / page titles)

Every page title on rxdb.info ends with `| RxDB - JavaScript Database`, a 29-character suffix. Because the pages under `src/pages` *also* concatenated the site title into their own title, most of them rendered the brand three times. Verified live on 2026-07-30:

```
Legal Notice - RxDB - JavaScript Database | RxDB - JavaScript Database
RxDB for Professionals - RxDB - JavaScript Database | RxDB - JavaScript Database
RxDB User Survey - RxDB - JavaScript Database | RxDB - JavaScript Database
```

Two reasons this costs us clicks:

1. **Truncation.** Google cuts desktop titles around 60 characters. Nine of the ten highest-impression article titles are longer than that; four are over 90.
2. **Boilerplate rewriting.** Google's [title link documentation](https://developers.google.com/search/docs/appearance/title-link) names repeated boilerplate as a reason it replaces your title with one of its own choosing. A rewritten title is a search listing we no longer control.

The motivation was an analysis of our Search Console export: 69 non-brand queries where we rank at position 10 or better but click through at under 75% of our own average CTR for that position, together about 4,200 clicks below baseline.

## What changed

- **`docusaurus.config.ts`**: `title` from `'RxDB - JavaScript Database'` to `'RxDB'`. This is the only global lever for the suffix. Docusaurus builds it from `siteConfig.title` in `useTitleFormatter` (`@docusaurus/theme-common/src/utils/titleFormatterUtils.tsx`), so there is no separate suffix setting.
- **16 pages under `src/pages`**: dropped the `- ${siteConfig.title}` concatenation, since the global suffix already supplies the brand. Removed the `useDocusaurusContext` imports that became unused.
- **`constants.ts`**: added `HOME_TITLE`, the descriptive title for the two pages that stand alone in search results with no parent topic for context (the homepage and `/consulting/`).
- **`theme/ThemeProvider/TitleFormatter/`** (new): skips the suffix when the page title is `HOME_TITLE`, mirroring the guard Docusaurus already applies when the page title equals the site title. Without it the homepage would render `RxDB - JavaScript Database | RxDB`.
- **`theme/PageMetadata/`**: corrected a comment. It claimed the swizzle strips the site-title suffix; it does not, because docs and plugin pages import `PageMetadata` from `@docusaurus/theme-common` directly and never reach the swizzled component. The live titles confirm the suffix is still applied everywhere.

### Resulting titles

| | Before | After |
| --- | --- | --- |
| `crdt.html` | `CRDT - Conflict-free replicated data type Database \| RxDB - JavaScript Database` (79) | `CRDT - Conflict-free replicated data type Database \| RxDB` (57) |
| `/legal-notice/` | `Legal Notice - RxDB - JavaScript Database \| RxDB - JavaScript Database` (70) | `Legal Notice \| RxDB` (19) |
| `/premium/` | `RxDB for Professionals - RxDB - JavaScript Database \| RxDB - JavaScript Database` (80) | `RxDB for Professionals \| RxDB` (29) |
| homepage | `RxDB - JavaScript Database` | unchanged |

The homepage is deliberately untouched: it is the highest-impression page on the site and its title is already only 26 characters.

## Notes for review

- **`ThemeProvider/TitleFormatter` is an unlisted swizzle target.** It is a real theme component shipped by `theme-classic`, but it does not appear in `docusaurus swizzle --list`, so a Docusaurus upgrade could move it. The file documents the symptom (homepage rendering `RxDB - JavaScript Database | RxDB`) and the fallback: delete the override and give the homepage a title that reads correctly with the suffix appended, e.g. `Local-First JavaScript Database`. Happy to take that simpler route instead if you would rather not carry the swizzle.
- **`src/pages/release.tsx` had `title={'Code - ...'}`**, which looks like a copy-paste from `code.tsx`. I preserved the label as `Code` rather than renaming it, since that is a content decision. Worth a separate fix.

## Todos

- [ ] Tests — none added; this is a docs/config change with no runtime behavior to assert on, and the repo has no title-rendering test to extend.
- [x] Documentation — the reasoning is documented inline in `docusaurus.config.ts`, `constants.ts`, and the new formatter.
- [ ] Typings — no typing changes.
- [ ] Changelog — not user-facing for the library.

## Verification

- `npx tsc --noEmit` over `docs-src`: **482 errors before, 482 after**, identical set with only line numbers shifted. No new type errors. (The pre-existing errors come from running `tsc` without the generated Docusaurus types and an unbuilt root package.)
- A full `docusaurus build` was not run here: it needs the root `rxdb` package built, which this environment did not have. CI will cover it.

https://claude.ai/code/session_01WdFsKbki2tz7pjYUL8xBzA

---
_Generated by [Claude Code](https://claude.ai/code/session_01WdFsKbki2tz7pjYUL8xBzA)_

### DIFF
--- a/docs-src/docusaurus.config.ts
+++ b/docs-src/docusaurus.config.ts
@@ -33,7 +33,18 @@ const rehypePrettyCodeOptions: RehypePrettyCodeOptions = {
 
 /** @type {import('@docusaurus/types').Config} */
 const config: Config = {
-    title: 'RxDB - JavaScript Database',
+    /**
+     * Docusaurus appends this to every page title as `<page title> | <title>`
+     * (see useTitleFormatter in @docusaurus/theme-common), so it doubles as the
+     * sitewide title suffix. Keep it short: Google truncates titles around 60
+     * characters and treats repeated boilerplate as a reason to rewrite the
+     * title link, which loses us control of the search listing.
+     * @link https://developers.google.com/search/docs/appearance/title-link
+     *
+     * Pages that need the longer descriptive form set it explicitly (the
+     * homepage and /consulting/ pass HOME_TITLE below).
+     */
+    title: 'RxDB',
     tagline: 'Realtime JavaScript Database',
     favicon: '/files/logo/logo.svg',
     headTags: [
--- a/docs-src/src/constants.ts
+++ b/docs-src/src/constants.ts
@@ -1,6 +1,14 @@
 export const NON_PREMIUM_COLLECTION_LIMIT = 13;
 
 /**
+ * Descriptive page title for the two pages that stand alone in search results
+ * and have no parent topic to give them context: the homepage and /consulting/.
+ * Everywhere else the page sets its own topical title and Docusaurus appends
+ * ` | RxDB` from the `title` in docusaurus.config.ts.
+ */
+export const HOME_TITLE = 'RxDB - JavaScript Database';
+
+/**
  * Premium tier prices in dollars per month, billed annually.
  * Shown on the premium page and used to calculate
  * the lead value on the premium-submitted pages.
--- a/docs-src/src/pages/chat-rules.tsx
+++ b/docs-src/src/pages/chat-rules.tsx
@@ -1,16 +1,14 @@
-import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
 import Layout from '@theme/Layout';
 import React, { } from 'react';
 
 
 export default function Chat() {
-    const { siteConfig } = useDocusaurusContext();
 
 
 
     return (
         <Layout
-            title={`Chat - ${siteConfig.title}`}
+            title={'Chat'}
             description="RxDB Community Chat Rules"
         >
             <main>
--- a/docs-src/src/pages/chat.tsx
+++ b/docs-src/src/pages/chat.tsx
@@ -1,10 +1,8 @@
-import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
 import Layout from '@theme/Layout';
 import React, { useEffect } from 'react';
 import { triggerTrackingEvent } from '../components/trigger-event';
 
 export default function Chat() {
-    const { siteConfig } = useDocusaurusContext();
 
     useEffect(() => {
         triggerTrackingEvent('join_chat', 0.40);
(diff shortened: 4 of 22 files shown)


pubkey:
`test-others-replications` is red on this PR. It is unrelated to the change, and I am not fixing it here.

**The failure** is `replication-google-drive.test.ts` → "live replication should realtime sync on both sides", at [`test/replication-google-drive.test.ts:1179`](https://github.com/pubkey/rxdb/blob/master/test/replication-google-drive.test.ts#L1179). `colB`'s own initial document was missing from the compared state:

```
+ actual - expected
  [
    { age: 66, id: 'a-init',  name: 'colA init', updatedAt: 1785493951643 },
-   { age: 49, id: 'b-init',  name: 'colB init', updatedAt: 1785493951647 }
  ]
```

**Why it is not this PR.** Every file in this branch is under `docs-src/` (22 files: the Docusaurus config, `src/pages`, and two theme components). This job builds and tests `src/`, which the diff does not touch.

**Why it failed.** That assertion is the only convergence check in the test with a shortened timeout:

```js
await awaitCollectionsHaveEqualState(collectionA, collectionB, 'initial', 1000);
```

`awaitCollectionsHaveEqualState` defaults to `timeout = 8000` and every other call in the same test uses that default (`'insert'`, `'delete'`, and the rest). The `'initial'` check gets 1/8th of the budget, and it is the step that has to wait for two peers to finish converging over WebRTC after `awaitInitialReplication()` on both sides. On a slow runner one second is not enough, and the symptom is exactly what we see: one side's own initial doc has not propagated yet.

So the likely fix is dropping the `1000` and letting it use the default, or raising it. That is a one-line change to an unrelated test, so it belongs in its own PR rather than buried in a titles change. Happy to open it if you want.

I could not re-run the job to demonstrate the flake (the integration lacks `rerun-failed-jobs` permission). The most recent master run, `9915cf9c`, which is this branch's base, was green, so this is not a standing master failure.

---

nickchomey:
This is a highly welcomed change. It will reduce the clutter in the open tabs.

PR #8899, opened 2026-08-03
Fix the 482 TypeScript errors of the docs

pubkey:
## This PR contains:

- IMPROVED typings
- A BUGFIX

Running `tsc --noEmit` inside `docs-src/` reported 482 errors, so the docs code was effectively not typechecked. `npm run check-types` at the root does not cover it because the root `tsconfig.json` excludes `docs-src`.

The errors came from three groups:

**1. `allowImportingTsExtensions` was missing (282 errors)**

`docs-src/src/components/error-messages.tsx` and `default-compressible-types.tsx` import from `../../../src/`, which pulls the whole RxDB core into the docs program. The core uses explicit `.ts` file endings in its imports, so every one of them raised `TS5097`. The docs `tsconfig.json` now sets `allowImportingTsExtensions`, same as the root `tsconfig.json`.

**2. The `@theme/*` module declarations were not referenced (about 60 errors)**

Docusaurus ships the ambient `declare module '@theme/...'` declarations inside the theme packages. They are not picked up automatically, so every swizzled component in `docs-src/src/theme/` failed with `TS2307`. Added `docs-src/src/docusaurus-theme.d.ts` with the triple-slash references.

**3. Real type errors in the remaining files**

- `components/video-button.tsx` imported `IconProps` from `./arrow-right`, the file is at `./icons/arrow-right`.
- `components/docs-footer.tsx` used `WrapperProps` without importing it. With the import in place, the accesses to `props.children.type.frontMatter` needed a type for the compiled MDX content component.
- `components/modal.tsx` wrapped AntD's `Modal` in `forwardRef` and passed the ref down. AntD's `Modal` is a `React.FC<ModalProps>` and never accepted a ref, so the ref was a no-op. It is a plain function component now.
- `components/video-box.tsx` passed `onClose` to that modal, which AntD does not support. `onCancel` and `onOk` already handle closing.
- `components/core-concept-section.tsx` passed `language='js'` to the swizzled `CodeBlock`, which ignores it.
- `components/price-tag.tsx` and `components/steps.tsx` needed `React.CSSProperties` / `ReactElement` annotations (React 19 removed the global `JSX` namespace).
- `theme/CodeBlock/index.tsx` imported the `WordWrap` type from `@docusaurus/theme-common/internal`, which does not export it. It is derived from `useCodeWordWrap` now.
- `theme/MDXComponents/index.tsx` imported the named export `Pre` through `@theme/CodeBlock`, whose ambient declaration only has a default export. It now imports from `@site/src/theme/CodeBlock`, the same file.
- `theme/PageMetadata/index.tsx` imported `Props` from `@theme/PageMetadata`, which does not exist in Docusaurus 3.9. The props come from `@docusaurus/theme-common` now.
- `theme/SearchBar/index.tsx` hit the mismatch between the normalized `AskAiConfig` of Docusaurus and the `DocSearchAskAi` type of DocSearch v4. `askAi` is typed from the Docusaurus theme config and cast back where the modal is rendered.
- `theme/Navbar/MobileSidebar/Layout/index.tsx` and `theme/DocSidebarItem/Category/index.tsx` needed explicit annotations.

**4. Six errors in `src/`**

These only showed up under the docs tsconfig, which is less strict than the root one, so the narrowing behaves differently. All fixes are behavior-neutral except one:

- `plugins/backup/file-util.ts` used the deprecated `fs.rmdirSync(path, { recursive: true })`, now `fs.rmSync()`.
- `plugins/utils/utils-object-dot-prop.ts` had `stringifyPath(pathSegments: never[])` and an untyped `currentPath = []` default.
- `plugins/utils/utils-object.ts`, `rx-collection.ts` and `rx-query.ts` needed explicit narrowing.

## Verification

- `tsc --noEmit` in `docs-src/`: 0 errors (was 482)
- `npm run check-types`: passes
- `npm run lint`: passes
- `npm run docs:build`: succeeds
- `npm run test:fast:memory`: 548 passing. The one failure is `replication-webrtc.test.ts` failing to load the `node_datachannel` native binary, which is an environment issue unrelated to this change.

## Todos
- [ ] Tests
- [x] Documentation
- [x] Typings
- [x] Changelog

https://claude.ai/code/session_01XLsy9Zt3ZLupoHcqHYKV6v

---
_Generated by [Claude Code](https://claude.ai/code/session_01XLsy9Zt3ZLupoHcqHYKV6v)_


pubkey:
`example-tauri` is red, but it fails on `master` too, so it does not come from this PR.

The last completed `master` run [30799404998](https://github.com/pubkey/rxdb/actions/runs/30799404998) has `example-tauri` as its only failing job, with the same assertion:

```
1) Heroes should insert a hero
   Expected length: 1
   Received length: 0
   at Context.<anonymous> (examples/tauri/test/specs/example.e2e.js:19:22)
```

The test inserts a hero and then waits a fixed `await browser.pause(100)` before it asserts that the row is rendered, so it depends on the insert and the re-render finishing within 100ms inside the Tauri webview on the runner. That is a timing problem in the example test, not something this PR touches: the changes in `src/` here are `fs.rmdirSync` to `fs.rmSync` in the backup plugin, plus type annotations and narrowing rewrites in `utils-object-dot-prop.ts`, `utils-object.ts`, `rx-collection.ts` and `rx-query.ts` that compile to the same JavaScript.

Leaving it out of scope here. Happy to send a separate PR that replaces the fixed pause with a `browser.waitUntil` poll if you want that fixed.

---

Issue #8893, opened 2026-08-02
Closing a database with live replication can cause an unhandled `BroadcastChannel` error


vmanot:
RxDB 17.4.0. On Bun 1.3.14, this minimal example fails when `db.close()` runs:

```ts
import { createRxDatabase } from 'rxdb';
import { replicateRxCollection } from 'rxdb/plugins/replication';
import { getRxStorageMemory } from 'rxdb/plugins/storage-memory';

const schema = {
    version: 0,
    primaryKey: 'id',
    type: 'object',
    properties: { id: { type: 'string', maxLength: 40 } },
    required: ['id']
} as const;

const db = await createRxDatabase({
    name: 'repro-' + Date.now(),
    storage: getRxStorageMemory(),
    multiInstance: true
});
await db.addCollections({ items: { schema } });

replicateRxCollection({
    collection: db.items,
    replicationIdentifier: 'repro',
    live: true,
    autoStart: true,
    pull: {
        handler: async () => ({ documents: [], checkpoint: undefined })
    }
});

await db.close();
await new Promise(resolve => setTimeout(resolve, 500));
```

Run with `bun --unhandled-rejections=strict`. It fails with:

```text
InvalidStateError: This BroadcastChannel is closed
```

Bun exits with code 101. I have not been able to reproduce it with Node 26.5.1.

I do not see the error when `multiInstance` is false or when I await `db.waitForLeadership()` before closing.

`db.close()` should close cleanly. My guess is that the close hook calls `die()` without returning its promise, so the close may finish too early. Should it wait for leader election to finish?

pubkey:
fixed here, please check https://github.com/pubkey/rxdb/pull/8901

PR #8901, opened 2026-08-03
Fix close() resolving before the leader election is finished (#8893)

pubkey:
## This PR contains:

A BUGFIX, IMPROVED TESTS, and a dependency update.

[#8893](https://github.com/pubkey/rxdb/issues/8893): closing an `RxDatabase` while a leader election is still running throws an unhandled `InvalidStateError: This BroadcastChannel is closed`. The reporter hit it on Bun with a live replication.

The issue had two halves.

### Half one, fixed upstream

https://github.com/pubkey/broadcast-channel/pull/1416 stops a dead elector from becoming leader and from posting on a closed channel, and makes `die()` idempotent. This PR bumps the dependency, which was pinned exactly at `7.3.0`, so without the bump an install of RxDB keeps resolving the broken version.

Measured with the reporter's example under `bun --unhandled-rejections=strict`:

| | 7.3.0 | 7.4.0 |
| --- | --- | --- |
| exit code | 101 | 0 |
| `InvalidStateError` | 3 | 0 |
| clean runs | 0/3 | 3/3 |

### Half two, fixed here

`await db.close()` resolved while the broadcast channel was still open, because two promises were dropped in `src/plugins/leader-election/index.ts`:

- `onClose()` called `has.die()` without returning it, so the `preCloseRxDatabase` hook did not wait for the election to stop.
- the `close()` wrapper called `removeBroadcastChannelReference()` without awaiting the `bc.close()` it returns.

That is what let the election outlive the database and post on a channel that `close()` was supposed to have closed. Both are now awaited.

## The testcase

`#8893 close() must not resolve before the broadcast channel is closed` patches `method.close` to record when the channel really closes, then asserts it has closed by the time `close()` resolves. It is a state check with no timing dependency, so it behaves the same on every runtime and channel method.

It was pushed failing first, on `ef5337f`, where it failed on five CI jobs (`storage-memory`, `storage-dexie`, `storage-remote`, `storage-sqlite-trial`, `windows`), each `1388 passing / 1 failing`. The fix in this PR turns it green.

Verified locally on the final commit:

```
npm run test:fast:memory        1401 passing, 0 failing
npm run test:bun:memory         exit 0 in 25s
leader-election, node+simulate  15 passing
leader-election, bun+native     15 passing
npm run lint                    clean
npm run check-types             clean
```

The `test:bun:memory` timing is worth noting: while the testcase was failing, that job hung for the full 30 minutes and was cancelled every run, because `mocha --bail` without `--exit` does not exit under Bun. With the fix it completes in 25 seconds. That is a pre-existing property of the script rather than something this PR changes, so any future failing test in that job will hang it the same way.

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Typings
- [x] Changelog


pubkey:
CI is red on purpose. This PR only adds the reproducing testcase for #8893, it does not contain the fix, so I am not going to push anything to turn the new test green.

The `windows` job shows the failure is isolated to the new test and nothing else regressed:

```
1388 passing (1m)
1 failing

1) leader-election.test.js
     cleanup
       #8893 close() must await the running leader election:
    AssertionError: close() resolved while the broadcast channel was still open
```

`test-without-fix`, `storage-localstorage` and `example-angular` are green. The remaining jobs that run the unit suite against a multi-instance storage will fail on the same single assertion.

On what a fix needs: awaiting `die()` and `removeBroadcastChannelReference()` in `src/plugins/leader-election/index.ts` satisfies the first assertion, but the reported `InvalidStateError` still happens. The in-flight `applyRun()` inside `broadcast-channel` does not re-check `isDead` after its `sleep(responseTime)`, so it still calls `beLeader()` and posts `tell` on the closed channel. Guarding that call site makes both assertions pass and the reported example run clean, so the complete fix looks like an RxDB change plus a `broadcast-channel` change.

---

pubkey:
All 25 checks finished. Five failures, all the same single intended assertion: `windows`, `storage-memory`, `storage-dexie`, `storage-remote`, `storage-sqlite-trial`, each reporting ~1388 passing / 1 failing on `#8893 close() must await the running leader election`. Nothing else regressed.

One separate finding worth recording, unrelated to what this PR changes.

`test-others-runtimes` did not fail, it was **cancelled after 28 minutes**, with orphan `npm run test:bun:memory` and `bun` processes left behind. The mocha run itself had already finished at 22s with `1274 passing, 1 failing`. The process then never exited.

This is not caused by anything the new testcase leaves behind. I isolated it locally against the full `test_tmp/unit.test.js`:

| run | result |
| --- | --- |
| new test skipped | exits, 23s |
| new test present, **failing** | hangs, killed at 5min |
| identical test body, made to **pass** | exits, 23s |

Same work in the last two rows, the only difference is whether the assertion throws. So it is the `--bail` abort path under Bun that does not exit:

```
"test:bun:memory": "... mocha test_tmp/unit.test.js --bail --timeout 60000"
```

There is no `--exit`, so mocha waits for the event loop to drain after bailing, and under Bun it never does. Any failing test in that job will burn the full job timeout instead of failing fast, so this will show up again the next time an unrelated test breaks on Bun.

I have not touched the CI scripts. Adding `--exit` would fix the symptom but would also stop mocha from surfacing genuine dangling-handle leaks by hanging, which may be why it is absent, so that is your call.

Pushed `f9836e1` in the meantime: the testcase now restores the patched `method.close` and kills the elector in a `finally`. That object is shared by every broadcast channel in the process, so the patch previously leaked into the rest of the suite whenever the assertions failed. It does not change what the test reproduces.

---

pubkey:
Correcting the record on this branch: the last check-in caught that CI went **fully green** on `2325dca`, which for a reproduction PR means the testcase had stopped reproducing the bug.

Cause was my own instrumentation. `2325dca` wrapped `method.postMessage`, and the promise it returns goes into the `_uMP` set that `BroadcastChannel.close()` awaits:

```js
var sendPromise = broadcastChannel.method.postMessage(broadcastChannel._state, msgObj);
broadcastChannel._uMP.add(sendPromise);
```

So the wrapper delayed exactly the close-versus-election ordering the testcase measures, and both assertions started passing on the CI runners. It still failed locally, which is the signature of a race I had made timing-dependent rather than deterministic.

`ef5337f` reverts the testcase to the `400071a` version, which only patches `method.close` and never feeds anything into `_uMP`. That version was the one actually verified as deterministic: it failed on five CI jobs and on both Node.js and Bun. The three commits in between (`f9836e1`, `351b548`, `2325dca`) were attempted cleanups that each made it worse, and none of them were needed.

Local verification of the reverted testcase, two runs of the parallel `test:fast:memory` command plus Bun:

```
test:fast:memory (parallel)  run1 exit=1  269 passing  1 failing
test:fast:memory (parallel)  run2 exit=1  294 passing  1 failing
bun + native BroadcastChannel        14 passing  1 failing
```

One discrepancy I could not resolve, worth knowing if you touch this testcase. Locally `test:node:memory` exits 5 on unhandled rejections from the still-running election, while CI at that same commit reported a clean `1388 passing, 1 failing`. The reason is that the two environments do not pick the same channel method: locally `broadcast-channel` selects the `node` method, while the CI stack traces from `f9836e1` and `351b548` clearly came from `methods/native.js`. So my local runs are not a faithful model of those jobs, and I am treating CI as the arbiter here rather than my machine.

The `test-others-runtimes` cancellation is unchanged and still the pre-existing `--bail`-without-`--exit` behavior described above.

---

pubkey:
replaced by https://github.com/pubkey/broadcast-channel/pull/1416

PR #8919, opened 2026-08-05
Upgrade Angular examples and dev dependency to Angular 22

pubkey:
## This PR contains:

- SOMETHING ELSE (dependency upgrade)
- A BUGFIX (flaky Tauri example e2e testcase)

The [Angular example](https://github.com/pubkey/rxdb/tree/master/examples/angular), the [Ionic example](https://github.com/pubkey/rxdb/tree/master/examples/ionic) and the root `@angular/core` devDependency that the `reactivity-angular` plugin is built and type checked against were still on Angular 21. This PR moves them to Angular 22.

### What changed

- All `@angular/*` packages in both examples go from `21.2.x` to `22.1.x` (core packages `22.1.0`, CDK and Material `22.1.1`, CLI and build tooling `22.1.3`).
- The root `@angular/core` devDependency goes from `21.2.18` to `22.1.0`. The `@angular/core` peer dependency stays `*` and optional, so RxDB keeps working with older Angular versions.
- Both examples move from TypeScript `5.9.3` to `6.0.3` because `@angular/build` v22 declares a `typescript: "&gt;=6.0 &lt;6.1"` peer dependency.
- TypeScript 6 errors on the deprecated `baseUrl` compiler option (`TS5101`), so the Angular example drops it from `tsconfig.json`. Two files relied on it for root-absolute imports and now use relative paths:
  - `src/app/components/hero-edit/hero-edit.component.ts`: `src/app/RxDB` becomes `../../RxDB`
  - `src/app/services/replication.ts`: `src/environments/environment-type` becomes `../../environments/environment-type`

No application code changed beyond those two import paths.

`.nvmrc` already pins Node `24.18.1`, which satisfies the Angular CLI v22 engine requirement (`^22.22.3 || ^24.15.0 || &gt;=26.0.0`), so no CI Node version change was needed.

### Also in this PR: FIX the flaky Tauri example e2e testcase

`example-tauri` failed on this branch while every other job passed. It is not caused by the Angular upgrade, and that is verifiable: the Tauri example consumes RxDB only through `rxdb: "file:rxdb-local.tgz"`, and building plus packing that tarball at base `53a634d` and at this branch produces **1999 identical files** (all `1371` files under `dist/` byte-identical by SHA-256), differing in exactly one line:

```
595c595
<     "@angular/core": "21.2.18",
---
>     "@angular/core": "22.1.0",
```

That is the `devDependencies` entry, which npm does not install for a dependency, and the Tauri app never imports `@angular/core`.

The real cause is a race in the testcase itself. It clicked submit, waited a fixed `browser.pause(100)`, then asserted the hero was rendered. An insert through the SQLite trial storage over the Tauri IPC bridge plus the re-render of the query subscription can exceed 100ms on a loaded runner, producing `Expected length: 1, Received length: 0`. `example-angular` and `example-ionic` absorb this class of race by retrying; `example-tauri` has no retry. It now waits for the element instead:

```js
const insertedHero = await $('.name[name="Iron Man"]');
await insertedHero.waitForExist({ timeout: 20000 });
```

### Verification

Run locally on Node 24.18.1:

- Root `npm run build`, `npm run check-types`, `npm run lint`: pass
- Angular example: `npm run lint`, `npm run build` (production), `ng build --configuration ssr-prod` (prerender succeeded and the rendered `index.html` contains the `heroes-list` output), `ng build --configuration capacitor`: all pass
- Angular example `npm run test:build` (testcafe e2e in Chrome 141): 3 passed
- Angular example `npm run test:ssr`: 3 passing
- Ionic example: `npm run build` and `npm run test` (testcafe e2e): 1 passed
- Tauri example `npm run test` (full `tauri build --no-bundle` plus wdio against `tauri-driver` 2.0.6 under xvfb): 2 passing

Note on the SSR run: Angular logs an `NG0205` (injector already destroyed) during server side rendering, coming from the module-level database singleton holding an `Injector` from an earlier request. This is pre-existing and unrelated to the upgrade. I reinstalled the example at Angular 21 and ran the same test to confirm: 4 `NG0205` occurrences on Angular 21 and 4 on Angular 22, with the test suite passing in both cases. Not addressed here.

Angular 22 marks `@angular/animations` and `@angular/platform-browser-dynamic` as deprecated. Both are listed in the Angular example's dependencies but neither is imported anywhere, so they were left in place rather than removed as part of this upgrade.

## Todos

- [x] Tests
- [ ] Documentation
- [ ] Typings
- [x] Changelog


pubkey:
CI status for `6338057`: everything green except `example-tauri`, which I am not fixing here because the failure is unrelated to this PR.

The jobs that actually cover this change both passed:

- `example-angular`: success (build, lint, testcafe e2e, SSR test, Capacitor build)
- `example-ionic`: success (build, testcafe e2e)

`example-tauri` failed on `Heroes should insert a hero`:

```
Expected length: 1
Received length: 0
Received array:  []
    at Context.<anonymous> (examples/tauri/test/specs/example.e2e.js:19:22)
```

Why this is not caused by the upgrade:

- The diff only touches `examples/angular`, `examples/ionic`, the root `@angular/core` devDependency, and a changelog file. `examples/tauri/package.json` has no Angular dependency at all.
- The root change is a devDependency bump. devDependencies of the packed `rxdb-local.tgz` are not installed by the examples, so it cannot reach the Tauri runtime.
- The preceding `master` run ([30964565473](https://github.com/pubkey/rxdb/actions/runs/30964565473)) was fully green, so this is not a pre-existing base branch breakage either.

The test looks racy rather than broken: it clicks submit, waits a fixed `await browser.pause(100)`, and then asserts the hero is rendered. On a loaded runner the insert plus re-render can take longer than 100ms. The `example-angular` and `example-ionic` jobs already guard against this class of flake by retrying (`npm run test:build || npm run test:build || npm run test:build`); `example-tauri` has no retry and no wait-for-element.

I could not re-run the job myself (the API returns `403 Resource not accessible by integration` for `rerun-failed-jobs`), so a re-run has to be triggered from the Actions UI. Happy to open a separate PR replacing that fixed `pause(100)` with a `waitUntil` on the element if you want the flake fixed properly, but I left it out of this one to keep the upgrade diff focused.

---

pubkey:
Correcting my earlier comment: I called `example-tauri` flaky off a single failure, then it failed a second time on re-run, which is not what a one-off flake looks like. So I stopped guessing and tested it. The conclusion holds, but now there is evidence behind it.

**1. It is not registry drift.** There is no committed lockfile (`package-lock.json` is gitignored) and the `example-tauri` job has no dependency cache, so every run resolves fresh. That made "a transitive dep published in the meantime" a real possibility. It is ruled out: PR #8913 (a webpack bump, touching nothing Angular or Tauri) ran `example-tauri` in the same 13:02 window and **passed** ([job 92313771451](https://github.com/pubkey/rxdb/actions/runs/31008336363/job/92313771451)).

**2. It is not this PR either, and this part is provable.** The Tauri example consumes RxDB only through `rxdb: "file:rxdb-local.tgz"`. So I built and packed that tarball twice on the same machine, once at base `53a634d` (with `@angular/core@21.2.18` actually installed) and once at this PR's `6338057` (with `22.1.0` installed), and diffed them:

- `dist/`: all **1371** files byte-identical (identical SHA-256 per file).
- Full tarball: **1999** files, `diff -r` reports exactly one difference:

```
595c595
<     "@angular/core": "21.2.18",
---
>     "@angular/core": "22.1.0",
```

That is the `devDependencies` entry. npm does not install a dependency's devDependencies, and the Tauri app never imports `@angular/core`. The Tauri example therefore receives byte-identical RxDB code on this branch and on master. Nothing in this diff can change its runtime behavior.

That leaves runner timing as the only variable, which fits the test:

```js
await submit.click();
await browser.pause(100);
const elements = await $$('.name[name="Iron Man"]');
expect(elements).toHaveLength(1);
```

A fixed 100ms for an insert through the SQLite trial storage over the Tauri IPC bridge, plus the subscription re-render, is thin. `example-angular` and `example-ionic` absorb this class of race by retrying (`npm run test:build || npm run test:build || npm run test:build`); `example-tauri` has no retry and no wait-for-element, so it fails outright when the runner is slow.

I have left it alone because it is a different example and outside the scope of an Angular upgrade. Re-running the job should clear it, and I still cannot do that myself (`rerun-failed-jobs` returns `403 Resource not accessible by integration`). If you would rather have it fixed properly, the targeted change is to replace the fixed pause with a wait:

```js
await submit.click();
const hero = await $('.name[name="Iron Man"]');
await hero.waitForExist({ timeout: 10000 });
```

Say the word and I will add it here or open it as its own PR.

---

PR #8940, opened 2026-08-10
feat(db-viewer): add the RxDB database viewer plugin

pubkey:
## This PR contains:

- A NEW FEATURE (the `db-viewer` plugin)
- IMPROVED DOCS
- IMPROVED TESTS
- IMPROVED typings

There is no way to look at the data of a running RxDatabase. Debugging means `console.log`ing query results, and there is nothing that shows which index a query used, what replication actually sent, or what a write changed.

This adds `rxdb/plugins/db-viewer`, a database viewer and editor for a running [RxDatabase](https://rxdb.info/rx-database.html).

```ts
import { mountRxDBDbViewer } from 'rxdb/plugins/db-viewer';
const dbViewer = mountRxDBDbViewer(db);
```

## The UI is not in your bundle

The viewer UI is a React app that is built into **one self contained HTML file**, published with the docs at `https://rxdb.info/html/db-viewer.html`, and loaded into an **iframe**. The plugin is only the bridge that answers it over `postMessage`. This follows the pattern the `dev-mode` plugin already uses for its tracking iframe, including the `?version=` query.

```
your app                          iframe (rxdb.info)
┌───────────────────────────┐     ┌───────────────────────────┐
│ RxDatabase                │     │ the viewer UI             │
│ rxdb/plugins/db-viewer  ──┼────▶│ asks for documents,       │
│   the bridge              │◀────┼─  counts, schema, plans   │
└───────────────────────────┘     └───────────────────────────┘
        postMessage, inside the same browser
```

| | before | after |
|---|---|---|
| plugin output, raw | 199 KB | 35 KB |
| plugin output, gzipped | 42.3 KB | 9.5 KB |
| angular example bundle | 1.07 MB | 973 KB |

**No data leaves the browser.** `postMessage` is a call between two documents, not a network request, and the page makes no request of its own once loaded. The only thing fetched is the HTML.

### What ships in rxdb

- `protocol.ts` the versioned message contract, `collect.ts` everything that reads RxDB internals, `bridge.ts` the RPC server plus the change, replication and live streams, `index.ts` the mount that creates the iframe. No UI code, no stylesheet, no fonts.

### What the viewer does

- **Table grid** with multi-select, inline cell editing, sorting and paging at 100 rows per page.
- **JSON view** with syntax highlighting and the "updated Ns ago" highlight while observing.
- **Document drawer** that stages edits and previews the exact `upsert()` in a WILL RUN block, changed lines highlighted, before Apply runs anything.
- **Blast-radius confirmation** for bulk deletes: match count, that deletes replicate to peers, and that tombstones remain until cleanup. The delete button stays disabled until the collection name is typed.
- **Live activity map** drawing app → collections → remote with 60s rolling counters, sparklines, particle lanes, burst bands above ~200 events/s and node pulses rate-limited to 4 Hz.
- **Schema panel** sampling stored documents and reporting their real types and presence against the declared schema.
- **Query lab** explaining the used index, its bounds, examined and discarded counts, and findings for uncovered fields, `$regex` scans and in-memory re-sorts.
- **Replication and Changes panels**, **Storage panel** with tombstone counts and cleanup, and a **stacked read-only layout** below 640px.

## Notes for review

- **Cross-origin safety.** The bridge only reads messages that come out of its own iframe window *and* carry the origin the page was loaded from, and always replies to that origin rather than `*`. The viewer cannot know the app's origin up front, so it posts a payload-free `hello` to `*` and pins the origin from the `welcome` it gets back; no data is ever posted to an unpinned origin. Both guards have tests.
- **Nothing crossing the boundary may hold a function**, so `DbViewerConnection.onDisconnect` becomes a `canDisconnect` flag reached with a method. A test asserts a snapshot survives `structuredClone`.
- **`Observe` changed meaning.** An `RxQuery` lives in the host page and its `$` cannot cross the iframe, so observing re-runs the query when the host reports a write to that collection. Documented as a limitation.
- **New `viewerUrl` option** so the page can be self-hosted, which is needed for offline apps and for a strict `frame-src` policy. The docs tell you to `curl` the file; it is deliberately not added to the npm package, which would cost every install 286 KB for an opt-in devtool.
- **New `onClose` option**, because the viewer draws its own close button and the embedding app needs to take down its own chrome.
- The generated HTML is gitignored and produced by `npm run build:db-viewer`, which `docs:build` now runs, so it cannot go stale.
- `mountRxDBDbViewer()` throws `DBV1` when there is no DOM. `RxDatabase.mountDbViewer()` is a `pluginMissing` stub, matching `registerWebMCP()`.
- `isLeader()` is read behind a try/catch because it throws without the leader-election plugin; the Instances panel reports leadership as unknown rather than claiming a roster RxDB does not publish.
- The plugin is documented on its own page only, not linked from `cleanup.md`, `dev-mode.md`, `rx-query.md` or `rx-database.md`.
- No new npm dependency: React, webpack and babel were already dev-dependencies.

## Todos

- [x] Tests
- [x] Documentation
- [x] Typings
- [x] Changelog

### Verification

- `npm run lint` and `npm run check-types` clean (the latter now also type-checks the viewer app).
- codespell clean over every path `test-code-style` scans; `check-em-dashes` and `check-code-block-line-length` clean.
- Full fast memory suite: 1437 passing, 0 failing.
- `test/unit/db-viewer.test.ts`: 25 tests covering the message guards, the bridge and the collectors against a real database, including that a foreign origin is ignored and that a snapshot survives a structured clone.
- `db-viewer/test`: 21 tests for the pure helpers, run by the new `npm run test:db-viewer` in CI.
- Driven end to end in Chromium with the host on one origin and the viewer on another: 255 documents in the grid, drawer opens, WILL RUN highlights the staged line, and Apply's write was confirmed by querying the real database from the host page. All five tool panels render.

### Known CI state

`test-others-runtimes` is red and **not caused by this PR**: the same job hangs identically on master since the deno 2.9.5 bump (#8960), and locally with this branch's test file removed.

This branch also carries two deno fixes in separate commits (`--exit` for the `test:deno:*` scripts, and not passing `Infinity` as the `Deno.Kv.list()` limit). Both are correct on their own merits and verified locally on deno 2.9.5, but **they do not make the CI job green** — it still hangs with zero output. See [this comment](https://github.com/pubkey/rxdb/pull/8940#issuecomment-5528972821) for exactly what is and is not established. They can be dropped from this branch or split into their own PR on request.

The other 24 checks are green.

---

https://claude.ai/code/session_018adcs7t6gXavJ9D7hB14KX


pubkey:
`example-react` is failing on this PR, and it is not caused by this PR. It is broken on `master`.

**What fails**

The testcafe run never gets past its first selector:

```
Cannot obtain information about the node because the specified selector
does not match any node in the DOM tree.
> | Selector('#insert-box input[name=name]')
```

That is a 30s selector timeout, not an assertion mismatch. The example app never renders at all.

**Why it is not this PR**

This branch changes no file under `examples/react`. Its only changes reachable from that example are additive to the core: a `mountDevtool()` stub on `RxDatabase`, the devtool types, and a `./plugins/devtool` entry in `exports`.

Verified rather than inferred. I built rxdb from a clean `af6fb65` worktree (no devtool code present, confirmed by `node_modules/rxdb/plugins/devtool` being absent), packed it, installed that tarball into `examples/react`, rebuilt, and loaded the app:

```
RESULT: selector NOT found within 30s (same as CI)
heroes-list present: 0
page errors: [ 'TypeError: a is not a constructor', ... ]
```

Identical failure with none of this PR's code in the bundle. The app throws `TypeError: a is not a constructor` during startup, so React renders nothing and `#insert-box` never exists.

The timeline also lines up: `example-react` was green on this PR's previous run, when it was based on `128854a`. It only started failing after the rebase onto `af6fb65`, so the regression landed in the 68 master commits in between.

**Repro without this branch**

```sh
git worktree add /tmp/rxdb-master origin/master
cd /tmp/rxdb-master && npm install && npm run build
npx yarn pack . --filename /tmp/rxdb-master.tgz
cd <rxdb>/examples/react
cp /tmp/rxdb-master.tgz ./rxdb-local.tgz
rm -rf node_modules/rxdb && npm i --legacy-peer-deps && npm run build
# serve build/ and open it: the page is blank and the console shows
# "TypeError: a is not a constructor"
```

**Status**

I did not fix it here, because it is a master regression unrelated to the devtool and fixing it in this PR would mix two unrelated changes. I also could not re-run the job: `rerun-failed-jobs` returns `403 Resource not accessible by integration` for me. Someone with write access re-running it will not help either, since this reproduces deterministically rather than flaking.

Everything else on this PR is green, and I am still watching it.

---

pubkey:
Correcting my previous comment: I said "everything else on this PR is green", and that was wrong. `example-vue` is also not green. The final tally on `5d03f83` is 23 success, 1 failure (`example-react`), 1 cancelled (`example-vue`).

`example-vue` is a different problem from `example-react`, so the previous comment's diagnosis does not carry over to it.

**What happens**

The job ran from 08:02:42 to 08:32:44, exactly 30 minutes, so it hit `timeout-minutes: 30` rather than failing an assertion. The cleanup shows testcafe and Chrome still alive when the runner killed it:

```
Terminate orphan process: pid (2977) (xvfb-run)
Terminate orphan process: pid (6506) (npm run test)
Terminate orphan process: pid (3203) (chrome_crashpad_handler)
```

**What I checked**

I ran the same clean-master experiment as before: rxdb built from `af6fb65` with no devtool code (`node_modules/rxdb/plugins/devtool` absent), installed into `examples/vue`, rebuilt, and loaded the app.

Unlike the react example, the vue app **renders fine**:

```
rendered body length: 28
body preview: "RxDB - vuejs\nHeroes\n\n\nSubmit"
page errors: only ERR_CONNECTION_REFUSED / 404 for the sync endpoint,
             which is expected with no local replication server
```

No `TypeError: a is not a constructor`. So whatever hangs the vue job in CI is in the e2e harness, not in the app failing to boot.

**What I did not establish**

I could not reproduce the 30 minute hang locally, because it needs the CI harness (xvfb, the pouchdb sync server, testcafe against a real Chrome). So I cannot say from direct evidence whether the vue timeout is a pre-existing master problem or a flake.

What I can say: this branch changes no file under `examples/vue`, and `example-vue` passed on this PR's previous run against the old base `128854a`. So it is not the devtool diff, and it started with the rebase onto `af6fb65`, the same window that broke `example-react`.

Still watching the PR.

---

pubkey:
Closing the loop on my two earlier comments about `example-react` and `example-vue`: both are green now, and all 25 checks pass on `4d99aa6`.

The branch was rebased onto `ed249b9`, which brought in `74ef218 refactor: switch to named imports for ajv and async-test-util (#9030)`. That is the same import-interop class as the `TypeError: a is not a constructor` I had reproduced from a clean master checkout, so the react failure was indeed a master regression that master has since fixed. `example-vue` had been cancelled at its 30 minute timeout and now finishes in 4:35.

Nothing in this PR changed to make them pass; please read my two earlier comments as resolved rather than as open issues.

---

pubkey:
`test-others-runtimes` is red on this PR. It is not caused by these changes: the same job hangs on master.

**On master**, [run 33747641664](https://github.com/pubkey/rxdb/actions/runs/33747641664) for `d069027` ("Update dependency deno to v2.9.5", #8960) has `test-others-runtimes` as its only non-success job, cancelled after 30m16s. That commit is a renovate version bump and contains none of this PR's code.

**On this PR**, the job is cancelled after 30m15s with an identical log signature. In both jobs the last output is `# transpiling DONE`, then nothing at all until the runner kills it, and both end with the same orphan process list:

```
Terminate orphan process: pid (timeout)
Terminate orphan process: pid (bash)
Terminate orphan process: pid (npm run test:deno:dexie)
Terminate orphan process: pid (sh)
Terminate orphan process: pid (deno)
```

So `deno run ... npm:mocha ./test/unit.test.ts` hangs before mocha prints a single line. Because the step is `timeout 30m bash -c 'until npm run test:deno:dexie; do sleep 5; done'`, a hang there always burns the full 30 minutes and cancels the job.

One observation that may matter for the fix: both logs show `Going to install stable version 2.9.5`, including on this branch, whose `main.yml` still pins `deno-version: "2.6.5"` (this branch is based on `ed249b9` and predates #8960). So the pin does not appear to be taking effect, and pinning an older deno may not be enough to work around it on its own.

The other 24 checks on this PR are green.

---

pubkey:
Followed up on my previous comment: I installed deno 2.9.5 locally and reproduced the `test-others-runtimes` hang, so I could stop guessing. It is two separate master bugs, both now fixed in `a85a31d`.

First, I confirmed it is not this branch: the run hangs identically with `test/unit/db-viewer.test.ts` commented out of `test/unit.test.ts`.

**1. The deno runs never exit.** With unbuffered output the picture is clear:

```
  init.test.ts
    ✔ ensure dev-mode is activated
    ✔ init storage
    ✔ clear BroadcastChannel tmp folder
START TEST SERVERS
    ✔ start test servers (84ms)

  4 passing (87ms)
```

The tests pass in 87 ms and then the process hangs forever. The servers `init.test.ts` starts on 18000/18001/18006/18007 keep the event loop alive. `config/.mocharc.cjs` already sets `exit: true`, but the `test:deno:*` scripts invoke mocha without `--config`, so they never picked it up the way the node runs do. Older deno exited regardless; 2.9.5 does not. Adding `--exit` makes the run finish in 0.96 s.

In CI this matters doubly, because the step is `timeout 30m bash -c 'until npm run test:deno:dexie; do sleep 5; done'` — a hang there always burns the full 30 minutes and cancels the job.

**2. With the runs finishing, `test:deno:denokv` then failed for real:**

```
Error: Limit must be a positive integer: received Infinity
  at Kv.list (ext:deno_kv/01_db.ts)
  at queryDenoKV (src/plugins/storage-denokv/denokv-query.ts:106)
```

`limit` is `skip + limit`, which is `Infinity` when the query has no limit (`denokv-query.ts:29`). Deno rejects that since 2.9. Omitting the option reads the whole range, which is what an unlimited query means.

Verified locally on deno 2.9.5: `test:deno:dexie` 1279 passing and `test:deno:denokv` 1272 passing, both exiting 0. `check-types`, `lint` and the node fast-memory suite (1437 passing) are unaffected.

Neither fix has anything to do with the db-viewer plugin. They are in their own commit and are easy to split into a separate PR if you would rather land them independently of this feature.

---

pubkey:
`storage-mongodb` went red on `a85a31d` with 817 passing, 1 failing:

```
1) reactivity.test.ts
     preact-signals.test.ts
       should get the signal and clean up correctly:
   AssertionError: Expected values to be strictly deep-equal:
   0 !== 1
     at test/unit/reactivity.test.ts:200
```

Not caused by this PR, and not mongodb-specific — it is a race in the test itself:

```ts
await collection.insert(schemaObjects.humanData());
assert.deepStrictEqual(querySignal.value.length, 1);   // line 200
```

`insert()` resolves once the write is done, which is before the query has re-emitted and updated the preact signal. Three lines above, the *initial* value is already waited for with `await waitUntil(() => !!querySignal.value)` for exactly this reason; the value after the insert was not. It passes when the emission happens to land in the same tick and loses the race on a slower storage.

Fixed in `6b15914` by waiting for it the same way. The assertion keeps its meaning: `waitUntil()` still fails, by timeout, if the signal never picks the insert up.

My commit before this one touched only the deno test scripts, `denokv-query.ts` and a changelog file, none of which the mongodb job's reactivity test loads.

I do not have permission to re-run jobs on this repo (`rerun-failed-jobs` returns 403), so I cannot retry the check directly; the push re-triggers CI instead. I will keep watching the PR until it is green.

---

pubkey:
Correction to my previous comment: **the two deno fixes did not fix `test-others-runtimes`.** It hung again on `6b15914` and hit the 30 minute timeout. I overstated what I had shown, and want to be precise about what is and is not established.

The job log confirms `--exit` is being used:

```
> deno run ... npm:mocha ./test/unit.test.ts --bail --exit
# transpiling DONE (4 CPUs)          16:10:27
##[error]The operation was canceled.  16:38:26
```

28 minutes, zero output. The bun step just above it passes and prints its whole suite, so the job gets that far fine.

What is established:

- It is not this PR. It hangs identically on master (`d069027`), and locally with `test/unit/db-viewer.test.ts` commented out.
- `--exit` fixes a real latent bug. Without it the deno runs never exit: locally I get `4 passing (87ms)` and then the process hangs forever, because the test servers keep the event loop alive and the `test:deno:*` scripts never passed `--config ./config/.mocharc.cjs`, so they never picked up its `exit: true`.
- The `Infinity` limit in `queryDenoKV()` is a real deno 2.9 incompatibility that surfaces once a run does complete.

What is **not** established: why CI still hangs. Locally on deno 2.9.5 the full suite passes (`test:deno:dexie` 1279 passing, `test:deno:denokv` 1272 passing, both exit 0), including when I run the bun step first the way CI does, and with no lingering listeners on 18000/18001/18006/18007. So my local environment differs from the runner in a way I have not identified, and the two fixes are necessary but not sufficient.

Both fixes are in their own commits and are correct on their own merits, but since they do not deliver a green job here they are arguably just noise in a db-viewer PR. Happy to drop them from this branch, or split them into a separate PR, whichever you prefer.

I still cannot re-run jobs on this repo (403), so I cannot retry the check directly.

---

pubkey:
`test-others-replications` is red on `bb4352f`. It is not caused by these changes: the NATS test connects to the broker ~110 ms before the docker container has bound the port.

**What fails**

```
replication-nats.test.js
  init
    1) wait for server to be reachable
       NatsError: CONNECTION_REFUSED
0 passing (4ms)
```

Four milliseconds — the test did not wait for anything. The timestamps in the same log show why:

```
15:12:19.8594961  [1] 1) wait for server to be reachable      ← client gives up
15:12:19.9413015  [0] [INF] Listening for client connections on 0.0.0.0:4222
15:12:19.9413374  [0] [INF] Server is ready
```

The broker came up 82 ms *after* the test had already failed. `concurrently` starts `nats:start` and mocha in parallel, so this is a pure startup race.

**Why the retry loop does not catch it**

`test/replication-nats.test.ts` opens the connection at module scope, in an IIFE that runs on import:

```ts
const connectionSettings = { servers: 'localhost:4222' };
const connectionStatePromise = (async () => {
    const nc = await connect(connectionSettings);   // one attempt, no retry
    ...
})();
```

and the reachability test then does:

```ts
it('wait for server to be reachable', async () => {
    await connectionStatePromise;   // ← already rejected
    await waitUntil(async () => { ... });   // never reached
```

So the `waitUntil` that is supposed to wait for the server can never run: the single `connect()` has already settled before mocha executes the first `it`. The test named "wait for server to be reachable" is the one thing in the suite that cannot wait for the server to be reachable.

**Why it is not this PR**

- The same job passed on the previous head of this same branch, `6b15914`, with identical db-viewer changes.
- The only difference on `bb4352f` is the rebase onto `3323fd1`. The eight new master commits touch `doc-cache.ts`, `incremental-write.ts`, the react hooks and their tests, plus a docs page — nothing NATS-related, nothing in the CI setup.
- Nothing in this PR touches NATS, docker, or test ordering.

**Fix**

`nats.js` has an option for exactly this: `waitOnFirstConnect: true` makes the initial `connect()` retry instead of rejecting on `CONNECTION_REFUSED`. Applied to the test's own connection (not to `connectionSettings`, which is also handed to `replicateNats`), that makes the module-scope promise survive a slow container start and lets the existing `waitUntil` do its job.

I have deliberately **not** put that into this branch. This PR already carries three unrelated master fixes (`a85a31d`, `6b15914`) whose place here is still an open question, and adding a fourth would make that worse. Happy to open it as its own PR, or to fold it in here — your call.

I do not have permission to re-run failed jobs on this repository (`rerun-failed-jobs` returns 403), so I cannot retry the check myself. A re-run should be enough to make it green, since this is a timing race and not a deterministic failure.


---

PR #8941, opened 2026-08-10
[BREAKING] Add end-to-end TypeScript typings for mango queries

pubkey:
## This PR contains:
- IMPROVED typings
- IMPROVED TESTS
- IMPROVED DOCS

A query against a typed `RxCollection<DocType>` was only partially type-checked. The selector keys were already constrained to the document's dot-paths, but the operator payloads were `MangoQueryOperators<any>`, the `sort` type accepted any string key, the `index` option accepted any string, the chained query builder (`.where('x').gt(1)`) was untyped, the update operators (`$set`, `$inc`, ...) accepted anything, and `doc.get()` returned `any` for any string path. This means `{ age: { $regex: 'foo' } }` compiled on a `number` field and a typo in a sort field compiled and silently sorted by a non-existent field.

With this PR, the whole query surface is type-checked against `DocType`. This is a types-only change: no runtime behavior changes, no new runtime code, no new dependencies.

### What is typed now

- **Selector keys**: unchanged mechanism (template-literal dot-paths like `'address.city'`), capped at a documented depth (see below). A typo in a field name is a compile error.
- **Per-field-type operators**: `$gt`/`$gte`/`$lt`/`$lte` only on numbers and strings, `$regex`/`$options` only on strings, `$elemMatch`/`$size` only on arrays (with `$elemMatch` typed against the array *item* type, which the old typing got wrong), `$mod` only on numbers, `$in`/`$nin` take arrays of the field's type, `$eq`/`$ne`/`$exists`/`$type` everywhere.
- **Sort**: `sort` accepts only dot-paths that exist on the document type, and one sort part must contain exactly one field, because the key order of a JSON object is not deterministic. Multiple sort fields are given as multiple sort parts. (The runtime keeps accepting multi-key parts and splitting them; the type marks them as an error because their order is ambiguous.)
- **Index**: the `index` option only accepts field paths of the document type, including readonly arrays for `as const` index definitions.
- **Chained query builder**: `.where('age')` only accepts existing field paths and returns a builder state (`RxQueryFieldSelector`) whose operator methods are constrained by that field's type, so `.where('age').regex('foo')` is a compile error while `.where('age').gt(10).lte(20)` compiles. `.sort()` accepts typed sort parts, `'field'` and `'-field'` strings.
- **Update operators**: `UpdateQuery<DocType>` types the field paths of `$set`/`$unset`/`$min`/`$max`/`$rename` (any known field), `$inc` (number fields), and `$push`/`$addToSet`/`$pop`/`$pullAll` (array fields, values typed by the item type). `doc.update()` and CRDT entries run on `WithDeleted<DocType>` so soft-deleting via `$set: { _deleted: true }` stays typed, matching the runtime.
- **`doc.get()` / `get$()` / `get$$()`**: the path is checked against the document type and the return value has that field's type. On `RxLocalDocument` the paths are typed relative to the `data` property, because the runtime prepends `data.` to the given path (the old type test asserted `get('data').foo`, which never worked at runtime).
- **Results**: result typing of `find()`/`findOne()`/`exec()` is unchanged. No result-type narrowing based on `$exists`/`$type` (out of scope for this version).

### Depth cap

Dot-paths are generated for up to **6 levels of nesting** (`MangoQueryPathsMaxDepth = 5` passed to `Paths<T, D>`, which yields paths of up to 6 segments). The cap is documented in the type's JSDoc and in the docs. Deeper fields need the escape hatch.

As part of this, the shared `Paths<T>` utility type now special-cases arrays: it generates `items.${number}.field` paths instead of recursing into `keyof Array` (which used to generate garbage paths like `items.push` and `items.length` and was the main source of type instantiations).

### Backwards compatibility and escape hatch

- Untyped collections (`RxCollection<any>`, `MangoQuery` without a type argument) and documents with index signatures (`{ [k: string]: any }`) behave as before; when a field's type cannot be determined it falls back to `any`, so no operator constraint applies. `RxCollection<any>` stays assignable to and from `RxCollection<DocType>` in both directions (covered by type tests, this constraint is why the sort part of untyped documents resolves to `any`).
- The escape hatch for dynamically built queries is a cast at the call site: `collection.find(dynamicQuery as MangoQuery<DocType>)`. A `MangoQuery<any>` object is no longer directly assignable to a typed query parameter, because the exactly-one-key sort part cannot accept a wide object type. This is documented in `rx-query.md`. Same for updates via `UpdateQuery<any>`.
- Hand-built loose query objects against typed collections will start erroring; that is the feature. Inside this repo, 21 test sites needed an `as any` cast; almost all of them are tests that intentionally build invalid queries to assert runtime validation errors (QU13/QU14/QU17), pass storage-level fields (`_deleted`, `_id`) into user-level query types, or exercise the runtime's multi-key sort splitting.
- `RxDocument<{}>` (the default) keeps plain `string` paths for `get()`/`get$()`/`get$$()` via `MangoQueryPathsOrString`, because it is used internally as "any document".
- The minimum supported TypeScript version is now explicitly 4.1 (template literal types). The TypeScript tutorial page previously claimed 3.8; it is updated and the changelog states the new floor. (De facto the previous typings already used template literal types, so 4.1 was already required.)

### Compile time benchmark

Measured with `tsc --extendedDiagnostics` (TypeScript 5.9.3, Node 22), best of interleaved runs.

**Isolated benchmark** (a 6-level nested document type with ~30 typed queries, only the mango query types in scope):

| | before | after |
|---|---|---|
| Type instantiations | 28,862 | 51,850 |
| Check time | 0.27s | ~0.3s |
| Memory used | ~75 MB | ~70 MB |

The instantiation increase comes from the exactly-one-key sort part, which excludes all other paths per sort key (quadratic in the number of paths, computed once per document type and then cached).

**Whole repo** (`tsc` over the full project, 3 runs each):

| | before | after |
|---|---|---|
| Total time | 12.5s / 12.7s / 12.5s | 11.1s / 11.7s / 11.4s |
| Check time | 9.29s | 7.61s (-18%) |
| Type instantiations | 730,961 | 523,043 (-28%) |

The per-operator constraints did not blow up compile times, so nothing had to be relaxed; the array special-casing in `Paths` more than pays for the added conditional types and the sort part exclusion.

### Type tests

`test/typings.test.ts` (wired into `npm run test:typings`) gets new blocks asserting both directions for every feature:
- valid queries compile: nested dot-paths, correct operators per field type, array `$elemMatch` on item selectors and scalar items, typed sort with nested paths, typed builder chains, typed `index` (including `as const`), typed updates including `$set: { _deleted: true }`, typed `doc.get()`/`get$()`, untyped/loose collections, and the cast-at-the-call escape hatch.
- invalid queries fail to compile via `@ts-expect-error`: typoed field names (top-level, nested, inside `$or`, in `where()`, in `sort()`, in `index`, in `$set`, in `get()`), wrong-type operators (`$regex` on number, `$gt` on boolean, `$elemMatch`/`$size` on non-arrays, `$inc` on strings, `$push` on non-arrays), wrong-typed values, multi-key sort parts, and invalid sort paths.

One TypeScript quirk worth knowing for review: distributing the sort keys over a conditional chain like `Paths<T> extends infer K ? K extends string ? ...` silently fails to enforce the exactly-one shape; the distribution must happen over a naked type parameter of a helper generic (`MangoQuerySortPartDistribute`). The broken variant compiles and accepts multi-key sort parts, which is why the type tests assert the negative case.

## Todos
- [x] Tests
- [x] Documentation
- [x] Typings
- [x] Changelog

https://claude.ai/code/session_013rQyZu2L4HAAJc91NupfZh


pubkey:
This is a false positive of the reproduction check: the PR is a types-only feature, so the changed runtime tests are expected to pass without the source changes. The runtime test edits are only `as any` casts on tests that intentionally build invalid queries (QU13/QU14 validation and storage-level fields), added so those tests keep compiling under the stricter typings.

The tests that verify the change are the type tests in `test/typings.test.ts`, which run through `npm run test:typings` at compile time, not in the browser suite. I verified they fail without the source changes: checking out the previous `src/types` and rebuilding yields 9 errors, 8 `TS2578: Unused '@ts-expect-error' directive` (the operator and sort constraints that the old typings did not catch, for example `$regex` on a number field) and one `TS2322` on a valid query that the old typings wrongly rejected (`{ tags: 'foo' }` matching an item of an array field).

---

PR #8994, opened 2026-08-20
ADD vector search plugin based on TurboQuant

pubkey:
## This PR contains:

- A NEW FEATURE (the `vector-turboquant` plugin)
- IMPROVED DOCS
- IMPROVED TESTS
- IMPROVED typings

RxDB has the `vector` plugin with exact distance functions, but no index. To run a similarity search you have to fetch every embedding from the storage and compare it with the query. The [vector database article](https://rxdb.info/articles/javascript-vector-database.html) measures that at around 700ms for 10k documents, and it grows linearly. Keeping the raw `float32` embeddings in memory instead is expensive: 100k documents with 768 dimensions need 293 MB.

This PR adds a vector index that stores the embeddings in a compressed form, so a search does not read the raw vectors at all.

## What it does

The plugin implements [TurboQuant](https://arxiv.org/abs/2504.19874), a data-oblivious quantizer from Google Research, in the same spirit as [turbovec](https://github.com/RyanCodrai/turbovec). An embedding is stored with 1, 2 or 4 bits per dimension instead of 32 bits:

1. **Normalize**: the length of a vector is stripped off and stored separately.
2. **Rotate**: a randomized block Walsh-Hadamard transform turns the coordinates of any dataset into the same normal distribution. It runs in `d * log(64)` time instead of the `d * d` of a full random orthogonal matrix and preserves dot products exactly. Blocks of 64 keep the padding at 63 dimensions at most, while padding to the next power of two would waste 33% for the common sizes like 768 or 1536.
3. **Quantize**: a Lloyd-Max codebook for the normal distribution, computed with the Lloyd iteration. Because the rotation already normalized the distribution, the same codebook fits every dataset, so there is no training step and no rebuild when the collection grows.
4. **Renormalize**: quantization shortens a vector and makes every later score too small. One dot product per vector during encoding gives the correction factor that removes the bias.
5. Optional per coordinate calibration (the TQ+ step), fitted by minimizing the reconstruction error on a sample.

A search rotates the query into the same space and scores it against the packed codes with a per byte lookup table. The vectors are never decompressed.

### API

```ts
import { addRxPlugin } from 'rxdb/plugins/core';
import { RxDBVectorTurboQuantPlugin } from 'rxdb/plugins/vector-turboquant';
addRxPlugin(RxDBVectorTurboQuantPlugin);

const vectorIndex = await myCollection.addVectorIndex({
    identifier: 'semantic-search',
    dimensions: 384,
    bitWidth: 4,
    distance: 'cosine',
    embedding: docData => docData.embedding
});

const results = vectorIndex.search(queryVector, 10);
const documents = await vectorIndex.searchDocuments(queryVector, 10);
```

- `addVectorIndex()` reads the collection once and then follows the change stream, so inserts, updates and deletes are reflected in the next search.
- `search()` supports `cosine`, `dotProduct` and `euclidean`, plus an `allowlist` of ids so a vector search can be combined with a normal Mango query.
- `TurboQuantIndex` also works standalone, and `serializeTurboQuantIndex()` writes the whole index into a binary blob so it does not have to be rebuilt on the next start.
- `error$` emits the errors that a user provided `embedding` function threw while a change event was processed.

## Measured results

Recall on 10k random vectors with 384 dimensions and 200 queries. `recall@k` is the share of queries where the exact nearest neighbor was inside the first `k` results:

| Bit width | recall@1 | recall@4 | recall@10 | Memory for 100k x 768 |
| --------- | -------- | -------- | --------- | --------------------- |
| float32   | 100%     | 100%     | 100%      | 293 MB                |
| 4         | 82.5%    | 100%     | 100%      | 37 MB                 |
| 2         | 42.0%    | 73.0%    | 88.0%     | 18 MB                 |
| 1         | 13.0%    | 30.5%    | 46.5%     | 9 MB                  |

Random vectors are the hardest case because in high dimensions all of them have nearly the same distance to each other. The relative error of the estimated scores is about 10% at 4 bit, which matches the distortion of the Lloyd-Max quantizer at that bit width. Scanning 100k vectors of 768 dimensions at 4 bit takes about 100ms on a single core.

The Lloyd-Max solver reproduces the published values from J. Max, "Quantizing for minimum distortion" (1960) to four decimals, and the rotation turns a one-hot vector, the worst case for the quantizer, into coordinates with a variance of 1.00 and a kurtosis of 3.02.

## Todos

- [x] Tests
- [x] Documentation
- [x] Typings
- [x] Changelog

41 testcases in `test/unit/vector-turboquant.test.ts` cover the codebook, the rotation, the index, the calibration, the serialization and the collection integration. They pass against both the memory and the dexie storage. `npm run lint` and `npm run check-types` are clean and `npm run docs:build` succeeds.

New docs page at `docs-src/docs/vector-search.md`, linked in the sidebar and from the vector database article, the fulltext search page and the RxPipeline page. The outdated sentence in the vector database article saying RxDB has no dedicated vector plugin was updated.

https://claude.ai/code/session_013XoBX9R7x2RZvKHUvavxXS

---
_Generated by [Claude Code](https://claude.ai/code/session_013XoBX9R7x2RZvKHUvavxXS)_


nickchomey:
This looks like a great idea. I wonder if it might benefit from some sort of existing wasm implementation that uses SIMD etc? There's various JS and TS repos in github that implement things like this. A few selected at random:

https://github.com/zlaabsi/turboquant-wasm
https://github.com/a-tokyo/quantvec
https://github.com/teamchong/turboquant-wasm

pubkey:
Thanks, that is worth weighing. I checked all three: `turboquant-wasm` (321 stars), `quantvec` (10), `@zlaabsi/turboquant-wasm` (7). They are real and published on npm, so this is a genuine option, but all three are young, and a hard dependency would tie a core plugin to their release cadence.

Where the time actually goes in this implementation, measured on Node.js 22 on a single core:

- Scanning `100k` vectors of `768` dimensions at 4 bit: about `100` ms
- Storing one vector: about `65` µs, most of it the rotation

SIMD would make the scan several times faster. But in a local-first app the embedding model dominates: transformers.js needs a few hundred milliseconds per document, so at the corpus sizes that fit on a user device the scan is rarely the bottleneck. The `allowlist` option cuts it down further by restricting the scan to the result of a normal query.

The arguments against bundling WASM into this plugin:

- RxDB has no WASM dependency today and targets browser, Node.js, Electron, React Native, Capacitor, Deno and Bun. Relaxed SIMD is narrower than that. `turboquant-wasm` documents Chrome 114+, Firefox 128+, Safari 18+ and Node 20+, and WASM on React Native is still rough. Plain TypeScript runs everywhere RxDB runs.
- WASM instantiation is async, so `search()` would either stop being synchronous or would need an explicit init step before the first query.
- Every consumer pays the bundler setup cost of loading a `.wasm` asset, which works differently in Vite, webpack, Metro and Deno.

A middle path fits the existing conventions better: keep the dependency-free TypeScript scorer as the default and make the scorer pluggable, the same way the [SQLite RxStorage](https://rxdb.info/rx-storage-sqlite.html) takes the engine from the user through `sqliteBasics` instead of bundling one. Then anyone who wants SIMD can supply a WASM scorer, and the plugin still works on runtimes without it.

I have not changed anything for this, because taking on a third party dependency for a core plugin is the maintainer's call rather than mine.

---

nickchomey:
1. as mentioned previously, those libraries were selected at random. There may be other, better, more mature ones. Perhaps even something implemented directly in zig or rust that could be used/ported. 
2. It is worth noting that transformers.js uses WASM (by default) or WebGPU

pubkey:
Decision on this: the plugin stays pure TypeScript.

The deciding argument is one I had underweighted, and it is stronger than the portability and dependency points I listed above: crossing the JavaScript to WASM boundary costs latency on every call. A search has to copy the query vector in and the results back out, and RxDB optimizes for latency. At the corpus sizes that fit on a user device, that per-call overhead eats into what SIMD wins back inside the kernel.

So no dependency and no change here. Thanks for raising it, the links were worth checking.

---

nickchomey:
So no transformers.js then on non-chromium browsers (which don't yet support webgpu)? How will the embedding work without wasm?

pubkey:
You are right, and my portability argument was the weak part of that answer. transformers.js runs on WASM by default and only uses WebGPU when you pass `device: 'webgpu'`, so on the browser path a user who computes embeddings locally already has WASM. Nothing changes about transformers.js here, it keeps working exactly as before. I also claimed WASM support is uneven on Deno, which is simply wrong.

What actually decides it is narrower than "WASM is not available", and the two cases are not symmetric:

- An embedding crosses the boundary once per document and that crossing is amortized over a few hundred milliseconds of model compute. Paying it is obviously worth it, which is why transformers.js does.
- A search crosses per query for work that is much smaller. At `10k` documents of `384` dimensions the whole scan is a few milliseconds, so the fixed marshalling cost is a real fraction of it rather than a rounding error.

So the plugin keeps the scorer on the JavaScript side, while the model next to it stays on WASM. Those are consistent positions, not contradictory ones. React Native is the remaining case where WASM itself is a problem, and it is a smaller part of the argument than I made it sound.

I have corrected the FAQ entry accordingly in 3204e7c8b: it now leads with the per-call boundary cost, states that transformers.js runs on WASM by default, and drops the Deno claim.

On your earlier point that the three libraries were picked at random and better or more mature ones may exist, including something in Rust or Zig: agreed, and I only evaluated those three. If a mature implementation does turn out to beat the TypeScript scorer end to end, including the marshalling, that is worth revisiting with a benchmark rather than on priors.

---

PR #9029, opened 2026-09-01
docs(jobs): full job titles on the tiles, and a page title worth indexing

pubkey:
Follow-up to #9028, which introduced the two tiles on `/jobs`.

## What changes

**The tiles now carry the full job title.** They showed a short label, `Schwerpunkt Frontend` and `Schwerpunkt Video und Social Media`, so a visitor scanning the page never saw what the position is actually called until they opened the modal. Both tiles now show the same heading the advert does:

- Werkstudent (m/w/d) Developer Experience, Schwerpunkt Frontend
- Werkstudent (m/w/d) Developer Experience, Schwerpunkt Video und Social Media

The `short` field is gone from the data, since the title covers both the tile heading and the `aria-label`.

**The page title is now `Jobs & Stellenangebote bei RxDB`.** It was `Jobs`, which says nothing in a search result and matches nothing anyone types. Because the new title already carries the brand, the swizzled `TitleFormatter` has to skip the sitewide ` | RxDB` suffix for it, exactly as it already does for `HOME_TITLE`, otherwise the tab reads `... bei RxDB | RxDB`. It stays an explicit list rather than a rule such as "ends with RxDB", because 13 documentation pages (`Partial Sync with RxDB` and similar) end with the brand and do rely on the suffix being appended.

**The developer-experience framing is out of the page-level text.** Both the meta description and the intro paragraph presented the whole page as being about developer experience. That belongs to the two adverts, which say it in their own titles, not to a page that lists whatever happens to be open.

**Mobile order fix.** The tile row used the shared `.inner` class, which becomes `flex-direction: column-reverse` below the mobile breakpoint. On a phone the video advert therefore appeared above the frontend one, in the opposite order to the desktop layout. The row is now a plain flex container with the same desktop behaviour (centered, wrapping, stretched), so both viewports read top to bottom in the same order. `.column-mobile` does not help here, it has lower specificity than `.block .inner` and loses.

## Verified

Built the docs and drove the built page with Playwright at 1440x1000 and 390x844:

- `<title>` on `/jobs/` is `Jobs & Stellenangebote bei RxDB`, with no suffix, and `og:title` matches
- the homepage title is unchanged, and all 13 docs titles that end with the brand still get ` | RxDB`
- both tiles render the full title, 2 and 3 lines, and stay equal height (344px each)
- the modal heading matches the tile that opened it
- mobile order is frontend first, video second
- the advert still opens scrolled to the top (`scrollTop` 0), so the earlier keyboard-focus fix is intact
- eslint clean on all three changed files

Screenshots are in the review thread rather than the repo.

### DIFF
--- a/docs-src/src/constants.ts
+++ b/docs-src/src/constants.ts
@@ -9,6 +9,12 @@ export const NON_PREMIUM_COLLECTION_LIMIT = 13;
 export const HOME_TITLE = 'RxDB - JavaScript Database';
 
 /**
+ * Title of the /jobs/ page. Like HOME_TITLE it already names the brand, so the
+ * swizzled TitleFormatter must not append the ` | RxDB` suffix to it.
+ */
+export const JOBS_TITLE = 'Jobs & Stellenangebote bei RxDB';
+
+/**
  * Premium tier prices in dollars per month, billed annually.
  * Shown on the premium page and used to calculate
  * the lead value on the premium-submitted pages.
--- a/docs-src/src/pages/jobs.tsx
+++ b/docs-src/src/pages/jobs.tsx
@@ -8,6 +8,7 @@ const FILE_EVENT_ID = 'jobs-link-clicked';
 import { triggerTrackingEvent } from '../components/trigger-event';
 import { IframeFormModal, Modal } from '../components/modal';
 import { Button } from '../components/button';
+import { JOBS_TITLE } from '../constants';
 
 /**
  * Pipedrive webform for job applications. Both positions use the same form,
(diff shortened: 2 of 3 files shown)


pubkey:
`example-react` is failing on this PR, and it is not this PR's failure.

**What fails**

```
✖ insert/edit/remove a hero
  1) Cannot obtain information about the node because the specified
     selector does not match any node in the DOM tree.
     > | Selector('#insert-box input[name=name]')
  at examples/react/test/e2e.test.js:23
  1/1 failed (31s)
```

The 31s is the `--selector-timeout 30000` running out. `react-scripts build` reports `Compiled successfully` earlier in the same job and `ws` serves the bundle on 8888, so the build is fine and the app simply never renders `#insert-box` in the browser.

**Why it is not this PR's**

The diff is one file, `docs-src/src/pages/jobs.tsx`. It touches no example, no source and no test. The same job fails identically on the base commit, master `fa4f4c3` ([run 33494752127](https://github.com/pubkey/rxdb/actions/runs/33494752127), job `example-react`): same test name, same selector, same line 23, same `1/1 failed (31s)`. `test-others-replications` is red on that master run too, so it may follow here for the same reason.

**No fix to port**

There is no open fix I could bring into this PR, and root-causing why the React example stops rendering is not something to bolt onto a docs change. Two things narrow it for whoever picks it up: the last fully green master run is `f7db479` on 2026-08-27, and the failure is a runtime one in the app rather than a build or serve problem.

I do not have permission to re-run jobs on this repository (the API returns 403), so I cannot spend the one confirming re-run myself. I will keep watching the PR until it is merged or closed.

---

PR #8948, opened 2026-08-12
bug-report test: document not found after cleanup() + re-insert of the same primary key

kilbot:
Filled in the bug-report template, per its instructions.

The sequence is: insert → remove() → cleanup(0) → insert the same primary key again. The second insert succeeds, but `findOne()` returns null for the document that was just written. Closing and reopening the database shows the document was stored correctly — the live instance is what's wrong. A fresh `find()` with a selector also misses the document while `count()` returns 1, so it isn't query-cache staleness.

Fails on current master with both `DEFAULT_STORAGE=memory` and `DEFAULT_STORAGE=dexie`, and with `eventReduce: false` too:

```
AssertionError [ERR_ASSERTION]: findOne() must return the document that was just inserted
```

While digging we noticed the conflict-retry for inserting a previously-deleted key (`reInserts` in rx-storage-helper.ts) emits its change event with `previousDocumentData: null` even though the write carried the tombstone as `previous` — that might be a useful place to start looking. Possibly related: #7984 (bulk revival of soft-deleted docs, SQLite).

Context: our production app keeps hitting a class of malformed-JSON errors in the premium OPFS storage, and delete → re-insert of the same primary key turned out to be the trigger. This test is the smallest storage-independent piece we could distill from it. The OPFS-specific side (index files growing by one row per delete → re-insert cycle) I'll file separately in rxdb-premium-issues with its own reproduction.

Finally — an apology for the earlier AI-heavy submissions here and in rxdb-premium-issues (#8841, premium PR 26). I'm honestly too busy to write these investigations up myself, so I lean on AI tooling; all I knew first-hand was that I kept getting this class of malformed-JSON bugs without knowing where they came from. I've tried to keep this one to exactly what the template asks for: a failing test.

PR #9059, opened 2026-09-03
FIX document not found after cleanup() and re-insert of the same primary key

pubkey:
## This PR contains:

A BUGFIX and IMPROVED TESTS.

Fixes the bug reported by the failing test in https://github.com/pubkey/rxdb/pull/8948.

The sequence `insert` → `remove()` → `cleanup(0)` → insert the same primary key again writes the second document correctly, but the live instance does not find it:

```ts
await collection.insert({ passportId: 'foobar' });
const doc = await collection.findOne('foobar').exec(true);
await doc.remove();
await collection.cleanup(0);

await collection.insert({ passportId: 'foobar' });

await collection.findOne('foobar').exec();                        // null
await collection.find({ selector: { passportId: 'foobar' } }).exec(); // []
await collection.find().exec();                                   // [doc]
await collection.count().exec();                                  // 1
```

### Root cause

The out-of-order protection of the `DocumentCache` (from #8609) compares document states by revision height only. When the cleanup purges the tombstone, the re-insert starts a new revision chain at height `1`, which is lower than the height `2` of the cached tombstone. So the change event of the re-insert is rejected and the cache keeps the tombstone as the latest known state of that document.

`findOne(primaryKey)` and the primary key fast-path of `find()` read the latest state through the doc-cache, so they see the tombstone and return nothing. `find()` without a selector and `count()` go to the storage and are therefore correct, which is why this does not look like query-cache staleness. Reopening the database creates a fresh cache and the document shows up again.

### The fix

The cached latest state may now also be overwritten when the cached state is deleted, the incoming state is not deleted, and the incoming state was written later (`_meta.lwt`). Across a cleanup boundary the revision height is meaningless, but the write time still orders the two states.

Comparing the write time keeps the out-of-order protection intact: a stale `INSERT` event that arrives after the delete carries a lower `lwt` than the tombstone, so it still does not resurrect the document.

An alternative fix would be to evict deleted entries from `cacheItemByDocId` after a cleanup pass. That was not taken because it drops the cache item while `RxDocument` instances for that document can still be alive, and `RxDocument.getLatest()` throws when the cache item is gone. It also only repairs the instance that ran the cleanup, while the fix here works for every instance that receives the change event.

## Todos
- [x] Tests
- [ ] Documentation
- [x] Typings
- [x] Changelog

Two regression tests, both verified to fail on `master` and pass with the fix:

- `cleanup.test.ts`: `#8948 must find a document that was re-inserted after the cleanup purged its tombstone` covers the whole reported sequence.
- `doc-cache.test.ts`: one test for the re-insert that restarts the revision chain, one that asserts an older change event still does not resurrect a deleted document.

`npm run lint`, `npm run check-types`, `npm run test:fast:memory` and `npm run test:fast:dexie` are green.

https://claude.ai/code/session_01VGzhxpqmBaCeNkqfh3tZj5

---
_Generated by [Claude Code](https://claude.ai/code/session_01VGzhxpqmBaCeNkqfh3tZj5)_

### DIFF
--- a/orga/changelog/fix-doc-cache-stale-tombstone-after-cleanup.md
+++ b/orga/changelog/fix-doc-cache-stale-tombstone-after-cleanup.md
@@ -1,0 +1,1 @@
+- FIX `DocumentCache`: a document that is re-inserted after the cleanup purged its tombstone was no longer found by `findOne()` and by primary key queries, because the new revision chain starts at height `1` and the out-of-order protection kept the cached tombstone. [#8948](https://github.com/pubkey/rxdb/pull/8948)
--- a/src/doc-cache.ts
+++ b/src/doc-cache.ts
@@ -136,7 +136,7 @@ export class DocumentCache<RxDocType, OrmMethods> {
                         const currentLatest = cacheItem[1];
                         if (
                             !currentLatest ||
-                            !isOlderDocumentState(documentData._rev, currentLatest._rev)
+                            isNewerDocumentState(documentData, currentLatest)
                         ) {
                             cacheItem[1] = documentData;
                         }
@@ -200,6 +200,31 @@ export class DocumentCache<RxDocType, OrmMethods> {
             return cacheItem[1];
         }
     }
+}
+
+/**
+ * Determines if the given document state has to overwrite
+ * the currently cached latest state of that document.
+ */
+function isNewerDocumentState<RxDocType>(
+    checkState: RxDocumentData<RxDocType>,
+    currentLatest: RxDocumentData<RxDocType>
+): boolean {
+    /**
+     * When the cleanup has purged the tombstone of a deleted document,
+     * a write to the same primary key starts a new revision chain at height 1.
+     * That state has a lower revision height than the cached tombstone
+     * while still being the newer state, so here we compare by the write time
+     * instead of by the revision.
+     */
+    if (
+        currentLatest._deleted &&
+        !checkState._deleted &&
+        checkState._meta.lwt > currentLatest._meta.lwt
+    ) {
+        return true;
+    }
+    return !isOlderDocumentState(checkState._rev, currentLatest._rev);
 }
 
 /**
--- a/test/unit/cleanup.test.ts
+++ b/test/unit/cleanup.test.ts
@@ -205,6 +205,41 @@ describe('cleanup.test.js', () => {
         });
     });
     describe('issues', () => {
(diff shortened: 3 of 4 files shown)


pubkey:
CI status on `3a306e9`: **24 of 25 checks pass**. The one exception is `test-others-runtimes`, which ended as `cancelled`. It is not a failure caused by this PR, and I am not widening the PR to fix it. Details, since the job log is easy to misread:

**The bun half of that job passed.** `npm run test:bun:memory` ran the complete unit suite on this commit and finished green at `16:02:01`, including the `last.test.ts` teardown assertions (all storage instances, collections, BroadcastChannels and leader electors closed). `cleanup.test` and `doc-cache.test` are part of that suite.

**The deno half never started a test.** The next step ran:

```
timeout 30m bash -c 'until npm run test:deno:dexie; do sleep 5; done'
```

`npm run transpile` completed at `16:02:09`, and then the log is completely silent until the runner killed the job at `16:30:21`. `deno run … npm:mocha ./test/unit.test.ts` printed nothing at all: no mocha banner, no `DEFAULT_STORAGE: dexie`, no `# use RxStorage: dexie`. Those lines appear within seconds on every other suite, so the process hung during deno startup, before any test body ran. A hang caused by the code in this diff would have printed the suite header and the tests up to the hang.

**Two things about the job make this self-inflicted by the workflow rather than by a commit:**

1. The step is already wrapped in a `until … do sleep 5; done` retry loop, and the deno cache key right above it carries the comment "deno install randomly fails". The retry loop exists because this step is known to be flaky.
2. The retry loop can never actually retry. The step's own `timeout 30m` is equal to the job's `timeout-minutes: 30`, and the job clock started about two minutes earlier, so the job always dies before the inner timeout fires and hands control back to the loop. Any deno stall therefore burns the whole job instead of being retried once.

Fixing that means editing `.github/workflows/main.yml` (lower the step timeout below the job timeout, or raise the job timeout), which is unrelated to the doc-cache change and would widen this PR, so I have left it alone. Happy to open it as a separate PR if you want it.

I have re-run the failed job once. If it goes green, this PR is green across the board.

---

pubkey:
Correction to the last line of my previous comment: I could **not** re-run the job. The re-run API returns `403 Resource not accessible by integration`, so I do not have Actions write permission here.

Everything else in that comment stands. The re-run needs a click on your side, or the job stays `cancelled` until the next push to this branch triggers a fresh run.

---

PR #9082, opened 2026-09-10
docs: rewrite the realtime-database article and drop the Firebase recommendation

pubkey:
## This PR contains:

Improved docs. A rewrite of `docs-src/docs/articles/realtime-database.md` (rxdb.info/articles/realtime-database.html).

The page was a short definition piece that positioned the **Firebase Realtime Database** as the reference product for realtime sync, and the FAQ pointed readers at Firebase and Supabase as the tools to use. Google itself now tells new customers to start with Cloud Firestore instead of the Realtime Database, and the page carried no numbers a reader (or an AI answer engine) could cite.

The rewrite keeps the three-meanings structure and the existing voice, roughly doubles the content, and replaces the Firebase recommendation with sourced facts.

### What changed

- **Three meanings of realtime**: added an overview table plus the hard/firm/soft real-time split, and browser-specific reasons why JavaScript cannot give timing guarantees (Chrome throttles chained timers in hidden tabs to one check per minute after 5 minutes, GC pauses, long tasks).
- **Realtime replication**: added the push-vs-poll math (10,000 clients polling every 5s is 2,000 req/s of mostly empty responses), the transport comparison, and a new section on what a change stream alone does not solve (missed events, ordering, offline writes, conflicts, multi-tab).
- **New section "How Realtime Replication Works Under the Hood"**: checkpoint iteration, push with the assumed master state, and the event stream as a trigger only, with a runnable `replicateRxCollection()` sample that includes the `RESYNC` on reconnect.
- **New section "Managed, Self-Hosted and Local-First Realtime"**: a three-column comparison over source of truth, read latency, offline writes, query engine, cost driver, scaling ceiling and data ownership. Self-hosted server-authoritative is its own column, because metering, vendor limits and lock-in come from renting a backend, not from keeping the truth on a server. Links the downsides-of-offline-first page so the tradeoff stays honest.
- **New section "RxDB as a Cached Materialized View of Your Backend"**: the self-hosted case where full local-first is off the table. Pull-only replication over your own change stream, writes going through your API with the accepted state coming back on that stream, partial sync for scoping, plus what the setup keeps (local queries, reactive updates, one connection per client, multi-tab via leader election) and what it gives up (offline writes, a staleness window bounded by your stream).
- **New section "Where the Hosted Realtime Databases Fall Short"**: the documented Realtime Database limits (200,000 simultaneous connections, 1,000 writes/second, 32 levels of nesting, 768 byte keys, 16 MB/256 MB write sizes, 100 connections on Spark), the query limit ("sort *or* filter on a property, but not both"), Blaze pricing ($5/GB stored, $1/GB downloaded), and a quote of Google's own "We recommend new customers start with Cloud Firestore", each with a source link and the date it was checked.
- **Fairness sections kept**: "When a Hosted Realtime Backend Still Makes Sense" and a note that teams already on Firestore can use the Firestore replication plugin as a migration bridge.
- **New section "How to Pick a Realtime Database"**: five checks (offline behavior, where the query runs, conflict handling, cost model, exit path).
- **FAQ rewritten**: no longer recommends Firebase or Supabase, adds questions on offline sync, transports, the 16ms budget, using RxDB as a cache in front of your own server, and whether to start a new project on the Realtime Database.
- Frontmatter description rewritten without the banned words, Firebase logo image removed, internal links added throughout (replication, partial sync, reactivity, leader election, transactions/conflicts, rx-storage, offline-first, optimistic-ui, local-first-future, websockets article, the two Firebase alternative pages).

Verified: no em dashes, no banned words, every relative link resolves to an existing file, and the page compiles with the MDX 3 compiler.

## Todos

- [ ] Tests 
- [x] Documentation
- [ ] Typings
- [ ] Changelog 

https://claude.ai/code/session_01XeLydydiCNEQjCiRCQvwSj

### DIFF
--- a/docs-src/docs/articles/realtime-database.md
+++ b/docs-src/docs/articles/realtime-database.md
@@ -104,7 +104,9 @@ const replicationState = await replicateRxCollection({
         async handler(lastCheckpoint, batchSize) {
             const minTimestamp = lastCheckpoint ? lastCheckpoint.updatedAt : 0;
             const response = await fetch(
-                `https://example.com/api/sync/?minUpdatedAt=${minTimestamp}&limit=${batchSize}`
+                `https://example.com/api/sync/` +
+                `?minUpdatedAt=${minTimestamp}` +
+                `&limit=${batchSize}`
             );
             const documents = await response.json();
             return {
--- a/docs-src/docs/articles/realtime-database.md
+++ b/docs-src/docs/articles/realtime-database.md
@@ -173,24 +173,72 @@ A competent realtime application is engineered to offer feedback or results swif
 
 The 16ms budget is also the reason why the two other meanings of realtime matter here. A query that has to cross the network can never hit it. This is what the [optimistic UI](./optimistic-ui.md) pattern and the [local-first](./local-first-future.md) architecture are about: the write goes to the local database first, the UI re-renders from local state, and the replication happens in the background.
 
-## Server-Authoritative Realtime vs Local-First Realtime
+## Managed, Self-Hosted and Local-First Realtime
 
-Two architectures both call themselves realtime, and the difference decides what your app can do when the network is bad.
+Three architectures all call themselves realtime, and the word "server-authoritative" covers two of them that behave differently in practice.
 
-**Server-authoritative** means the server holds the truth and the client holds a cache. Reads that miss the cache go to the server, and a write is confirmed when the server accepted it. The Firebase Realtime Database, Firestore and most hosted realtime backends work this way.
+**Managed realtime backend**: a vendor runs the server and you rent access to it. The truth lives in their cloud, the query engine runs there, and you are billed per operation. The Firebase Realtime Database and Firestore work this way.
 
-**Local-first** means the client holds a full database that is the source of truth for the UI. Reads and writes are local and synchronous from the app's point of view, and the [replication](../replication.md) moves data in the background.
+**Self-hosted server-authoritative**: your own server, on a VPS, on bare metal, or on a Raspberry Pi in the office, holds the truth. The client keeps a cache. Nothing here is metered by a vendor, and the data stays on hardware you control.
 
-| Property | Server-authoritative | Local-first |
-| --- | --- | --- |
-| Read latency | Network round trip on a cache miss | Local storage access |
-| Writes while offline | Queued, often with limits | Normal writes, persisted locally |
-| Query engine | Runs in the cloud, billed per operation | Runs on the device |
-| Cold start without network | Blocked or empty | Full app with all cached data |
-| Cost driver | Reads, writes and transferred bytes | The device the user already paid for |
-| Backend choice | The vendor's | Any, or none |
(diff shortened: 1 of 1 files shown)


docs-src/docs/articles/realtime-database.md
```diff
@@ -70,34 +169,120 @@ await db.heroes.find({
...
+
+## Server-Authoritative Realtime vs Local-First Realtime
+
+Two architectures both call themselves realtime, and the difference decides what your app can do when the network is bad.
+
+**Server-authoritative** means the server holds the truth and the client holds a cache. Reads that miss the cache go to the server, and a write is confirmed when the server accepted it. The Firebase Realtime Database, Firestore and most hosted realtime backends work this way.
+
+**Local-first** means the client holds a full database that is the source of truth for the UI. Reads and writes are local and synchronous from the app's point of view, and the [replication](../replication.md) moves data in the background.
+
+| Property | Server-authoritative | Local-first |
```
nickchomey: This table is conflating server-authoritative with managed cloud dbs. If I host my own backend on a vps, bare metal, raspberry pi, or whatever, then most of this doesn't apply.

I would love to see a 3rd column, or separate section, about self-managed server-authoritative, which is what I intend to do. I can't do full local first for reasons, so intend for rxdb in the browser to be a cached materialized view of my backend.

docs-src/docs/articles/realtime-database.md
```diff
@@ -70,34 +169,120 @@ await db.heroes.find({
...
+| --- | --- | --- |
+| Read latency | Network round trip on a cache miss | Local storage access |
+| Writes while offline | Queued, often with limits | Normal writes, persisted locally |
+| Query engine | Runs in the cloud, billed per operation | Runs on the device |
+| Cold start without network | Blocked or empty | Full app with all cached data |
```
nickchomey: This is not a fair comparison. They either both have the cached data or not. Id remove this row altogether

docs-src/docs/articles/realtime-database.md
pubkey: Fair point, the table was treating "server-authoritative" and "managed cloud database" as the same thing. Fixed in 67857d0.

The table now has three columns: managed realtime backend, self-hosted server-authoritative, and local-first, with rows for source of truth, read latency, offline writes, query engine, cost driver, scaling ceiling and data ownership. The section states plainly that the metering and the lock-in are properties of renting a backend, not of keeping the truth on a server, so self-hosting drops that part of the column.

Your use case also got its own section, "RxDB as a Cached Materialized View of Your Backend": pull-only replication over your own change stream, writes going through your API with the accepted state coming back on that stream, partial sync for scoping what each client holds, plus what the setup keeps (local queries, reactive updates, one connection per client, multi-tab via leader election) and what it gives up (offline writes, and a staleness window bounded by your stream).

---

docs-src/docs/articles/realtime-database.md
pubkey: Agreed, that row compared nothing. Whether the app starts with data offline depends on what was cached, not on where the source of truth lives. Removed in 67857d0.

---

PR #9112, opened 2026-09-15
docs: add "IndexedDB Query" article

pubkey:
## This PR contains:

Improved docs. One new article page plus the internal links that connect it to the existing IndexedDB cluster.

The docs explain how to *use* IndexedDB ([IndexedDB Tutorial](https://rxdb.info/indexeddb-tutorial.html)), why it is *slow* ([Slow IndexedDB](https://rxdb.info/slow-indexeddb.html)), and how to model *relations* in it ([IndexedDB Relationships](https://rxdb.info/indexeddb-relationships.html)), but nothing covers the single thing developers hit first and hardest: **querying**. "How do I query IndexedDB", "how do I sort IndexedDB results", "how do I paginate IndexedDB", and "why is my IndexedDB query slow" are high-intent searches with no page on rxdb.info to answer them, and they lead directly to the RxDB query engine.

This PR adds `docs-src/docs/articles/indexeddb/indexeddb-query.md` (slug `indexeddb-query.html`) to fill that gap.

### What the article covers

1. **How querying works in IndexedDB.** The four read operations (`get`, `getAll`, `count`, `openCursor`) and their key-only variants, which value types are valid keys and the cross-type sort order, building conditions as `IDBKeyRange` bounds including the prefix-range trick, reading by primary key vs. through a secondary index, cursor directions (`next`, `prev`, `nextunique`, `prevunique`) and the three ways to move one, and why the index order is the only ordering available.

2. **Transaction lifetime while querying.** A read transaction goes inactive when control returns to the event loop, so it cannot survive an `await` on anything that is not an IndexedDB request. Shows the failing version next to the working one, explains why a cursor walk driven from `onsuccess` is safe, and covers the readonly/readwrite concurrency rules that follow from the same mechanism.

3. **How the right index improves performance.** Full scan vs. index range, the compound-index field-order rule worked through with both `['priority', 'category']` and `['category', 'priority']` on the same query, letting the index also serve the sort, key-only reads (`count`, `getAllKeys`, `openKeyCursor`), batching reads with `getAll()` instead of one `continue()` per record, and the write and upgrade cost of every added index.

4. **Running queries in a Web Worker.** IndexedDB reads do not block the main thread but the deserialize, filter and sort work around them does. Worker and main-thread code, the structured clone cost of the returned result, SharedWorker as one worker per origin instead of per tab with its support history, and the measured startup tradeoff. Links the Worker and SharedWorker RxStorages.

5. **A complex query implemented on the raw API.** A `<Steps>` walkthrough of "the ten open todos in category `work` with the highest priority, skipping the first ten": storing the boolean as `0`/`1` because booleans are not valid keys, the `['category', 'done', 'priority']` compound index, the `['work', 0, []]` upper bound (arrays sort after every other key type), and a `prev` cursor with `advance()` for skip and an early `resolve()` for limit. Followed by what breaks as soon as one requirement moves.

6. **The limits of IndexedDB querying.** No query language, one index per read, no `$or`, sorting only by an index, skip is a walk, booleans/`null`/`undefined` cannot be indexed, compound indexes need every field present, no substring or case-insensitive matching, no aggregation beyond `count()`, no joins, no `EXPLAIN`, no reactivity.

7. **How RxDB solves it.** Mango queries with the full operator set, indexes declared in the RxSchema, the query planner plus the `index` override and the Query Optimizer, reactive results via EventReduce, count/populate/full-text, and the same query running on any RxStorage.

8. **Query performance across storages.** Renders the existing browser storage benchmark with `<PerformanceChart>`, limited to the read-side metrics via `skipMetrics`, then reads the query and count numbers off it. Followed by the comparison table and the FAQ.

The page names no third-party IndexedDB wrapper and links to no competitor site.

### Accuracy

Claims were checked against source rather than written from memory:

- The Mango operator list comes from the operators actually registered in `src/rx-query-mingo.ts`.
- The "a boolean index field must be `required`" note comes from the `SC38` check in `src/plugins/dev-mode/check-schema.ts`.
- The count-query example was verified against `isSelectorSatisfiedByIndex` in `src/query-planner.ts` so it does not need `allowSlowCount`.
- Key types, the `getAll(query, count)` signature and `SharedWorker` support are cited to MDN, linked in the page.
- The 43% figure is the existing sharding benchmark from `slow-indexeddb.md`, attributed to sharding, not re-derived.
- **No new measurements were taken.** Every number in the benchmark and Worker sections comes from the measured values already in `docs-src/src/components/performance-data.ts`, and each was verified against that file before pushing.

### Internal linking

Registered in `docs-src/sidebars.js`, and linked in both directions per the repo's article linking rule. Inbound in-text links were added on relevant keyword phrases in:

- `articles/indexeddb/indexeddb-tutorial.md` (two places: the index-query step and the "Limited querying" limit)
- `articles/indexeddb-alternative.md` ("manually iterating over cursors")
- `articles/indexeddb/best-indexeddb-wrapper.md` ("iterating a cursor by hand")
- `articles/indexeddb/indexeddb-relationships.md` ("an index on a non-unique field")
- `slow-indexeddb.md` ("picking the right index for a query")
- `rx-query.md` (the query planner section)

Outbound, the article body carries in-text links to the cluster pages rather than a bare Follow Up list.

### Verification

`npm run docs:build` passes locally: exit 0, no errors, and Docusaurus runs with `onBrokenLinks: 'throw'`, so every internal link in the new page and the six edited pages resolves. The rendered page was checked for the chart and the new sections. The only build warnings are pre-existing broken anchors on unrelated pages.

## Todos
- [ ] Tests 
- [x] Documentation
- [ ] Typings 
- [ ] Changelog 

The header image `/headers/indexeddb-query.jpg` still needs to be added, same as for the other article pages.

https://claude.ai/code/session_01Njn8rcgx56cEiaiUs2r7w7


nickchomey:
Nice to see this doc. Definitely makes it clear how complicated it is to use indexeddb and how  much easier it all is with rxdb. Some comments:

1. you reference `onupgradeneeded` various times, but dont really ever explain what it is, when/how it is used. at least a link to this page might be helpful? https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Using_IndexedDB
2. you say `Indexes are declared once in the RxSchema, next to the fields they cover. There is no onupgradeneeded`. What is the actual implication of this? does rxdb use some other mechanisms that make it easier/faster to change indexes and schemas? or is it just a convenience that allows you to not worry about it?
3. when you say `Every index is a second copy of a field that the browser keeps in sync. So each index makes writes slower and takes disk space, which matters because the browser enforces a storage quota`,  it would be useful to give at least some indication of how much space indexes use and how much overhead it causes per write. Im sure it is impossible to be precise, but maybe some approximation/rule of thumb can be shared?
4. the FAQ seems nonsensical - it looks like its just answers without any actual questions... 

Also, you might be interested in this Interop 2027 proposal that I made for `extendedlifetime` sharedworkers. https://github.com/web-platform-tests/interop/issues/1382. If you could add an anecdote about how useful this would be to rxdb, your clients, and web devs in general, that would be helpful.

